PR #26, opened 2022-11-10
Rename authenticatedRequest and revisit tests

paulomarg:
We've agreed that the name for the `authenticatedRequest` middleware doesn't properly convey its purpose, so we're renaming it to `validateSession`.

We also agreed that the `ensureInstalled` middleware shouldn't _fail_ for non-embedded apps, but just proxy the request through to `validateSession` so we don't break apps unnecessarily.

In the process of making these updates, I noticed that some of the tests were "over-mocking" some functions that took away from the things we're actually testing.

### DIFF
--- a/src/__tests__/integration/oauth.test.ts
+++ b/src/__tests__/integration/oauth.test.ts
@@ -81,8 +81,8 @@ describe('OAuth integration tests', () => {
         next();
       });
       app.use('/test', shopify.app({afterAuth, webhookHandlers}));
-      app.get('/installed', shopify.ensureInstalled(), installedMock);
-      app.get('/authed', shopify.validateSession(), authedMock);
+      app.get('/installed', shopify.ensureInstalledOnShop(), installedMock);
+      app.get('/authed', shopify.validateAuthenticatedSession(), authedMock);
 
       const callbackInfo = await beginOAuth(app, shopify, config);
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ import {AppConfigParams, ShopifyApp, AppConfigInterface} from './types';
 import {AuthConfigInterface} from './auth/types';
 import {WebhooksConfigInterface} from './webhooks/types';
 import {
-  createValidateSession,
+  createValidateAuthenticatedSession,
   createCspHeaders,
   createEnsureInstalled,
 } from './middlewares/index';
@@ -36,12 +36,12 @@ export function shopifyApp<
     config: validatedConfig,
     api,
     app: createShopifyApp({api, config: validatedConfig}),
-    validateSession: createValidateSession({
+    validateAuthenticatedSession: createValidateAuthenticatedSession({
       api,
       config: validatedConfig,
     }),
     cspHeaders: createCspHeaders({api}),
-    ensureInstalled: createEnsureInstalled({
+    ensureInstalledOnShop: createEnsureInstalled({
       api,
       config: validatedConfig,
     }),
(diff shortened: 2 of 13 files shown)


src/__tests__/integration/oauth.test.ts
```diff
@@ -315,13 +315,13 @@ async function makeInstalledRequest(
...
   } else {
-    expect(response.status).toBe(500);
+    expect(response.status).toBe(config.embedded ? 500 : 302);
     expect(mock).not.toHaveBeenCalled();
   }
 }
 
 // Fires a valid request to check that the authenticated middleware allows requests through
-async function makeAuthenticatedRequest(
+async function makevalidateSession(
```
teddyhwang, line 324: I don't think the word `make` is necessary and `validSession` is enough:

```suggestion
async function validSession(
```

src/index.ts
```diff
@@ -15,7 +15,7 @@ import {AppConfigParams, ShopifyApp, AppConfigInterface} from './types';
 import {AuthConfigInterface} from './auth/types';
 import {WebhooksConfigInterface} from './webhooks/types';
 import {
-  createAuthenticatedRequest,
+  createValidateSession,
   createCspHeaders,
   createEnsureInstalled,
```
teddyhwang, line 20: this is outside the scope of this PR but is there a need to prefix these with `create`? I get that these are functions returning the middleware function because we need to pass in the api and config arguments to it but I don't think there are any conflicts and I think the word `create` here may be misleading to actually creating an authenticated request or creating a valid session.

src/middlewares/__tests__/ensure-installed.test.ts
```diff
@@ -112,12 +113,27 @@ describe('ensureInstalled', () => {
     );
   });
 
-  it('returns 500 if non-embedded app', async () => {
+  it('calls validateSession for non-embedded apps with a log', async () => {
```
mkevinosullivan, line 116: Should we be checking that the log message is output?

src/index.ts
teddyhwang: unless we have name conflicts then I don't see why we can't just drop `create`

teddyhwang: let's save this for a separate PR but I think having the following is sufficient:

```
validateSession(api, config)(req, res)
```

teddyhwang (commented):
I'm so sorry for not thinking of this earlier but I re-thought this and I think we should make: `validateSession` into `validateAuthenticatedSession` and `ensureInstalled` into `ensureInstalledOnShop`.

teddyhwang:
> > I'm so sorry for not thinking of this earlier but I re-thought this and I think we should make: `validateSession` into `validateAuthenticatedSession` and `ensureInstalled` into `ensureInstalledOnShop`.
> 
> Those feel a little bit on the wordy side, but I don't hate the clarity they bring.

I thought the same but given they're exposed at the template level, I really wanted to be explicit on what they were doing to bring clarity for users since I know this has confused folks on the Ruby side

mkevinosullivan (approved):
I know it makes the filenames longer but `ensure-installed.ts` should probably change to `ensure-installed-on-shop.ts` and `validate-session.ts` to `validate-authenticated-session.ts` - I believe that's idiomatic of JS/TS?

paulomarg:
> I know it makes the filenames longer but ensure-installed.ts should probably change to ensure-installed-on-shop.ts and validate-session.ts to validate-authenticated-session.ts - I believe that's idiomatic of JS/TS?

Absolutely, that was my mistake.

teddyhwang (approved):

PR #34, opened 2022-11-30
Issue 585: multiple app on the same host can't run side by side

cquemin:

Fixes #[585](https://github.com/Shopify/shopify-api-node/issues/585)
Fix suggested by [treboryx](https://github.com/treboryx) on the [`Shopify-api-node` repo](https://github.com/Shopify/shopify-api-node/pull/607)

Session management can be implemented with relational databases. As part of the initialization of the table that will persist session, we are querying the information schema table to check if we have already created the session table. This query does not filter per database. Therefore when there are multiple databases on the same host  we may return that a table has been created but it is not necessary for the correct database.
This prevents multiple app that runs on the same host to be able to work side by side.

This is relevant only for `MySql` and `Postgres` because of the way information tables are server wide and not database specific. For the other implementation of the session storage this is not an issue.

This PR add the current database as projection criteria on the select statement that looks for the session table. It also modify the battery of tests that run for both `MySql` and `Postgre` so that we have 2 databases created on the same host and we make sure that the execution of the tests run side by side for those 2 databases.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-session-storage-mysql/src/__tests__/mysql.test.ts
+++ b/packages/shopify-app-session-storage-mysql/src/__tests__/mysql.test.ts
@@ -56,6 +56,14 @@ describe('MySQLSessionStorage', () => {
     await exec(`podman rm -f ${containerId}`);
   });
 
-  batteryOfTests(async () => storage);
-  batteryOfTests(async () => storage2);
+  const tests = [
+    {dbName: 'shopitest', sessionStorage: async () => storage},
+    {dbName: 'shopitest2', sessionStorage: async () => storage2},
+  ];
+
+  for (const {dbName, sessionStorage} of tests) {
+    describe(`with ${dbName}`, () => {
+      batteryOfTests(sessionStorage);
+    });
+  }
 });


packages/shopify-app-session-storage-postgresql/src/__tests__/postgresql.test.ts
```diff
@@ -42,13 +60,18 @@ describe('PostgreSQLSessionStorage', () => {
...
 
   afterAll(async () => {
     await storage.disconnect();
+    await storage2.disconnect();
+
     await exec(`podman rm -f ${containerId}`);
   });
 
   batteryOfTests(async () => storage);
+  batteryOfTests(async () => storage2);
```
mkevinosullivan, line 76:
```suggestion
  const tests = [
    {dbName: 'shopitest', sessionStorage: async () => storage},
    {dbName: 'shopitest2', sessionStorage: async () => storage2},
  ];

  for (const {dbName, sessionStorage} of tests) {
    describe(`with ${dbName}`, () => {
      batteryOfTests(sessionStorage);
    });
  }
```
It's a bit of a nit bit if you structure the call to the tests like this, you'll end up with a more descriptive test output, like
Screenshot 2022-12-01 at 2 23 14 PM

packages/shopify-app-session-storage-mysql/src/__tests__/mysql.test.ts
```diff
@@ -36,12 +47,16 @@ describe('MySQLSessionStorage', () => {
...
 
   afterAll(async () => {
     await storage.disconnect();
+    await storage2.disconnect();
+
     await exec(`podman rm -f ${containerId}`);
   });
 
   batteryOfTests(async () => storage);
+  batteryOfTests(async () => storage2);
```
mkevinosullivan, line 61: See my comment in the `postgresql.test.ts` file ... can also be applied here.

mkevinosullivan (approved):
I've ran the tests and it all seems good!

PR #35, opened 2022-12-01
Replace auth sub-app components with middleware

paulomarg:

Currently, a considerable amount of app logic is tucked away behind the `app` sub-app - the endpoints for oauth and webhook are automatically defined by it, which makes it harder than it needs to be for apps to override behaviours, e.g. adding logic after oauth completes.

Unplugging the auth endpoints from `app`, exposing them instead as `shopify.auth.begin/callback`, which are both regular middlewares. To make it possible for apps to add their own post-callback logic, we're also exposing a new `redirectToShopifyOrAppRoot` middleware that just handles redirecting back to the appropriate location.

I'll be updating all of the docs later on once I fix up the webhooks component as well.

## Checklist


packages/shopify-app-express/src/auth/types.ts
```diff
@@ -1,25 +1,17 @@
...
+  begin: () => RequestHandler;
+  callback: () => RequestHandler;
 }
 
-export interface AuthBeginParams {
+export interface AuthCallbackParams extends ApiAndConfigParams {
   req: Request;
   res: Response;
   api: Shopify;
   config: AppConfigInterface;
```
mkevinosullivan, line 16: If it's now `extends ApiandConfigParams`, we don't need these `api` and `config` params to be explicitly included

mkevinosullivan (approved):
Will the `create...` methods be renamed in another PR?

.changeset/famous-kids-smell.md
```diff
@@ -0,0 +1,2 @@
+---
```
teddyhwang, line 1: what is this file?

packages/shopify-app-express/src/auth/index.ts
```diff
@@ -1,30 +1,28 @@
...
 
-  subApp.get(config.auth.callbackPath, async (req: Request, res: Response) => {
-    await config.logger.debug('Initiating auth callback request');
+export function auth({api, config}: ApiAndConfigParams): AuthMiddleware {
+  return {
+    begin(): RequestHandler {
+      return async (req: Request, res: Response) =>
+        redirectToAuth({req, res, api, config});
+    },
+    callback(): RequestHandler {
```
teddyhwang, line 15: since there are no arguments here, is there any reason this can't be the actual express middleware function itself?

paulomarg: I was wondering about that - we've used this pattern of making all of the middlewares we expose functions to make it possible to introduce arguments to them in the future without making it a breaking change.

paulomarg:
> Will the create... methods be renamed in another PR?

Yes, there are more PRs coming!

packages/shopify-app-express/src/auth/auth-callback.ts
```diff
@@ -7,25 +7,29 @@ import {
...
-import {authBegin} from './auth-begin';
 
 export async function authCallback({
   req,
   res,
   api,
   config,
-  afterAuth,
 }: AuthCallbackParams) {
+  await config.logger.info('Initiating auth callback request');
```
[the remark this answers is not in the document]
paulomarg: Yeah, we should reword this one.

.changeset/famous-kids-smell.md
teddyhwang: That seems odd? What is the CI check doing?

paulomarg:
I tried to build this and got a failure, so I added a build step to the CI workflow, to make sure we're not setting ourselves up for failing when deploying something.

.changeset/famous-kids-smell.md
mkevinosullivan: See bullet no. 2 [here](https://github.com/Shopify/shopify-app-js/blob/next/RELEASING.md#releasing-shopify-app-js-packages)

This file is the result of running `yarn changeset --empty`, which is indicating to `changesets` that the dev doesn't want to anything to the `CHANGELOG.md` files for this PR.

The CI check is seeing if there's a changeset file being added for the PR - it's a programmatic check to see if the dev has added an entry for the CHANGELOG, like the CI check for the tick box in the PR template!

PR #51, opened 2022-12-08
Add KV session storage package

paulomarg:

Closes https://github.com/Shopify/first-party-library-planning/issues/509

Now that we have separate packages for session storage and they're not tied to the express package, we can re-introduce our KV storage offering.

Re-creating the KV storage as its own package, and updating it to work with the v6 `Session` interface.

## Type of change

## Checklist


paulomarg:
Am I missing any docs?

mkevinosullivan (approved):

PR #57, opened 2022-12-13
Validate session before embedding

paulomarg:

Closes https://github.com/Shopify/first-party-library-planning/issues/518

There is an edge case where one can

1. Run app in development mode
1. Install the app
2. Stop development server
3. Uninstall the app, so the `APP_UNINSTALLED` webhook never makes it to the app
4. Restart the server and go to the app's `/` endpoint

In this case, the app would still have a database entry for the shop (so it is considered installed), but it would be uninstalled from Shopify's perspective.

When the user tries to go to the `/` endpoint, the app embeds itself into Shopify thinking everything is ok, but that ends up loading a `Page not found` error because the app isn't actually installed any more.

Addressing this scenario by doing a pre-flight check to see if the token we see in the database is _actually_ a valid one, by making a request to Shopify.

This request is only triggered if the app is supposed to be embedded, but it is currently loading from _outside_ of the Admin. Essentially, this should only happen when the app is being installed / re-installed.

In that scenario, if the token is now invalid we know we need to reinstall the app, so we redirect to the login path instead of embedding into Shopify.

## Type of change

## Checklist


packages/shopify-app-express/src/middlewares/ensure-installed-on-shop.ts
```diff
@@ -97,3 +87,78 @@ export function deleteAppInstallationHandler(
...
+  req: Request,
+  res: Response,
+): string | undefined {
+  if (typeof req.query.shop !== 'string') {
+    config.logger.error(
+      'ensureInstalledOnShop did not receive a shop query argument',
+      {shop: req.query.shop},
+    );
+
+    res.status(500);
```
teddyhwang, line 103: this should be a `422`: https://stackoverflow.com/questions/3050518/what-http-status-response-code-should-i-use-if-the-request-is-missing-a-required
```suggestion
    res.status(422);
```

packages/shopify-app-express/src/middlewares/ensure-installed-on-shop.ts
```diff
@@ -97,3 +87,78 @@ export function deleteAppInstallationHandler(
...
+
+  const shop = api.utils.sanitizeShop(req.query.shop);
+
+  if (!shop) {
+    config.logger.error(
+      'ensureInstalledOnShop did not receive a valid shop query argument',
+      {shop: req.query.shop},
+    );
+
+    res.status(500);
```
teddyhwang, line 116:
```suggestion
    res.status(422);
```

teddyhwang (commented):
LGTM, one minor feedback

PR #61, opened 2022-12-14
Optimize findSessionsByShop

mkevinosullivan:

If the Redis store is being used for store objects other than of type `Session`, `findSessionsByShop` breaks because it iterates over _all_ keys to find the ones specific for a shop, assuming that all objects are of type `Session`.

For each session added to the redis store, we add a key based on the shop name and add the session id key to a set of all session ids for that shop.

This allows us to quickly find all sessions for a given shop, and avoids having to iterate over all sessions in the store to find the ones for a given shop.

## Type of change

## Checklist

- [ ] ~I have added/updated tests for this change~ - not applicable
- [ ] ~I have documented new APIs/updated the documentation for modified APIs (for public APIs)~ - not applicable

### DIFF
--- a/.changeset/warm-dots-applaud.md
+++ b/.changeset/warm-dots-applaud.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-session-storage-redis': patch
+---
+
+Optimize findSessionsByShop with additional shop-based key, listing array of corresponding session ids
--- a/packages/shopify-app-express/docs/reference/processWebhooks.md
+++ b/packages/shopify-app-express/docs/reference/processWebhooks.md
@@ -15,7 +15,7 @@ Make sure you use this middleware on a `.post()` route.
 
 `{[topic: string]: WebhookHandler | WebhookHandler[]}`
 
-Defines the webhooks your app will listen to, and how to handle them. See [the `@shopify/shopify-api` documentation](https://github.com/Shopify/shopify-api-js/blob/main/docs/usage/webhooks.md) for the allowed values.
+Defines the webhooks your app will listen to, and how to handle them. See [the `@shopify/shopify-api` documentation](https://github.com/Shopify/shopify-api-js/blob/main/docs/guides/webhooks.md) for the allowed values.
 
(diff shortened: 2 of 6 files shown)


cquemin (commented):
LGTM.

As this was failing before, I'd suggest to add this as part of the test to ensure that other data can safely live in the redis DB.
```javascript
type RedisClient = ReturnType<typeof createClient>;
let client: RedisClient;

describe('RedisSessionStorage', () => {
  let storage: RedisSessionStorage | undefined;
  let containerId: string | undefined;
  beforeAll(async () => {
    const configPath = resolve(__dirname, './redis.conf');
    const runCommand = await exec(
      `podman run -d -p 6379:6379 -v ${configPath}:/redis.conf redis:6 redis-server /redis.conf`,
      {encoding: 'utf8'},
    );
    containerId = runCommand.stdout.trim();

    // Give the container a lot of time to set up since polling is ineffective with podman
    await wait(10000);

    storage = new RedisSessionStorage(dbURL);

    // Add different type of object the in the DB
    await initWithNonSessionData();

    await storage.ready;
  });

  afterAll(async () => {
    await storage?.disconnect();
    await client.disconnect();

    if (containerId) await exec(`podman rm -f ${containerId}`);
  });

  batteryOfTests(async () => storage!);
});

async function initWithNonSessionData() {
  client = createClient({url: dbURL.toString()});
  await client.connect();

  // Json type different from json array contrary to what is expected by Session.fromPropertyArray
  await client.set(
    'this_is_an_id',
    JSON.stringify('{"key": "with a random value"}'),
  );

  // Type json array as expected by the the Session.fromPropertyArray but does not contain session data
  await client.set('12563', '["Ford", "BMW", "Fiat"]');

  // Not a json value
  await client.set('256647', 'any random value really');
}
```

cquemin:
@paulomarg I actually quite like the idea of wiping all sessions, it seems like a low cost migration. Do you think there are some bad side effect to delete all sessions ? The client would simply re authenticate no ?

cquemin:
I see your point. However that would be only for the one that use our Redis implementation without modification and that update.
Do we know how used is our Redis implem?

PR #62, opened 2022-12-16
Add migration logic and tests

mkevinosullivan:

With the addition of a new set of keys to optimize `findSessionsByShop`, existing Redis session stores will stop working with the new package version.

During initialization, the library will check if there's an admin record, if not, will migrate the data; if there is, will check if the data has been migrated.

## Type of change

## Checklist

- [ ] ~I have added a changelog entry, prefixed by the type of change noted above~ - not applicable, will be added to the `update_redis_implementation` branch
- [ ] ~I have documented new APIs/updated the documentation for modified APIs (for public APIs)~ - not applicable

### DIFF
--- a/packages/shopify-app-session-storage-redis/src/redis.ts
+++ b/packages/shopify-app-session-storage-redis/src/redis.ts
@@ -2,7 +2,7 @@ import {createClient, RedisClientOptions} from 'redis';
 import {Session} from '@shopify/shopify-api';
 import {SessionStorage} from '@shopify/shopify-app-session-storage';
 
-import {migrateToVersion2_0_0} from './migrations';
+import {migrateToVersion1_0_1} from './migrations';
 
 type RedisClient = ReturnType<typeof createClient>;
 export interface RedisSessionStorageOptions extends RedisClientOptions {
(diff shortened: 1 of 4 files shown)


packages/shopify-app-session-storage-redis/src/redis.ts
```diff
@@ -151,5 +153,42 @@ export class RedisSessionStorage implements SessionStorage {
...
+    await this.migrate();
+  }
+
+  private async migrate() {
+    const migrationsRecord = await this.client.get(this.fullKey('migrations'));
+    if (migrationsRecord) {
+      // need to check if we need to migrate
+      // as of 2.0.0, this code is illustrative only!
+      const migrations = JSON.parse(migrationsRecord);
+      if (!migrations.migrateToVersion2_0_0) {
```
cquemin, line 165: There seems to be a lot of common code between the 2 branches of your `if`?Can they be mutualized ?

cquemin (approved):
This is a much better solution. I would have thought the DB wipe was good enough but this is definitely future proof!

packages/shopify-app-session-storage-redis/src/redis.ts
mkevinosullivan: Yes! Now rewritten.

PR #65, opened 2022-12-19
Modify postgres to make table and column names case sensitive

mkevinosullivan:

Fixes Shopify/shopify-api-js#460

Adding `"` around table and column names so that case sensitivity is preserved.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-session-storage-postgresql/src/__tests__/mytest.ts
+++ b/packages/shopify-app-session-storage-postgresql/src/__tests__/mytest.ts
@@ -1,0 +1,105 @@
+import * as child_process from 'child_process';
+import {promisify} from 'util';
+
+import pg from 'pg';
+
+function wait(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(() => resolve(), ms));
+}
+async function poll(
+  func: () => Promise<boolean>,
+  {interval = 100, timeout = 5000} = {},
+) {
+  const start = Date.now();
+  while (true) {
+    const elapsed = Date.now() - start;
+    if (elapsed > timeout) {
+      throw Error('Timeout');
+    }
+
+    try {
+      const success = await func();
+      if (success) return;
+    } catch {
+      /* lol empty */
+    }
+    await wait(interval);
+  }
+}
+
+(async () => {
+  const exec = promisify(child_process.exec);
+
+  const dbURL = new URL(
+    'postgres://shopify:passify@localhost:5433/migrationtest',
+  );
+  const sessionTableName = 'ShopifySessions';
+
+  const runCommand = await exec(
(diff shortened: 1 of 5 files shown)


cquemin:
@mkevinosullivan I believe this will effectively "loose" all previously stored session since we will now be pointing at `ShopifySession` instead of `shopifysession` before your changes. If we do that should we then import all data ?

mkevinosullivan:
> @mkevinosullivan I believe this will effectively "loose" all previously stored sessions since we will now be pointing at `ShopifySession` instead of `shopifysession` before your changes. If we do that should we then import all data ?

I'm going to add a flag to let the dev decide if they want case-sensitivity or not ... doing a migration here might be trickier ... although it's should just be as simple as running ...

```
ALTER TABLE shopifysession RENAME TO "ShopifySession";
ALTER TABLE "ShopifySession" RENAME COLUMN isonline TO "isOnline";
ALTER TABLE "ShopifySession" RENAME COLUMN onlineaccessinfo TO "onlineAccessInfo";
ALTER TABLE "ShopifySession" RENAME COLUMN accesstoken TO "accessToken";
```

Let me try writing a migration ...

paulomarg:
If we have this sort of migration strategy in multiple offerings, should we bake the logic for it into the base package?

cquemin:
@paulomarg we can but should we ? As I am starting to think that each provider will have its way to handle migration, maybe even a recommended migration manager .... It is a genuine question. As we are starting to be in muddy waters: borders between giving sample usages and  making tech choices.. Maybe we can simply keep it vanilla everywhere but this may mean rewriting a migration mechanism so it is future proof...

packages/shopify-app-session-storage-postgresql/src/__tests__/mytest.ts
```diff
@@ -0,0 +1,105 @@
+import * as child_process from 'child_process';
```
cquemin, line 1: Does this file need to stay?

packages/shopify-app-session-storage-postgresql/src/postgresql.ts
```diff
@@ -176,4 +178,31 @@ export class PostgreSQLSessionStorage implements SessionStorage {
...
+
+    // Allow multiple apps to be on the same host with separate DB and querying the right
+    // DB for the session table exisitence
+    const rows = await this.query(query, [
+      this.options.sessionTableName.toLowerCase(),
+    ]);
+    return Array.isArray(rows) && rows.length === 1;
+  }
+
+  private async migrateSessionTableV1_0_1(): Promise<void> {
```
cquemin, line 195: Ideally I merge my [PR](https://github.com/Shopify/shopify-app-js/pull/95) (hint hint for a review) and then you'll have to add this in the migrations file to be run as part of the new migration system :D

cquemin (commented):
Small question for me. 
FYI: indirect dependency on https://github.com/Shopify/shopify-app-js/pull/95

mkevinosullivan:
Replaced with #113

PR #66, opened 2022-12-19
Handle errors when loading sessions to validate

### LINKED ISSUE #63: Q: How to handle InvalidJwtError, InvalidSessionError using shopify.validateAuthenticatedSession()?
Hi all,

Regarding last versions, would like to get into the idea of what was before now's wrapped into shopify, how do you guys handle InvalidJwtError on session?

```
// All endpoints after this point will require an active session
app.use("/api/*", shopify.validateAuthenticatedSession());
```

If the session is corrupted the server goes down, it means we've got to get back old methods of handling errors, instances of ShopifyErrors.

Any advice on how to do it?

Do I have to open the middleware function and super call this with params req, res? Is this really a solution to an integrated session handling verification?

ps. custom express error hadling doesn't work, it crashes on shopify.validateAuthenticatedSession()

Thanks,

paulomarg:

Fixes #63 

There are currently some errors in `validateAuthenticatedSession` that are not being handled, when the JWT is invalid or when the session storage fails to load the actual data. When they occur, the middleware is throwing an error instead of responding properly, which causes apps to crash because Express 4 doesn't handle async errors.

Improving error handling in the middleware, so that the errors don't bubble out of the call.

We return:

- a 401 if the session token is invalid. This is related to the App Bridge session token, and not to the Admin API session, so we don't return the 403 for re-authentication, instead just failing the request as something went wrong on the app's client side.
- a 500 for non-Shopify errors.

## Type of change

## Checklist


packages/shopify-app-express/src/middlewares/validate-authenticated-session.ts
```diff
@@ -18,17 +18,38 @@ export function validateAuthenticatedSession({
...
+        await handleSessionError(req, res, error);
+        return undefined;
+      }
 
-      let shop = req.query.shop;
+      let session: Session | undefined;
+      if (sessionId) {
+        try {
+          session = await config.sessionStorage.loadSession(sessionId);
+        } catch (error) {
```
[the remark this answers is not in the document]
paulomarg: Here we _have_ to catch any errors because the session storage might be user-provided, so we can't tell what kind of errors we might get, and we don't want those to crash the app.

cquemin (approved):

packages/shopify-app-express/src/middlewares/validate-authenticated-session.ts
```diff
@@ -61,17 +82,6 @@ export function validateAuthenticatedSession({
         }
       }
 
-      const bearerPresent = req.headers.authorization?.match(/Bearer (.*)/);
-      if (bearerPresent) {
-        if (!shop) {
-          shop = await setShopFromSessionOrToken(
```
walkingbrad, line 67: Hey @paulomarg, I believe the removal of this code path introduced an issue #94. I have a comment there at the bottom explaining my particular situation, but there are a handful of other people with the issue there as well.

Without having much context, my guess is that we still need this fallback logic to set the `shop` variable when there are zero sessions in session storage (and no `shop` param in the request, of course).

walkingbrad: Sorry, I meant specifically the code below:
```
const payload = await api.session.decodeSessionToken(token);
shop = payload.dest.replace('https://', '');
```

PR #76, opened 2022-12-24
Postgres changed from single connection to pool

markustripp:
### Overview/summary

The Postgres SessionStorage interface establishes a single connection to the Postgres database. When this connection died for some reason, the entire Shopify app does not work anymore. There is no mechanism for connection management and reconnection. I had this problem with some of my Shopify apps that suddenly stopped working. Sometimes after a few hours, sometimes after a couple of days. Fun to debug. I deploy my Shopify apps to fly.io with my database at Supabase. The reason why the connection gets closed is somewhere in the infrastructure between fly.io and Supabase. Hard to say. Anyway, using the connection pool from the pg module guarantees a reliable connection between the app and the database.

### Fix

In my PR, I simply replaced the pg.Client with the Pool from the pg module which manages the connections. As the Pool class does not have access to the database variable, I have to use the connection string parser from the pg module and store the database in a separate variable. Loom tests pass.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-session-storage-postgresql/src/postgresql.ts
+++ b/packages/shopify-app-session-storage-postgresql/src/postgresql.ts
@@ -133,25 +133,9 @@ export class PostgreSQLSessionStorage implements SessionStorage {
     await this.client.connect();
   }
 
-  private async hasSessionTable(): Promise<boolean> {
-    const query = `
-      SELECT tablename FROM pg_catalog.pg_tables WHERE tablename = $1 AND schemaname = $2
-    `;
-
-    // Allow multiple apps to be on the same host with separate DB and querying the right
-    // DB for the session table exisitence
-    const rows = await this.query(query, [
-      this.options.sessionTableName,
-      this.client.database,
-    ]);
-    return Array.isArray(rows) && rows.length === 1;
-  }
-
   private async createTable() {
-    const hasSessionTable = await this.hasSessionTable();
-    if (!hasSessionTable) {
-      const query = `
-        CREATE TABLE ${this.options.sessionTableName} (
+    const query = `
+        CREATE TABLE IF NOT EXISTS ${this.options.sessionTableName} (
           id varchar(255) NOT NULL PRIMARY KEY,
           shop varchar(255) NOT NULL,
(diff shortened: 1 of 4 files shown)


SeanMythen:
I'm running into this same issue myself, where after a few hours of inactivity, my app will crash due to the database not being connected anymore.

My error typically looks something like this: 
```
node:events:491
throw er; // Unhandled 'error' event
^
Error: Connection terminated unexpectedly
at Connection.<anonymous> (/app/node_modules/pg/lib/client.js:132:73)
```

After doing research I'm not sure if it has anything to do with a deeper error handling issue within the pg library
https://github.com/brianc/node-postgres/issues/1611#issuecomment-558566873

Has using Pool fixed the issue entirely for you?

SeanMythen:
How can we get this merged into the main branch? Is all that is needed for a further review is to sign the CLA?

SeanMythen:
@conspireagency is this an issue you face too using the postgres adapter?

packages/shopify-app-session-storage-postgresql/src/postgresql.ts
```diff
@@ -43,6 +45,9 @@ export class PostgreSQLSessionStorage implements SessionStorage {
     }
     this.options = {...defaultPostgreSQLSessionStorageOptions, ...opts};
     this.ready = this.init();
+
+    const config = parse(dbUrl.toString());
+    this.database = config.database || 'postgres';
```
cquemin, line 50: I would suggest to use the right side of your assignment straight where `this.database` is used L145. I think it increases readability and reduces ever so slightly the cognitive load.

cquemin (changes requested):
Thanks a lot for your contribution. I think this is a great improvement.

However since we offer many session storage implementations, would you be able to make the same changes for the other implementation where relevant? Otherwise we would have inconsistency between the implementations.
 
After you run the tests locally with commands like: `yarn test packages/shopify-app-session-storage-postgresql`. You will need to add some change logs. We are using `changeset` on this repo so you will need to run `yarn changeset` and follow the instructions on the screen. After that, don't forget to push the modified/generated file.

PR #95, opened 2023-01-18
Added generic migration mechanism with application to scope column increase

### LINKED ISSUE #78: scope column with 255 chars is not enough
When having a few `unauthenticated_*` scopes the 255 limit is not enough. and cause the app install to fail

- `@shopify/shopify-app-*` package and version: 1.0.1

## Expected behavior

no crash

## Actual behavior

crash when trying to save the scope string into the DB

1. add a few scopes `unauthenticated_` to the App
2. install the app
3. crash when trying to save the scope string into the DB

cquemin:

Fixes #78  
Fixes [shopify-api-js/457](https://github.com/Shopify/shopify-api-js/issues/457)

The RDBMS session storage implementations all define the column `scope` as a `varchar(255)`, but there are valid use case where scope could be longer

Thos increase the size of the column from 255 to 1024. For SQLite, as there is no ALTER command in change the size of a column, we need to create a migration for the existing session stored. So that we can migrate all exsiting session in a new table where `scope` has 1024 length.

Since we altering the database this should be done through a migration mechanism as put in place for the `redis session storage`. This PR makes a generic version of `session storage migrator`. There are interfaces and common classes in the `session-storage` package whilst you can find database specific implementation in each implementation of the session storage.

In particular, on top of the new `Migrator` classes, there is an interface that defines an `DBConnection`. And an implementation of it for each session storage. This `DBConnection` is then used, internally in the session storage class and by the migrator class (when migration/s is/are required).

The behaviour and the public interface of the `Session storage` should not be broken. Fields have been added to the option interface but with a default value for them. The migrator is executed internally and transparently for the user.

## Type of change

## Checklist


packages/shopify-app-session-storage-mysql/src/mysql.ts
```diff
@@ -153,6 +153,10 @@ export class MySQLSessionStorage implements SessionStorage {
       `;
       await this.query(query);
     }
+
+    // Update for existing tables
+    await this.query(`ALTER TABLE ${this.options.sessionTableName}
```
paulomarg, line 158: I wonder if we can do this in a way that doesn't require us to run this code every time we start up, but I can't think of a clean way other than setting up migrations like we did for sqlite.

paulomarg (commented):
Code-wise looks fine to me, but I wonder if we want to add migrations all around.

packages/shopify-app-session-storage-mysql/src/mysql.ts
cquemin: As discussed I'll generalized migration for all session storage and allow it to be customizable as well.

packages/shopify-app-session-storage-redis/src/redis-session-storage-migrator.ts
```diff
@@ -0,0 +1,54 @@
...
+  constructor(
+    db: RedisEngine,
+    opts: Partial<SessionStorageMigratorOptions> = {},
+  ) {
+    // The name has already been decided whith the last migration
+    opts.migrationTableName = 'migrations';
+    super(db, opts);
+  }
+
+  async initMigrationPersitance(): Promise<void> {
```
mkevinosullivan, line 18:
```suggestion
  async initMigrationPersistance(): Promise<void> {
```

packages/shopify-app-session-storage-redis/src/redis-session-storage-migrator.ts
```diff
@@ -0,0 +1,54 @@
...
+    // nothing to do here
+    return Promise.resolve();
+  }
+
+  async hasVersionBeenApplied(versionName: string): Promise<boolean> {
```
mkevinosullivan, line 23:
```suggestion
  async hasMigrationBeenApplied(migrationName: string): Promise<boolean> {
```

packages/shopify-app-session-storage-redis/src/redis-session-storage-migrator.ts
```diff
@@ -0,0 +1,54 @@
...
+    const migrations = await this.getMigrationRecords();
+    const found =
+      migrations.has(versionName) && (migrations.get(versionName) || false);
```
mkevinosullivan, line 26:
```suggestion
      migrations.has(migrationName) && (migrations.get(migrationName) || false);
```

mkevinosullivan: Why use `has` and `get`?  Wouldn't one of them suffice?

packages/shopify-app-session-storage-redis/src/redis-session-storage-migrator.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    return Promise.resolve(found);
+  }
+
+  async saveAppliedVersion(versionName: string): Promise<void> {
```
mkevinosullivan, line 31:
```suggestion
  async saveAppliedMigration(migrationName: string): Promise<void> {
```

packages/shopify-app-session-storage-redis/src/redis-session-storage-migrator.ts
```diff
@@ -0,0 +1,54 @@
...
+    const migrations = await this.getMigrationRecords();
+    migrations.set(versionName, true);
```
mkevinosullivan, line 33:
```suggestion
    migrations.set(migrationName, true);
```

packages/shopify-app-session-storage-redis/src/redis.ts
```diff
@@ -1,16 +1,27 @@
...
 export interface RedisSessionStorageOptions extends RedisClientOptions {
   sessionKeyPrefix: string;
   onError?: (...args: any[]) => void;
+  migratorOptions: SessionStorageMigratorOptions | null;
 }
 const defaultRedisSessionStorageOptions: RedisSessionStorageOptions = {
   sessionKeyPrefix: 'shopify_sessions',
+  migratorOptions: {
+    migrationTableName: 'migrations',
+    versionColumnName: '',
```
mkevinosullivan, line 22:
```suggestion
    migrationNameColumnName: '',
```

packages/shopify-app-session-storage-sqlite/src/sqlite.ts
```diff
@@ -1,26 +1,48 @@
...
 
-export interface SQLiteSessionStorageOptions {
-  sessionTableName: string;
-}
 const defaultSQLiteSessionStorageOptions: SQLiteSessionStorageOptions = {
   sessionTableName: 'shopify_sessions',
+  sqlArgumentPlaceholder: '?',
+  migratorOptions: {
+    migrationTableName: 'shopify_sessions_migrations',
+    versionColumnName: 'version',
```
mkevinosullivan, line 22:
```suggestion
    migrationNameColumnName: 'migration_name',
```

packages/shopify-app-session-storage-sqlite/src/sqlite.ts
```diff
@@ -116,30 +133,33 @@ export class SQLiteSessionStorage implements SessionStorage {
           state varchar(255) NOT NULL,
           isOnline integer NOT NULL,
           expires integer,
-          scope varchar(255),
+          scope varchar(1024), -- sqlite allows more than this limit without failing
```
mkevinosullivan, line 136:
```suggestion
          scope varchar(1024),
```
Why would it be important to keep this comment?

packages/shopify-app-session-storage-mysql/src/mysql.ts
```diff
@@ -1,12 +1,27 @@
...
+  SessionStorageMigratorOptions,
+  SessionStorageMigrator,
+  RdbmsSessionStorageMigrator,
+  RdbmsSessionStorageOptions,
+} from '@shopify/shopify-app-session-storage';
+
+import {migrationMap} from './migrations';
+import {MySqlEngine} from './mysql-engine';
+
+export interface MySQLSessionStorageOptions
```
paulomarg, line 14: This might break existing TS code since there are new mandatory fields in `RdbmsSessionStorageOptions`.

packages/shopify-app-session-storage-mysql/src/mysql.ts
```diff
@@ -1,12 +1,27 @@
...
+  extends RdbmsSessionStorageOptions {}
 
-export interface MySQLSessionStorageOptions {
-  sessionTableName: string;
-}
 const defaultMySQLSessionStorageOptions: MySQLSessionStorageOptions = {
   sessionTableName: 'shopify_sessions',
+  sqlArgumentPlaceholder: '?',
```
paulomarg, line 19: I feel like this should be a static parameter of the mysql implementation, and we would just have a generic `placeholders()` method called by the abstract class whenever `query` is called. So something like

```ts
this.query(query, [1, 2]);

function query(query, args) {
  this.queryImplementation(query, this.escapeArgsImplementation(args));
}
```

does that make sense? 😄

packages/shopify-app-session-storage-mysql/src/mysql.ts
```diff
@@ -151,17 +165,31 @@ export class MySQLSessionStorage implements SessionStorage {
...
-  }
-
   private databaseRowToSession(row: any): Session {
     // convert seconds to milliseconds prior to creating Session object
     if (row.expires) row.expires *= 1000;
     return Session.fromPropertyArray(Object.entries(row));
   }
+
+  private async initMigrator(
+    migratorOptions?: SessionStorageMigratorOptions | null,
```
paulomarg, line 179: REALLY nit-y: if we already allow `migratorOptions` to be optional, do we need the `null` value as well? We could just either get a well-formed object, or nothing at all (i.e. `undefined`), right?

packages/shopify-app-session-storage/src/rdbms-session-storage-migrator.ts
```diff
@@ -0,0 +1,88 @@
...
+        this.options.migrationTableName,
+      );
+    } else {
+      ifNotExist = 'IF NOT EXISTS';
+    }
+
+    const migration = `
+      CREATE TABLE ${ifNotExist} ${this.options.migrationTableName} (
+        ${this.options.versionColumnName} varchar(255) NOT NULL PRIMARY KEY
+    );`;
```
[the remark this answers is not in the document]
cquemin: @paulomarg doing so would mean I'd have a Migration specialized class for each DB implementation whereas currently they're all using the RDBMS one.
Which is fine anyway but worth mentioning.

packages/shopify-app-session-storage-redis/src/redis-connection.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  connect(): Promise<void> {
+    return this.client.connect();
+  }
+
+  async disconnect(): Promise<void> {
+    await this.client.quit();
+  }
+
+  keys(name: string): Promise<any> {
```
paulomarg, line 30: Should these functions that return promises all be marked as `async`?

packages/shopify-app-session-storage-sqlite/src/migrations.ts
```diff
@@ -0,0 +1,41 @@
+import {SqliteConnection} from './sqlite-connection';
+
+export const migrationMap = new Map([['internal_v101', migrateToV1_0_1]]);
+
+// need to migrate exisiting scope from varchar 255 to varchar 1024
+async function migrateToV1_0_1(connection: SqliteConnection): Promise<void> {
+  const tempTableName = `${connection.sessionDBIdentifier}_for_migration_toV101`;
+
+  //  1. rename exisiting table
+  const rename = `
```
paulomarg, line 10: Nit: would it be possible to run this migration in a transaction to prevent data loss?

paulomarg (approved):
LGTM, just nits and non-blocking comments.

packages/shopify-app-session-storage-redis/src/migrations.ts
```diff
@@ -1,21 +1,22 @@
...
-  fullKey: (name: string) => string,
-) {
+  connection: RedisConnection,
+): Promise<void> {
   const shopsAndSessions: {[key: string]: string[]} = {};
-  const keys = await client.keys('*');
+  const keys = await connection.keys('*');
   for (const key of keys) {
-    if (key.startsWith(sessionKeyPrefix)) {
+    if (key.startsWith(connection.sessionDBIdentifier)) {
```
mkevinosullivan, line 17:
```suggestion
    if (key.startsWith(connection.sessionTableName)) {
```

packages/shopify-app-session-storage-postgresql/src/migrations.ts
```diff
@@ -0,0 +1,13 @@
+import {PostgresConnection} from './postgres-connection';
+
+export const migrationMap = new Map([
+  ['migrateToVersion1_0_1', migrateToVersion1_0_1],
```
mkevinosullivan, line 4:
```suggestion
  ['migrateScopeFieldToVarchar1024', migrateScopeFieldToVarchar1024],
```

packages/shopify-app-session-storage-postgresql/src/migrations.ts
```diff
@@ -0,0 +1,13 @@
...
+]);
+
+// need change the sizr of the scope column from 255 to 1024 char
+export async function migrateToVersion1_0_1(
```
mkevinosullivan, line 8:
```suggestion
export async function migrateScopeFieldToVarchar1024(
```

packages/shopify-app-session-storage-sqlite/src/migrations.ts
cquemin: Good point !!!

mkevinosullivan (commented):
I think changing the `sessionTableName` that was in each individual Session Storage's Options type will become a breaking change, so the interface that the Options types all extend from now should probably keep that property name.

packages/shopify-app-session-storage-sqlite/src/migrations.ts
```diff
@@ -0,0 +1,41 @@
+import {SqliteConnection} from './sqlite-connection';
+
+export const migrationMap = new Map([['internal_v101', migrateToV1_0_1]]);
+
+// need to migrate exisiting scope from varchar 255 to varchar 1024
+async function migrateToV1_0_1(connection: SqliteConnection): Promise<void> {
```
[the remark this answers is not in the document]
paulomarg: I like that idea, it should make it easier to reason about them in the future.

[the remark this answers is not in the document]
cquemin: Yes it is a good idea!

packages/shopify-app-session-storage/src/types.ts
```diff
@@ -0,0 +1,179 @@
...
+  sessionDBIdentifier: string;
+  migratorOptions?: RdbmsSessionStorageMigratorOptions;
+}
+
+/**
+ * Define a common way for migrator to execute query on the underlying persistence layer
+ */
+export interface DBConnection {
+  /** the table used to store sessions */
+  sessionDBIdentifier: string;
```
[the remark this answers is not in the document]
cquemin: This is used in both RDBM and the base class, arguably they could have different name. But I thought this covered both cases no?

packages/shopify-app-session-storage-redis/src/migrations.ts
```diff
@@ -1,21 +1,22 @@
-import {createClient} from 'redis';
 import {Session} from '@shopify/shopify-api';
 
-type RedisClient = ReturnType<typeof createClient>;
+import {RedisConnection} from './redis-connection';
+
+export const migrationMap = new Map([
+  ['migrateToVersion1_0_1', migrateToVersion1_0_1],
```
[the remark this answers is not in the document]
cquemin: I dont believe we can change the version name because this has already been released and if we change it the migrator will determine that tis migration has not happened. I put a comment in the code and change the name of the function at least.

packages/shopify-app-session-storage-redis/src/redis-connection.ts
```diff
@@ -0,0 +1,62 @@
+import {createClient} from 'redis';
+import {DBConnection} from '@shopify/shopify-app-session-storage';
+
+type RedisClient = ReturnType<typeof createClient>;
+
+export class RedisConnection implements DBConnection {
+  sessionDBIdentifier: string;
```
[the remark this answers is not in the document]
cquemin: See here I don't really like that for Redis we have the word `table`, it doesn't really make sense...
How about `sessionIdentifier` ?

packages/shopify-app-session-storage-redis/src/redis-connection.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  constructor(
+    private client: RedisClient,
+    prefixKey: string,
```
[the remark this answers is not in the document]
cquemin: That was my french brain confusing me. You are absolutely right. I did pause for a bit when naming tis one wondering exactly the same but I reached the wrong conclusion! Thanks for pointing that one out! :)

packages/shopify-app-session-storage-redis/src/redis-connection.ts
```diff
@@ -0,0 +1,62 @@
...
+      finalKey = this.generateFullKey(baseKey);
+    }
+
+    return this.client.get(finalKey);
+  }
+
+  generateFullKey(name: string): string {
+    return `${this.sessionDBIdentifier}_${name}`;
+  }
+}
```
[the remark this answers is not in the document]
cquemin: Good idea!

packages/shopify-app-session-storage/src/types.ts
```diff
@@ -0,0 +1,179 @@
...
+  constructor(message: string) {
+    super(message);
+  }
+}
+
+/**
+ * Use to initialise session storage migrators
+ */
+export interface SessionStorageMigratorOptions {
+  migrationDBIdentifier: string;
```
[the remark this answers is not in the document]
cquemin: Same comment around the word `table` here it means non RDMBS systems which don't use the notion of table will have to use this....

packages/shopify-app-session-storage/src/types.ts
```diff
@@ -0,0 +1,179 @@
...
+   * @param shop shop of the session(s) to return
+   */
+  findSessionsByShop(shop: string): Promise<Session[]>;
+}
+
+/**
+ * define the option required to instantiate an RDBMS session storage implementation
+ */
+export interface RdbmsSessionStorageOptions {
+  sessionDBIdentifier: string;
```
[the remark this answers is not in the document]
cquemin: Makes sense to rename it how it use to be.

packages/shopify-app-session-storage-redis/src/migrations.ts
```diff
@@ -1,21 +1,23 @@
-import {createClient} from 'redis';
 import {Session} from '@shopify/shopify-api';
 
-type RedisClient = ReturnType<typeof createClient>;
+import {RedisConnection} from './redis-connection';
+
+export const migrationMap = new Map([
+  // Thi migration name cannot be chnaged as it has already been released
```
mkevinosullivan, line 6:
```suggestion
  // This migration name cannot be changed as it has already been released
```

mkevinosullivan (approved):

PR #102, opened 2023-01-27
Add shopify-api package as a peerDependency to each package in this repo

mkevinosullivan:

In apps that use the packages in this repo, sometimes each package can end up with its own copy of `shopify-api` in a sub-`node_modules` directory ... this creates conflict with the `shopify-api` package in the apps `node_modules` directory, resulting in issues with `Session.isActive()` resulting from the conflict.

Setting `@shopify/shopify-api` as a `peerDependencies` entry in the respective `package.json` files, so that only one copy of `@shopify/shopify-api` will be installed and used in an app.

## Type of change

## Checklist

- not applicable ~I have added/updated tests for this change~

### DIFF
--- a/.changeset/hip-cooks-pump.md
+++ b/.changeset/hip-cooks-pump.md
@@ -1,5 +1,4 @@
 ---
-'@shopify/shopify-app-express': patch
 '@shopify/shopify-app-session-storage': patch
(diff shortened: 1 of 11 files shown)


cquemin (approved):
Good catch!

PR #113, opened 2023-02-13
Kos/make pg case sensitive

mkevinosullivan:

Fixes Shopify/shopify-api-js#460

Adding `"` around table and column names so that case sensitivity is preserved.

## Type of change

## Checklist


packages/shopify-app-session-storage-postgresql/src/migrations.ts
```diff
@@ -2,12 +2,60 @@ import {PostgresConnection} from './postgres-connection';
...
 export async function migrateScopeFieldToVarchar1024(
   connection: PostgresConnection,
 ): Promise<void> {
-  await connection.query(`ALTER TABLE ${connection.sessionDBIdentifier} 
-    ALTER COLUMN scope TYPE varchar(1024)`);
+  await connection.query(`ALTER TABLE "${connection.sessionDBIdentifier}"
+    ALTER COLUMN "scope" TYPE varchar(1024)`);
+}
+
+export async function migrateToCaseSensitivity(
```
cquemin, line 16: Should this be done within a transaction to avoid data loss?
Comment made from @paulomarg on my `sqlite` migration :)

cquemin (changes requested):
LGTM apart from a missing transaction on the migration.

packages/shopify-app-session-storage-postgresql/src/migrations.ts
```diff
@@ -43,19 +43,31 @@ export async function migrateToCaseSensitivity(
...
       ];
     }
   }
 
-  for (const query of queries) {
-    await connection.query(query);
+  if (queries.length !== 0) {
+    // wrap in a transaction
+    queries.unshift(`BEGIN`);
+    queries.push(`COMMIT`);
```
cquemin, line 61: `COMMIT` should be after you execute your queries. So after line 66.

cquemin (changes requested):
I think the `COMMIT` is not located in the right place.

cquemin (approved):
All good, sorry I mis read the `push` for an `unshift`

PR #128, opened 2023-02-22
Remove unused dev dependencies

mkevinosullivan:

The `package.json` files within the `packages` folder contained unused dev dependencies.

`shopify-app-express` included various DB libraries as dev dependencies, but they are no longer required. This commit removes them.

Also removed `ts-node` which was for prototype purposes.


cquemin (approved):
Very nice!

PR #138, opened 2023-02-24
MySQL migrator unable to detect already applied migrations

### LINKED ISSUE #132: Duplicate entry 'migrateScopeFieldToVarchar1024' for key 'PRIMARY'"
ssion-storage/build/cjs/rdbms-session-storage-migrator.js:31:5)
2023-02-23 19:29:36 | backend    |     at async MySqlSessionStorageMigrator.applyMigrations 
                                   (/Users/metehankasap/Desktop/wishlist-app-newzi/web/node_modules/@shopify/shopify-app-session-storage/build/cjs/abstract-migration-engine.js:28:9) {
2023-02-23 19:29:36 | backend    |   code: 'ER_DUP_ENTRY',
2023-02-23 19:29:36 | backend    |   errno: 1062,
2023-02-23 19:29:36 | backend    |   sql: '\n' +
2023-02-23 19:29:36 | backend    |     '          INSERT INTO shopify_sessions_migrations (migration_name)\n' +
2023-02-23 19:29:36 | backend    |     "          VALUES('migrateScopeFieldToVarchar1024');\n" +
2023-02-23 19:29:36 | backend    |     '        ',
2023-02-23 19:29:36 | backend    |   sqlState: '23000',
2023-02-23 19:29:36 | backend    |   sqlMessage: "Duplicate entry 'migrateScopeFieldToVarchar1024' for key 'PRIMARY'"
2023-02-23 19:29:36 | backend    | }
2023-02-23 19:29:36 | backend    |

cquemin:
MySQL migrators was failing to detect that a migration was already applied because the query results are different for MySQL compared to other RDBMS.

Fixes #132

In `MySqlSessionStorageMigrator` the method `hasMigrationBeenApplied` has been overridden so that query results can be parsed differently than it is done generically, in the parent class `RdbmsSessionStorageMigrator`, for all other Rdbms.
Also test cases have been added for all session storages with migrators in order to validate that one-time initializations executed only once when session storage are starting up.

Another changes for this PR is to remove the argument on `SessionStorageMigratorOptions`. Because it wasn't working properly if the parameter was used. And there was no management of order or migrations execution when both internal and external migrations were added. 
I have kept the migrations as internal arguments for each migrator. This way this allows developers to use their own migration tools independently of our internal migrations and they should work side by sides without the need for developers to be aware of our migrations and their orders of execution.

## Type of change

## Checklist


packages/shopify-app-session-storage/src/abstract-migration-engine.ts
```diff
@@ -15,15 +14,22 @@ export abstract class AbstractMigrationEngine<
...
 
-  constructor(db: ConnectionType, opts: Partial<MigratorOptionType> = {}) {
+  constructor(
+    db: ConnectionType,
+    opts: Partial<MigratorOptionType> = {},
+    migrations: MigrationOperation[],
+  ) {
     this.options = {...defaultSessionStorageMigratorOptions, ...opts};
     this.connection = db;
+    this.migrations = migrations;
```
paulomarg, line 26: Should we validate the migrations at this point?

packages/shopify-app-session-storage-postgresql/src/postgres-migrator.ts
```diff
@@ -9,8 +10,9 @@ export class PostgresSessionStorageMigrator extends RdbmsSessionStorageMigrator
   constructor(
     dbConnection: PostgresConnection,
     opts: Partial<RdbmsSessionStorageMigratorOptions> = {},
+    migrations: MigrationOperation[],
   ) {
-    super(dbConnection, opts);
+    super(dbConnection, opts, migrations);
   }
```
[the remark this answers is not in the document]
cquemin: It is defined in the `RdbmsSessionStorageMigrator`, its parent class and seems to be working as expected as the added validated that instantiating 2 session storage one after the other does not run the migration twice.

packages/shopify-app-session-storage/src/abstract-migration-engine.ts
cquemin: How so? What sort of validations can we do ?

paulomarg (approved):
🚢

PR #193, opened 2023-04-06
Use decodeURIComponent on relevant dbUrl fields

### LINKED ISSUE #163: MySQL Storage: Password with special character will always access denied
# Issue summary

I would like to raise the issue regarding password input, currently when we use complex password like `sql123#@~` some of the characters will not properly handled or decoded during the login process of the `mysql2` package ([code reference to that package](https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L710) there's no trace of `decodeURIComponent` for the password field) and it will always result to access denied. 

We do the `encodeURIComponent` for the password on this line of [code in src/mysql.ts#L33](https://github.com/Shopify/shopify-app-js/blob/main/packages/shopify-app-session-storage-mysql/src/mysql.ts#L33). Right now, my workaround is to provide a password without special characters but I know this is not the best idea for production use.

That being said, I would like to request if we can have an option for the mysql connection credential to provide it via object instead of [URL string only](https://github.com/Shopify/shopify-app-js/blob/main/packages/shopify-app-session-storage-mysql/src/mysql-connection.ts#L58), just like the main package `mysql2` provided option for `mysql.createConnection({...})`

- `@shopify/shopify-app-session-storage-mysql@1.1.1` package and version:
- Node version: v16.19.1
- Operating system: MacOS M1

```
MySQL error: Access denied for root@127.0.0.1
```

1. Use password content with character `#` or any special character. (For my case I used `#`)
2. Try to connect using any of the provided steps by `MySQLSessionStorage` class, by class argument url string or using `withCredentials`

mkevinosullivan:

Passwords with special characters are denied access due to constructor encoding the user name, password and database name in the dbUrl.

Fixes #163

This change decodes the user, password and database name fields before passing them to the underlying mysql connection.

## Type of change

## Checklist

- ~I have documented new APIs/updated the documentation for modified APIs (for public APIs)~ not applicable

### DIFF
--- a/.changeset/tender-fireants-rhyme.md
+++ b/.changeset/tender-fireants-rhyme.md
@@ -1,5 +1,6 @@
 ---
 '@shopify/shopify-app-session-storage-mysql': patch
+'@shopify/shopify-app-session-storage-postgresql': patch
 ---
 
-Use decodeURIComponent on password, user, database name fields prior to calling underlying MySQL connection. Fixes #163
+Use decodeURIComponent on password, user, database name fields prior to calling underlying MySQL connection. Fixes #163. Also applied to postgresql adapter.


paulomarg (approved):

cquemin (approved):
If we have the same issue in other session storages I think we should address them as part of this PR as well.

PR #194, opened 2023-04-06
Add event handlers to redis client

### LINKED ISSUE #129: Improve RedisSessionStorage for Heroku-hosted apps
## Necessary Redis 6 + Heroku workaround isn't supported, results in app crashing on disconnect

For Redis 6 with Heroku apps, if certain event handlers aren't defined before connecting, the app will crash on disconnect. Heroku Redis instances disconnect/reconnect every 5 minutes so this results in a very unreliable server.

Here is a sample error log for this issue

```
Error: Socket closed unexpectedly
  File "/app/node_modules/@redis/client/dist/lib/client/socket.js", line 195, in TLSSocket.<anonymous>
    '{snip} lassPrivateFieldGet(this, _RedisSocket_instances, "m", _RedisSocket_onSocketError).call(this, new errors_1.SocketClosedUnexpectedlyError());
  File "node:events", line 628, in Object.onceWrapper
  File "node:events", line 525, in TLSSocket.emit
  File "node:domain", line 489, in TLSSocket.emit
  File "node:net", line 322, in <anonymous>
```

To prevent this behavior the known solution is to simply define these event handlers, even if no action is taken in the callback

```
- 'error'
- 'connect'
- 'reconnecting'
- 'ready'
```

[Here is an issue on the node-redis repo discussing implementation details.](https://github.com/redis/node-redis/issues/2032#issuecomment-1116883257)

I updated and built the package locally to fix an app that is running in production but would be great it there was a built in error handler on the class that we could pass a callback to. Happy to submit PR so we can use the public package going forward if you think this is worth implementing.

### LINKED ISSUE #160: Redis session storage 1.1.1 shuts down Redis connection
# Issue summary

Upgrading to `@shopify/shopify-app-session-storage-redis 1.1.1` in my Heroku production environment (I use sqlite session storage in development) caused my Heroku instance to start timing out requests (30s), never being able to serve my app.

Heroku logs show an `H12` error (just meaning request timeout). My error reporting tool (Rollbar) tells me it received `Error Socket closed unexpectedly` from Redis. I'm used to seeing that error log in _Heroku_ logs when I tail them, but based on my research I've understood it to be innocuous and that Redis just re-establishes its connection periodically. But I've never seen the error come through my error reporting tool (as if it's being thrown elsewhere or is somehow being caught/handled differently).

After a fresh deploy, I am sometimes able to load the app once or twice (enough to make me walk away unfortunately). But eventually no requests can be served (presumably the Redis connection is stuck closed).

💡 As I'm writing this, I actually wonder if the `onError` configuration option is being properly passed to the Redis client... that's the only explanation that's made sense so far (not having much insight into the session storage code). That would explain the closed connection, the error reporting difference, and intermittent behavior where it works for a short time but then stops working forever.

I've since reverted back to `1.0.2`

Env
- `@shopify/shopify-api` version: `6.2.0`
- Node version: `18`
- Operating system: `Heroku/Linux`

```
Error Socket closed unexpectedly
/app/node_modules/@redis/client/dist/lib/client/socket.js:181
.once('error', (err) => __classPrivateFieldGet(this, _RedisSocket_instances, "m", _RedisSocket_onSocketError).call(this, err))
.once('close', hadError => {
if (!hadError && __classPrivateFieldGet(this, _RedisSocket_isOpen, "f") && __classPrivateFieldGet(this, _RedisSocket_socket, "f") === socket) {
__classPrivateFieldGet(this, _RedisSocket_instances, "m", _RedisSocket_onSocketError).call(this, new errors_1.SocketClosedUnexpectedlyError());
}
})
.on('drain', () => {
```

## Expected behavior

The Redis connection stays open or at least re-connects, letting my app be served.

## Actual behavior

The Redis connection gets closed and ultimately blocks serving of my app.

1. Set up a Shopify app and a Heroku dyno with Heroku Redis with the below session storage configuration
2. Deploy to Heroku

```
const url = new URL(process.env.REDIS_TLS_URL) // or REDIS_URL if it's already TLS (just needs to start with `rediss:`)
const redisSessionStorage = new RedisSessionStorage(url, {
      url: url.toString(),
      socket: {
        tls: true,
        rejectUnauthorized: false,
      },
      onError(error) {
        console.error(error)
      },
    })
```

mkevinosullivan:

If certain event handlers aren't defined on the client before connecting, the app will crash on disconnect event (e.g., timeout).

Issue [comment](https://github.com/redis/node-redis/issues/2032#issuecomment-1116883257) regarding implementation details from node-redis repo.

Fixes #129
Fixes #160

Adding empty (non-acting) event handlers for `connect`, `reconnecting` and `ready` events, per the comment linked above.

## Type of change

## Checklist

### DIFF
--- a/.changeset/sixty-suns-brush.md
+++ b/.changeset/sixty-suns-brush.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-session-storage-redis': patch
+---
+
+Add event handlers to redis client to prevent crashing on disconnect event. Fixes #129, #160 (Thanks to @davidhollenbeckx for linking to issue and solution.)
--- a/packages/shopify-app-session-storage-redis/src/__tests__/redis.conf
+++ b/packages/shopify-app-session-storage-redis/src/__tests__/redis.conf
@@ -1,1 +1,2 @@
 user shopify on +@all allkeys  >passify
+timeout 2
--- a/packages/shopify-app-session-storage-redis/src/__tests__/redis.test.ts
+++ b/packages/shopify-app-session-storage-redis/src/__tests__/redis.test.ts
@@ -35,6 +35,10 @@ describe('RedisSessionStorage', () => {
     await wait(10000);
 
     client = createClient({url: dbURL.toString()});
+    client.on('error', (err) => {});
+    client.on('connect', () => {});
+    client.on('reconnecting', () => {});
+    client.on('ready', () => {});
     await client.connect();
   });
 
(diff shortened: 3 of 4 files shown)


paulomarg (approved):
Makes sense to me. Could we use jest timers to avoid the long-running test?

PR #205, opened 2023-04-13
Add redirectOutOfApp function

paulomarg:

Addresses https://github.com/Shopify/shopify-api-js/issues/749

Currently, there is no easy way for apps to redirect outside the scope of the app, because there are several ways for requests to come in, which puts the onus on developers to figure out the right mechanism of redirecting. Those mechanisms can sometimes be "hidden" in Shopify code so discovery isn't easy.

Adding a new `redirectOutOfApp` helper function that will detect what scenario we're in, and redirect out of the app using the appropriate mechanism.

## Type of change

## Checklist


abharvey (approved):
💸  Billing 🚀

PR #206, opened 2023-04-14
Allow URL string for new MySQLSessionStorage()

mkevinosullivan:
### WHY are these changes introduced?

In `1.0.2` and earlier of `MySQLSessionStorage`, the constructor could accept either a URL object or a URL string. The string option was accidentally removed, starting `1.1.0`.

Fixes #204

### WHAT is this pull request doing?

Allow `new MySQLSessionStorage()` to accept either a `URL` object or a string, e.g.,
```ts
const dbString = 'mysql://username:password@host/database';
const sessionStorage = new MySQLSessionStorage(dbString);
// OR
const sessionStorage = new MySQLSessionStorage(new URL(dbString));
```

## Type of change

- [x] Patch: Bug (non-breaking change which fixes an issue)

## Checklist

- [x] I have used `yarn changeset` to create a draft changelog entry (do NOT update the `CHANGELOG.md` files manually)
- [x] I have added/updated tests for this change
- [x] I have documented new APIs/updated the documentation for modified APIs (for public APIs)

PR #212, opened 2023-04-17
Allow URL string in redis, postgresql constructors

mkevinosullivan:

This aligns the constructor code with that of MySQL, even though the string-based constructor is not broken in Redis or PostgreSQL.

At the very least, it makes the code consistent and TypeScript correct.

Replication of #206.

## Type of change

- [x] Internal only

## Checklist


.changeset/empty-snakes-deny.md
```diff
@@ -0,0 +1,2 @@
+---
```
paulomarg, line 1: Should we say something here? Or did the mysql update cover this too?

paulomarg (approved):

PR #221, opened 2023-04-26
Remove Node 14, add Node 20

mkevinosullivan:

Updating CI to remove Node 14 (deprecated April 30, 2023), add Node 20

## Type of change

- [x] Internal only

## Checklist


.changeset/shaggy-clouds-invent.md
```diff
@@ -0,0 +1,2 @@
+---
```
cquemin, line 1: I wonder if this should warrant an actual changeSet message since we are changes the level of compatibility?
Probably good to have in the change logs rather needing to look into the code.

mkevinosullivan: I would say it's a patch release .. the only actual code changes relate to a test.  The library code itself doesn't change, so it technically would continue to work with Node 14

PR #264, opened 2023-05-18
Use new CLI env variables

paulomarg:

With https://github.com/Shopify/cli/pull/1789, the CLI will deprecate some env var names. While they will still be present so previous apps won't break, we should migrate new apps as soon as possible to the new format.

Using the new vars when available, but falling back to the current ones so apps don't break.

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/markdown_link_check.yml
+++ b/.github/workflows/markdown_link_check.yml
@@ -11,4 +11,4 @@ jobs:
         with:
           config-file: '.github/workflows/markdown_link_checker_config.json'
       - name: Ensure markdown links are only to main GitHub branches
-        run: 'grep --include="*.md" --exclude-dir=node_modules -Er "\/(tree|blob)" | (! grep -vE "\/(tree|blob)\/(main|master)")'
+        run: 'grep --include="*.md" --exclude-dir=node_modules -Er "\/(tree|blob)" | (! grep -vE "\/(tree|blob)\/(main|master|HEAD)")'


packages/shopify-app-express/src/__tests__/index.test.ts
```diff
@@ -36,6 +36,32 @@ describe('shopifyApp', () => {
   });
 
   it('properly defaults missing configs based on env vars', () => {
+    /* eslint-disable no-process-env */
+    process.env.SHOPIFY_APP_API_KEY = 'envKey';
+    process.env.SHOPIFY_APP_API_SECRET = 'envSecret';
+    process.env.SHOPIFY_APP_SCOPES = 'envScope1,envScope2';
+    process.env.SHOPIFY_APP_URL = 'https://envHost';
+    process.env.SHOP_CUSTOM_DOMAIN = '*.envCustomDomain';
```
byrichardpowell, line 44: I assume we provide some docs on the package so partners can know what env vars to provide.  Do we need to update these too?

byrichardpowell (approved):

packages/shopify-app-express/src/__tests__/index.test.ts
paulomarg: We're not using env vars in the github docs, they're just the defaults. We just use strings for these values [in the docs](https://github.com/Shopify/shopify-app-js/tree/main/packages/shopify-app-express#getting-started).

byrichardpowell: Ok, I still find the use of env vars in the package a bit confusing, but I think that's one for another time.

PR #340, opened 2023-07-12
Add gotchas section to the README

paulomarg:
There are some things that might trip up developers first coming into this template that can't be fixed from our side. This PR adds some notices and troubleshooting to the README to help guide folks through these issues.


.changeset/pink-readers-marry.md
```diff
@@ -0,0 +1,2 @@
+---
```
byrichardpowell, line 1: Is this needed?

paulomarg: Yes, every PR in this repo needs a changeset :)

PR #344, opened 2023-07-17
Update outdated docs

paulomarg:

We've changed how we're suggesting folks set up their exports in the `shopify.server` object in the template, and we should reflect that in the docs here to keep things consistent.

Updating the docs that include `shopify.server`.


byrichardpowell (approved):
:shipit:

PR #346, opened 2023-07-18
Allow App Bridge URL override

paulomarg:

Currently it isn't possible to test App Bridge locally with this package because it hardcodes the URL - which is fine since it will be evergreen in production.

Allowing an override env var to be set so that the internal references in this package can be overriden.


byrichardpowell (approved):
:shipit:

PR #347, opened 2023-07-18
Boundary headers export

paulomarg:

Currently, Remix apps will need a little bit of boilerplate code to ensure that we're handling the headers for responses we throw in some locations. Since that code depends on code that's internal to the package, we shouldn't put the burden on the app to handle it properly.

Exposing a new `boundary` item in the `shopifyApp` object which abstracts that boilerplate into the package.

While adding the new fields in `shopifyApp`, I noticed we weren't getting a squiggly for the unknown field, so I tightened up the types a little bit more.

### DIFF
--- a/packages/shopify-app-remix/README.md
+++ b/packages/shopify-app-remix/README.md
@@ -185,7 +185,8 @@ To load your app within the Shopify Admin, you need to:
 
 ## Authenticating admin requests
 
-`shopifyApp` provides methods for authenticating admin requests. To authenticate admin requests you can call `shopify.authenticate.admin(request)` in a loader or an action:
+`shopifyApp` provides methods for authenticating admin requests.
+To authenticate admin requests you can call `shopify.authenticate.admin(request)` in a loader or an action:
 
 ```ts
 // app/routes/**/*.tsx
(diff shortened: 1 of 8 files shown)


packages/shopify-app-remix/README.md
```diff
@@ -196,7 +197,30 @@ export const loader = async ({request}: LoaderArgs) => {
...
+
+### Headers
+
+It's important to note that the authentication functions in this package rely on throwing `Response` objects, which must be handled in your Remix routes using them.
+
+To do that, you can set up a [Remix `ErrorBoundary`](https://remix.run/docs/en/main/guides/errors).
+We provide some abstractions for the error and headers boundaries to make it easier for apps to set those up.
+
+```ts
+// app/routes/**/*.tsx
```
byrichardpowell, line 213: Please can we add the import to this example

packages/shopify-app-remix/src/boundary/__tests__/error.test.ts
```diff
@@ -0,0 +1,28 @@
+import {shopifyApp} from '../../index';
+import {testConfig} from '../../__tests__/test-helper';
+
+describe('Error boundary', () => {
+  it('returns a string when handling an ErrorResponse', () => {
```
byrichardpowell, line 5: What is it about returning a string that gets us the behaviour we want? I think there is something I'm not understanding here. Happy to chat on tuple

packages/shopify-app-remix/src/boundary/__tests__/headers.test.ts
```diff
@@ -0,0 +1,65 @@
+import {shopifyApp} from '../../index';
+import {testConfig} from '../../__tests__/test-helper';
+
+describe('Headers boundary', () => {
+  it('ignores non-error headers if errors are present', () => {
```
byrichardpowell, line 5: Suggest "returns only error headers if error headers are present"

packages/shopify-app-remix/src/boundary/__tests__/headers.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    const result = shopify.boundary.headers(headers);
+
+    // THEN
+    expect(result.get('parent')).toBeNull();
+    expect(result.get('loader')).toBeNull();
+    expect(result.get('action')).toBeNull();
+    expect(result.get('error')).toEqual('header');
+  });
+
+  it('merges all headers if no errors are present', () => {
```
byrichardpowell, line 25: Suggest "merged parent, loader & action headers if no error headers are present"

packages/shopify-app-remix/src/index.ts
```diff
@@ -94,16 +103,34 @@ export function shopifyApp<
...
       >(params),
     },
-    ...(isAdminApp(appConfig.distribution)
-      ? {}
-      : {login: loginFactory(params)}),
-  } as ShopifyApp<Config>;
+    boundary: {
+      error: errorBoundary,
+      headers: headersBoundary,
+    },
```
byrichardpowell, line 109: Why return this from the `shopify` object and not a separate export? Is this just future proofing? I guess that makes sense :+1:

byrichardpowell (approved):
Few minor changes, but otherwise :shipit:

PR #355, opened 2023-07-20
Add remix package

paulomarg:
This PR merges the Remix package into main in preparation for its v1 release 🎉


byrichardpowell (approved):
Approving so you are not blocked, but we might want to check version numbers.

PR #361, opened 2023-07-21
Add RemixPolarisLink component and components export

paulomarg:

Currently, embedded apps can't use Polaris `Link`s because client-side routing only works on remix apps when using the Remix `Link` component.

Since Polaris lets us override what component it uses to render links, we can underpin them with Remix links. However, their APIs are not the same so we need a component to translate between them.

## Type of change

## Checklist

### DIFF
--- a/.changeset/forty-steaks-scream.md
+++ b/.changeset/forty-steaks-scream.md
@@ -3,3 +3,62 @@
 ---
 
 Adding `AppProvider` component to abstract Shopify-specific app setup on the frontend side.
+This makes it easier for apps to set up the components it needs to work with Shopify.
+
+To make use of this in the Remix app template, you can update your `app/routes/app.jsx` file's `App` component from
+
+```ts
+export default function App() {
+  const {apiKey, polarisTranslations} = useLoaderData();
+
+  return (
+    <>
+      <script
+        src="https://cdn.shopify.com/shopifycloud/app-bridge.js"
+        data-api-key={apiKey}
+      />
+      <ui-nav-menu>
+        <Link to="/app" rel="home">
+          Home
+        </Link>
+        <Link to="/app/additional">Additional page</Link>
+      </ui-nav-menu>
+      <PolarisAppProvider
+        i18n={polarisTranslations}
+        linkComponent={RemixPolarisLink}
+      >
+        <Outlet />
+      </PolarisAppProvider>
+    </>
+  );
+}
+
+/** @type {any} */
+const RemixPolarisLink = React.forwardRef((/** @type {any} */ props, ref) => (
+  <Link {...props} to={props.url ?? props.to} ref={ref}>
+    {props.children}
+  </Link>
+));
+```
+
+to
+
+```ts
+import {AppProvider} from '@shopify/shopify-app-remix/react';
+
+export default function App() {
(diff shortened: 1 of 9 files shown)


packages/shopify-app-remix/src/components/RemixPolarisLink.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import {Link, LinkProps as RemixLinkProps} from '@remix-run/react';
+// Polaris' LinkProps doesn't match this type, so we need to import it directly
+import {LinkLikeComponentProps} from '@shopify/polaris/build/ts/src/utilities/link';
```
developit, line 4:
```suggestion
import type {LinkLikeComponentProps} from '@shopify/polaris/build/ts/src/utilities/link';
```

packages/shopify-app-remix/src/components/RemixPolarisLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+  if (target) {
+    remixProps.target = target;
+  }
+
+  return (
+    <Link ref={ref as React.Ref<HTMLAnchorElement>} {...remixProps} {...rest}>
+      {props.children}
+    </Link>
+  );
+}
```
developit, line 27: Almost none of this should be needed. Also, `props.ref` won't be there, you have to use forwardRef to obtain the outer ref from a function component.

```suggestion
export const RemixPolarisLink = forwardRef<HTMLAnchorElement, LinkLikeComponentProps>((props, ref) => {
  return <Link {...props} to={props.url ?? props.to} ref={ref} />;
})
```

paulomarg: I added the explicit checks because I wasn't sure if explicitly setting the `to` property would work (it's a mandatory prop), and the `onClick` handlers for them don't match up, so we need to have the intermediate function. `target` can probably be dropped though.

developit (approved):

byrichardpowell (approved):
I think we need to provide a little more capabilities on to the Provider re Polaris props.

packages/shopify-app-remix/package.json
```diff
@@ -61,9 +63,15 @@
     "tslib": "^2.5.0"
   },
   "peerDependencies": {
+    "@shopify/polaris": "*",
     "@remix-run/react": "*",
     "react": "*"
   },
+  "peerDependenciesMeta": {
+    "@shopify/polaris": {
+      "optional": true
```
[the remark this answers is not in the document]
paulomarg: Yeah!

.changeset/forty-steaks-scream.md
```diff
@@ -0,0 +1,57 @@
...
+      />
+      <ui-nav-menu>
+        <Link to="/app" rel="home">
+          Home
+        </Link>
+        <Link to="/app/additional">Additional page</Link>
+      </ui-nav-menu>
+      <PolarisAppProvider
+        i18n={polarisTranslations}
+        linkComponent={RemixPolarisLink}
```
byrichardpowell, line 28: To make it clear you no longer need it, do we also want to show the link Component in this example?

byrichardpowell (approved):

PR #382, opened 2023-08-01
Remix redirect helper

paulomarg:

Currently, when redirecting an authenticated request, the app will need to be aware of all of the possible scenarios for embedded apps:
- Document load / SSR request (document)
- Bounce page request (document-loading XHR)
- Remix link navigation request (JSON XHR)

We need to make it easier for apps to navigate to other pages without needing to keep track of this context.

Adding a new `redirect` helper that will look something like this:

```ts
export const loader = async ({ request }) => {
  const { session, redirect } = await authenticate.admin(request);

  const url = new URL(request.url);
  if (url.searchParams.get("redirectMe")) {
    return redirect("/app/additional");
  }

  return json({ shop: session.shop.replace(".myshopify.com", "") });
};
```

and be able to tell which scenario it is in to trigger the appropriate redirect response.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-remix/src/server/auth/admin/__tests__/exit-i-frame-path.test.ts
+++ b/packages/shopify-app-remix/src/server/auth/admin/__tests__/exit-i-frame-path.test.ts
@@ -32,7 +32,7 @@ describe('authorize.admin exit iframe path', () => {
       `<script data-api-key="${config.apiKey}" src="${APP_BRIDGE_URL}"></script>`,
     );
     expect(responseText).toContain(
-      `<script>window.open("${decodeURIComponent(exitTo)}", "_top")</script>`,
+      `<script>window.open("${decodeURIComponent(exitTo)}/", "_top")</script>`,
     );
     expectDocumentRequestHeaders(response);
   });
@@ -61,7 +61,7 @@ describe('authorize.admin exit iframe path', () => {
       `<script data-api-key="${config.apiKey}" src="${APP_BRIDGE_URL}"></script>`,
     );
     expect(responseText).toContain(
-      `<script>window.open("${decodeURIComponent(exitTo)}", "_top")</script>`,
+      `<script>window.open("${decodeURIComponent(exitTo)}/", "_top")</script>`,
     );
     expectDocumentRequestHeaders(response);
   });
--- a/packages/shopify-app-remix/src/server/auth/admin/types.ts
+++ b/packages/shopify-app-remix/src/server/auth/admin/types.ts
@@ -86,26 +86,6 @@ interface AdminContextInternal<
    * ```
    */
   cors: EnsureCORSFunction;
-
-  /**
-   * A function that redirects the user to a new page, ensuring that the appropriate parameters are set for embedded
-   * apps
-   *
-   * @example
-   * Redirecting the user to the app's home page
-   * ```ts
-   * // app/routes/admin/widgets.ts
-   * import { LoaderArgs, json } from "@remix-run/node";
-   * import { authenticate } from "../shopify.server";
-   * import { getWidgets } from "~/db/widgets.server";
-   *
-   * export const loader = async ({ request }: LoaderArgs) => {
-   *   const { session, redirect } = await authenticate.admin(request);
-   *   return redirect("/");
-   * };
-   * ```
-   */
-  redirect: RedirectFunction;
 }
 
 export interface EmbeddedAdminContext<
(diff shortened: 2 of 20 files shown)


packages/shopify-app-remix/src/auth/helpers/redirect.ts
```diff
@@ -0,0 +1,58 @@
...
+  TypedResponse,
+  redirect as remixRedirect,
+} from '@remix-run/server-runtime';
+
+import {BasicParams} from '../../types';
+
+import {renderAppBridge} from './render-app-bridge';
+import {getSessionTokenHeader} from './get-session-token-header';
+
+export type RedirectTarget = '_self' | '_blank' | '_parent' | '_top';
```
developit, line 11: Seems like we wouldn't want folks to use `_blank`, since it'll likely be blocked by popup blockers. `_parent` and `_top` are identical codepaths and both just trigger AB Remote Redirect. That leaves `_self`, which seems like it's identical to a normal Remix `redirect()`.

Maybe it would make sense to only provide a helper that redirects Admin, and leave normal HTTP redirects to Remix?


```js
import {redirect} from '@remix-run/server-runtime';

export function loader({request}) {
  const admin = authenticate.admin(request);

  // normal redirects: (for outside of Admin or in-iframe redirects)
  throw redirect('/foo');

  // what's provided here: (for redirecting Admin)
  throw admin.redirect('/foo');
}
```

packages/shopify-app-remix/src/server/auth/admin/__tests__/exit-i-frame-path.test.ts
```diff
@@ -32,7 +32,7 @@ describe('authorize.admin exit iframe path', () => {
       `<script data-api-key="${config.apiKey}" src="${APP_BRIDGE_URL}"></script>`,
     );
     expect(responseText).toContain(
-      `<script>window.open("${decodeURIComponent(exitTo)}", "_top")</script>`,
+      `<script>window.open("${decodeURIComponent(exitTo)}/", "_top")</script>`,
```
byrichardpowell, line 35: Why the addition of the "/" here? Is this because we are passing "/" instead of "" ?

packages/shopify-app-remix/src/server/auth/helpers/redirect-with-exitiframe.ts
```diff
@@ -5,6 +5,7 @@ import type {BasicParams} from '../../types';
 export function redirectWithExitIframe(
   params: BasicParams,
   request: Request,
+  redirectPath: string,
```
byrichardpowell, line 8: Why this addition? From what I can see we always pass `config.auth.path` to this argument, which is what the function body used to hard code anyway?

byrichardpowell (commented):
Part way through reviewing.  Leaving comments now so you don't need to ait for the full review before actioning.

packages/shopify-app-remix/src/server/auth/admin/__tests__/exit-i-frame-path.test.ts
paulomarg: This is just because I started using an `URL` object to build this in `render-app-bridge`, and it automatically adds the `/`.

packages/shopify-app-remix/src/server/auth/admin/types.ts
```diff
@@ -85,6 +86,26 @@ interface AdminContextInternal<
...
+   * import { authenticate } from "../shopify.server";
+   * import { getWidgets } from "~/db/widgets.server";
+   *
+   * export const loader = async ({ request }: LoaderArgs) => {
+   *   const { session, redirect } = await authenticate.admin(request);
+   *   return redirect("/");
+   * };
+   * ```
+   */
+  redirect: RedirectFunction;
```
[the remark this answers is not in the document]
paulomarg: It is not. I'll remove it from the context for non-embedded apps.

packages/shopify-app-remix/src/server/auth/helpers/redirect.ts
```diff
@@ -0,0 +1,77 @@
...
+        throw redirectWithAppBridgeHeaders(parsedUrl.toString());
+      } else if (embeddedSearchParam) {
+        throw renderAppBridge(params, request, {
+          url: parsedUrl.toString(),
+          target,
+        });
+      }
+    }
+
+    return remixRedirect(url, init);
```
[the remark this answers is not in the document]
paulomarg: We can just remove the check since this makes no sense for non-embedded apps.

packages/shopify-app-remix/src/server/auth/helpers/redirect.ts
```diff
@@ -0,0 +1,77 @@
...
+  params: BasicParams,
+  request: Request,
+): RedirectFunction {
+  const {config} = params;
+
+  return function redirect(url, init: RedirectInit) {
+    if (config.isEmbeddedApp) {
+      const {searchParams} = new URL(request.url);
+      const parsedUrl = new URL(url, config.appUrl);
+      const target = (typeof init !== 'number' && init?.target) || '_self';
```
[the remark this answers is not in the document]
paulomarg: Good callout!

byrichardpowell (approved):
:shipit: 

Once we have everything in [feature_react_exports](https://github.com/Shopify/shopify-app-js/tree/feature_react_exports) let's test everything there.

PR #385, opened 2023-08-02
Add separate export path for frontend-specific features

paulomarg:

We'll want to include some frontend-specific features in this package in the future, and currently the code structure is centered around server-side-only code, which makes it difficult to find where to put that new code as we add it.

This PR is simply moving things around into separate `server` and `react` folders and adding an ESLint rule to ensure we're not cross-importing files

## Type of change

## Checklist

### DIFF
--- a/.changeset/chatty-clouds-sparkle.md
+++ b/.changeset/chatty-clouds-sparkle.md
@@ -26,7 +26,7 @@ import shopify from '../../shopify.server';
 
 export async function loader({request}) {
   const shop = await authenticateExternalRequest(request);
-  const {admin} = await shopify.unauthenticated.admin(shop);
+  const {admin, session} = await shopify.unauthenticated.admin(shop);
 
   return json(await admin.rest.resources.Product.count({session}));
 }
--- a/packages/shopify-app-remix/src/server/__tests__/request-mock.ts
+++ b/packages/shopify-app-remix/src/server/__tests__/request-mock.ts
@@ -92,6 +92,9 @@ export async function validateMocks() {
     if (request?.headers) {
       expected.headers = {};
       actual.headers = {};
+
+      // eslint-disable-next-line no-warning-comments
+      // TODO: we're currently not checking the headers properly. We should fix this.
       Object.entries(request.headers).forEach(([key, value]) => {
         expected.headers[key] = value;
         actual.headers[key] = (init?.headers as any)[key];
--- a/packages/shopify-app-remix/src/server/__tests__/shopify-app.test.ts
+++ b/packages/shopify-app-remix/src/server/__tests__/shopify-app.test.ts
@@ -12,10 +12,7 @@ import {
 } from '../index';
 import {AppConfigArg} from '../config-types';
 
-import {TEST_SHOP, setUpValidSession, testConfig} from './test-helper';
-import {mockExternalRequest} from './request-mock';
-
-const REQUEST_URL = `https://${TEST_SHOP}/admin/api/${shopifyApiPackage.LATEST_API_VERSION}/customers.json`;
+import {testConfig} from './test-helper';
 
 describe('shopifyApp', () => {
   /* eslint-disable no-process-env */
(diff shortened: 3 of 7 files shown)


.changeset/lovely-swans-repeat.md
```diff
@@ -0,0 +1,10 @@
+---
+'@shopify/shopify-app-remix': minor
+---
+
+- Internally rearranged source files to create a better separation between backend and frontend code, so we can add frontend-specific exports in the future.
+- A new export path `@shopify/shopify-app-remix/react` will now contain those frontend exports.
```
byrichardpowell, line 6: Would only ship this change once we have something to put in the react export?  I'm in two minds here.  Shipping often vs shipping a large changelog that doesn't contain much the partner needs to action.

.gitignore
```diff
@@ -7,10 +7,7 @@ coverage/
 tsdocs/
 tsconfig.tsbuildinfo
 database.sqlite
-packages/**/*.d.ts
-packages/**/*.d.ts.map
-packages/**/*.js
-packages/**/*.js.map
+packages/**/build
```
byrichardpowell, line 10: For my own understanding, why this change?

packages/shopify-app-remix/package.json
```diff
@@ -12,15 +12,18 @@
...
   "exports": {
     ".": {
-      "import": "./build/cjs/index.js",
-      "require": "./build/cjs/index.js",
-      "types": "./build/ts/index.d.ts"
+      "types": "./build/ts/server/index.d.ts",
+      "default": "./build/cjs/server/index.js"
     },
-    "./adapters/*": "./build/cjs/adapters/*/index.js"
+    "./adapters/*": "./build/cjs/adapters/*/index.js",
```
byrichardpowell, line 22: Should adaptors be provided by the server export and aliased?

packages/shopify-app-remix/src/__tests__/test-helper.ts
```diff
@@ -33,203 +32,3 @@ export function testConfig(
...
-    expect(searchParams.get('host')).toBe(host);
-  }
-}
-
-export function expectLoginRedirect(response: Response) {
-  expect(response.status).toBe(302);
-
-  const {pathname} = new URL(response.headers.get('location')!, APP_URL);
-  expect(pathname).toBe('/auth/login');
-}
```
byrichardpowell, line 235: Is the thinking here that `/src/__tests__/test-helper.ts` is for shared test helpers whereas `/src/server/__tests__test-helper` is for server specific stuff? If so, I like it the thinking... I'm just wondering when `testConfig` would be used by anythign other than server specific stuff?

The React stuff we talked about doesn't need the test config for example

packages/shopify-app-remix/src/react/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  rules: {
+    'no-restricted-imports': [
+      'error',
+      {
+        patterns: [
+          {
+            group: ['../server/*'],
+            message:
+              'Importing from src/server in src/react is not allowed as they must be kept separate',
```
byrichardpowell, line 10: Worthwhile explaining why? E.g: It will blow up Remix due to different runtimes.

packages/shopify-app-remix/src/server/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  rules: {
+    'no-restricted-imports': [
+      'error',
+      {
+        patterns: [
+          {
+            group: ['../react/*'],
+            message:
+              'Importing from src/react in src/server is not allowed as they must be kept separate',
```
byrichardpowell, line 10: Worthwhile explaining why? E.g: It will blow up Remix due to different runtimes.

tsconfig.base.json
```diff
@@ -16,6 +16,7 @@
     "noUnusedLocals": true,
     "importHelpers": true,
     "strictNullChecks": true,
+    "stripInternal": true,
```
byrichardpowell, line 19: Why this change? I missed anything in this PR that needs this.

byrichardpowell (commented):
Much prefer this organisation.

My only other question is on the benefits of something like

```
import {...} from "@shopify/shopify-app-remix/react
import {...} from "@shopify/shopify-app-remix/server
import "@shopify/shopify-app-remix/server/adaptors/node
```

Then we always make it explicit if it's a server or react export.

.gitignore
paulomarg: It's just fewer things to keep in mind - everything we build goes into `build`, so we can just blanket ignore the folder as a whole, which allows us to have js files elsewhere if we ever need them.

.changeset/lovely-swans-repeat.md
paulomarg: I have a different PR coming with the `AppProvider` that will be exported out of this path. We can fix up the changelog once everything is combined to ensure it's reasonable.

packages/shopify-app-remix/package.json
paulomarg: I'd rather keep them fully separate because that makes it harder to accidentally import them and break apps.

packages/shopify-app-remix/src/__tests__/test-helper.ts
paulomarg: Yes, that was the thinking, but that's a great point about the config - we shouldn't ever need that for react stuff. I'll move that over and delete the file if it's empty.

tsconfig.base.json
paulomarg: We have one field somewhere that's marked as `@internal`, this removes it from the exported package types. I think that comes with the AppProvider component though, I broke the PR up in two and this landed in the wrong place. We'll still need it though.

paulomarg:
> Much prefer this organisation.
> 
> My only other question is on the benefits of something like
> 
> ```
> import {...} from "@shopify/shopify-app-remix/react
> import {...} from "@shopify/shopify-app-remix/server
> import "@shopify/shopify-app-remix/server/adaptors/node
> ```
> 
> Then we always make it explicit if it's a server or react export.

No objections, but we will still have to alias `server` to `.` so it may be more confusing than helpful.

byrichardpowell:
> No objections, but we will still have to alias server to . so it may be more confusing than helpful.

Yeah, good point on the aliasing.  Long term, I think aliasing allows us to deprecate `.` in version 2, whereas without this move we are stuck with `.` for ever.

Worth making this change and keeping a log of things that should be removed in version 2?

byrichardpowell (approved):

PR #407, opened 2023-08-18
[WIP] Unauthenticated admin client

byrichardpowell:
We've had multiple requests to be able to instantiate admin clients outside of requests from Shopify. 

This PR introduces a new API:

```
const shopify = shopifyApp(config)
shopify.unauthenticated.admin(shop)
```

This allows an app to get a shop from whatever means , and then get an Admin API client for that Request.  Possible use cases:

1. Background jobs
2. External Requests from non Shopify surfaces.

For example:

```
import shopify from "../shopify.server.js"
import {customAuthentication} from "../helpers"

export loader({request}) {
   const {shop} = customAuthentication(request)
   const {admin} = shopify.unauthenticated.admin(shop)
}
```

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-remix/src/server/auth/helpers/redirect.ts
+++ b/packages/shopify-app-remix/src/server/auth/helpers/redirect.ts
@@ -68,6 +68,7 @@ function isBounceRequest(request: Request) {
 function isDataRequest(request: Request) {
   const {searchParams} = new URL(request.url);
 
+  const isGet = request.method === 'GET';
   const sessionTokenHeader = Boolean(getSessionTokenHeader(request));
   const sessionTokenSearchParam = searchParams.has('id_token');
 
@@ -75,7 +76,7 @@ function isDataRequest(request: Request) {
     sessionTokenHeader &&
     !sessionTokenSearchParam &&
     !isBounceRequest(request) &&
-    !isEmbeddedRequest(request)
+    (!isEmbeddedRequest(request) || !isGet)
   );
 }
 


byrichardpowell:
@paulomarg I think we made a mistake in the previous API design.  Observe the differences here:

Webhook:
```
const { admin } = await authenticate.webhook(request);

admin.rest.get
admin.rest.AbandonedCheckout
```

Admin:
```
const { admin } = await authenticate.admin(request);

admin.rest.get;
admin.rest.resources.AbandonedCheckout;
```

I believe long term we want `admin.rest.resources.AbandonedCheckout` everywhere, but that would be a breaking change.  I'm thinking we log the issue as a V2 change, and accept the inconsistency for now?

packages/shopify-app-remix/src/server/clients/admin/graphql.ts
```diff
@@ -55,12 +46,11 @@ export function graphqlClientFactory({
...
-        params: {api, config, logger},
-        request,
-        error,
-        shop: session.shop,
-      });
+      if (handleClientError) {
+        throw await handleClientError({error, params, session});
+      }
+
+      throw new Error(error);
```
paulomarg, line 53: I think it might mess up the stack trace if we throw a new error at this point - would it be possible to re-throw `error`?

packages/shopify-app-remix/src/server/clients/admin/index.ts
paulomarg, line 1: Nit: to keep the pattern of having index.ts files be proper indexes, should we create a factory.ts file for this function? Is that overkill?

packages/shopify-app-remix/src/server/shopify-app.ts
```diff
@@ -93,6 +97,17 @@ export function shopifyApp<
   return shopify as ShopifyApp<Config>;
 }
 
+function unauthenticatedAdminContextFactory<
```
paulomarg, line 100: Wondering if it's too much to create a top-level `unauthenticated` folder for this kind of thing, since we're planning on adding more of them. We might want to punt extracting it until we actually add those methods, though.

paulomarg (commented):
I know it's WIP, but just wanted to add some thoughts :) - LGTM so far!

packages/shopify-app-remix/src/server/clients/admin/graphql.ts
byrichardpowell: Fixed in `reset-unauth-admin-client`

packages/shopify-app-remix/src/server/clients/admin/index.ts
byrichardpowell: Fixed in `reset-unauth-admin-client`

PR #408, opened 2023-08-18
Reorganise files

byrichardpowell:

It's not clear which files are used where.  This PR applies a bit more of a system

1. Move billing folder to `auth/admin` since it's only returned by that API
2. Add a helpers folder to `auth/admin`.  Move helpers that are only used by `auth/admin` to this folder.

How to review this:

1. Does this system make sense?
2. Did I miss anything that should also be moved?

## Type of change

## Checklist


paulomarg (approved):
This makes sense to me. The billing helper works with the Admin API so there's no point trying to reuse it anywhere else. Can't think of any other code that we might want to move.

byrichardpowell:
> Can't think of any other code that we might want to move.

The only one I could think of was `billing.require`, but I think we can let the use case prove itself before adding it.

PR #416, opened 2023-08-23
Adding documentation to remix package

paulomarg:
This PR begins to add documentation for the `shopify-app-remix` package, by adding reference pages for the objects exposed by the package.


packages/shopify-app-remix/src/server/types.ts
```diff
@@ -101,6 +97,166 @@ type SessionStorageType<Config extends AppConfigArg> =
...
+   *     case "CUSTOMERS_REDACT":
+   *     case "SHOP_REDACT":
+   *     default:
+   *       throw new Response("Unhandled webhook topic", { status: 404 });
+   *   }
+   *
+   *   throw new Response();
+   * };
+   * ```
+   */
```
byrichardpowell, line 212: Anyway we can pull all these examples into the docs?  Theres a lot of detail in the intellisense docs that isn't in the shopify.dev docs.

packages/shopify-app-remix/src/react/components/AppProvider/AppProvider.doc.example.tsx
```diff
@@ -0,0 +1,17 @@
+import {AppProvider} from '@shopify/shopify-app-remix/react';
+
+export default function App() {
+  return (
+    <AppProvider isEmbeddedApp apiKey="<api-key-from-partners-dashboard>">
+      {/* You can use App Bridge components within the provider */}
```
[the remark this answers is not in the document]
paulomarg: That's not strictly true because app bridge needs to be the first element in the body, but yeah, good point!

packages/shopify-app-remix/docs/generated/generated_static_pages.json
```diff
@@ -0,0 +1,135 @@
...
+              "title": "pnpm",
+              "code": "pnpm add @shopify/shopify-app-remix\n"
+            }
+          ]
+        }
+      },
+      {
+        "type": "Generic",
+        "anchorLink": "backend",
+        "title": "Set up your backend",
```
lizkenyon, line 41: Should we split this into two steps 1) Create the Shopify Object, 2) create an Auth route. Then we would have room to add a few more details to each.

packages/shopify-app-remix/docs/generated/generated_static_pages.json
```diff
@@ -0,0 +1,135 @@
+[
+  {
+    "id": "shopify-app-remix",
+    "title": "Shopify App package for Remix",
```
lizkenyon, line 4: I think what is tripping me up about this currently it feels like the expectation is the reader is following along like a tutorial, instead of each section being more standalone, where the reader is learning about concept possibly in isolation.

lizkenyon: Similar to the Checkout UI extension [references](https://shopify.dev/docs/api/checkout-ui-extensions)

byrichardpowell (approved):
Approving so you can ship once it's ready

Issue #418, opened 2023-08-24
session-storage-prisma Custom Model Name


TheRealFlyingCoder:
Pretty simple, I'd like to be able to pass an option for a different model name to the helper.

`const storage = new PrismaSessionStorage(prisma, 'shopifySession');`

Or alternatively just let the user pass through the delegate for their specific session model:

`const storage = new PrismaSessionStorage(prisma.shopifySession);`

The first is a non-breaking change by adding the optional `customModel?: keyof typeof T`

The second would be a breaking change but realistically be better DX

byrichardpowell:
Please see comment here: https://github.com/Shopify/shopify-app-js/pull/419#issuecomment-1701586502

PR #419, opened 2023-08-24
Prisma session storage - Quick boilerplate of adding custom model name to constructor

TheRealFlyingCoder:
Just putting this in as an idea of what could be done with regards to the enhancement I raised [here](https://github.com/Shopify/shopify-app-js/issues/418)

Let me know if you have any thoughts on the topic!

### DIFF
--- a/packages/shopify-app-session-storage-prisma/src/prisma.ts
+++ b/packages/shopify-app-session-storage-prisma/src/prisma.ts
@@ -7,14 +7,16 @@ export class PrismaSessionStorage<T extends PrismaClient>
   implements SessionStorage
 {
   private ready: Promise<any>;
+  private modelName: string;
 
-  constructor(private prisma: T) {
-    if (this.prisma.session === undefined) {
+  constructor(private prisma: T, private customModelName?: keyof typeof T) {
+    this.modelName = customModelName || 'session;
+    if (this.prisma[this.modelName] === undefined) {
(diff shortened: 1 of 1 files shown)


TheRealFlyingCoder:
> Hey @TheRealFlyingCoder
> 
> Thanks for opening a PR.
> 
> Please could you describe the use case why it's important to be able to customize the model name? I get what you've done here, I just wanna make sure this is a really important use case before we commit to maintaining the code.

Hey @byrichardpowell! 

In my scenario I am building an Ecommerce platform that ties into Shopify and other eCommerce providers, whilst also providing a public portal for users:

UserSession <- User logs into my public portal
ShopifyAppSession <-- Shopify app creates a session
WooCommerceSession <-- Example of a different integration requiring session management as well

Taking out the generic "Session" name makes it hyper specific to "Shopify app only" projects, assuming the project has no other session management requirements

paulomarg:
I think that's a fair ask - we should support using a table name other than `Session` for those cases.

In order to merge this though, we'll need a few more things such as adding tests and a changeset. If you'd like, we can take your changes and create a PR with them, but if you want to make them yourself we can review when we can.

TheRealFlyingCoder:
Hey @paulomarg,

Definitely prefer to leave it with your team, so you can make the decisions that best fit your patterns :+1:

I've already pulled out the code into a custom session storage to suit my needs, so i'm not waiting on this change :)

PR #431, opened 2023-09-06
Allow custom table names in Prisma schemas

### LINKED ISSUE #418: session-storage-prisma Custom Model Name

paulomarg:

Closes #418, #419

As proposed by @therealflyingcoder (thanks!), we should support table names other than `Session` in our prisma schema to give more flexibility to how apps can set up their DBs.

Adding an optional `Options` param to `PrismaSessionStorage`, so that apps can tweak the behaviour of the class. I ended up going with an object to make it more future-proof and easier to read when looking at the caller.

## Type of change

## Checklist

### DIFF
--- a/.changeset/green-pugs-matter.md
+++ b/.changeset/green-pugs-matter.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-session-storage-prisma': minor
+---
+
+Allow using custom tables for the Prisma session storage.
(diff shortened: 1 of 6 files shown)


byrichardpowell (approved):
Just the one comment about docs.

PR #437, opened 2023-09-11
update: shopify-app-session-storage-[mysql,postgresql] to respect SSL DB url param

### LINKED ISSUE #230: MySQLSessionStorage support for SSL/TLS: `server does not allow insecure connections, client must use SSL/TLS`
# Issue summary

I am creating a Shopify app with a MySQL Planetscale (https://planetscale.com/) database. The connection URL I have for looks like:

<pre>
DATABASE_URL='mysql://USERNAME:PASSWORD@HOST/TABLE?<b>ssl={"rejectUnauthorized":true}</b>'
</pre>

I'm passing that to the session storage constructor as it suggests in the documentation:
```
...
sessionStorage: new MySQLSessionStorage(DATABASE_URL),
...
```

But when I start my server locally, it looks like the SSL option isn't getting picked up:
```
 Error: unknown error: Code: UNAVAILABLE
  server does not allow insecure connections, client must use SSL/TLS
```
I've tried a number of different query parameters suggested in the planetscale repo to bypass this issue with no luck: 
```
?sslcert=/etc/ssl/cert.pem, ?ssl={}, sslaccept=strict,
```

Am I missing something on my side? I would expect that the ssl option would get picked up off the URL by MySQL2. I have verified that the URL is picked up correctly from my environment variables, and I have verified that I can connect using the same URL using their express.js example app: https://planetscale.com/docs/tutorials/connect-nodejs-app

- `@shopify/shopify-app-*` package and version: `"@shopify/shopify-app-express": "^2.0.0"`, `"@shopify/app": "3.45.1"`
- Node version: v20.0.0
- Operating system: macOS

## Expected behavior

Database URL with ssl query parameter should be passed to the session storage constructor,  a new session storage instance should be generated, and the app should make a connection to the DB.

## Actual behavior

Shopify app is unable to connect due to misconfigured SSL/TLS

Thanks for any help! Appreciate all the work that has gone into these tools-- much improved experience from even a year ago.

chr33s:

Fixes #230 

This updates the `shopify-app-session-storage-mysql` and `shopify-app-session-storage-postgresql` packages to respect SSL DB url param e.g. `mysql://root:password@127.0.0.1:3306?ssl=false`.

## Type of change

## Checklist


bitkidd:
Sorry, but I'd say that it is a bad PR, you completely ignore all the possible options that `pg-connection-string` can have, [link](https://github.com/brianc/node-postgres/tree/master/packages/pg-connection-string). 

The best option here is not to parse connection string and leave it be and then pass it into pool as `connectionString`.

chr33s:
@bitkidd agreed its not ideal, but I simply implemented the missing functionality using the fewest lines of code possible by reusing the existing architecture.

packages/shopify-app-session-storage-mysql/src/mysql-connection.ts
maruffahmed, line 1:
```javascript
(this.dbUrl.searchParams.get('ssl') !== 'false').toString()
```
In this case ```this.dbUrl.searchParams.get('ssl')``` will return either ```'{"rejectUnauthorized":true}'``` as a string or ```null``` if not found. I think the ```!=='false'``` check will not work here.

I would suggest the following solution :
```javascript
async init() {
    const sslConfig = this.dbUrl.searchParams.get('ssl');
    this.pool = await mysql__default["default"].createPool({
      connectionLimit: this.connectionPoolLimit,
      host: this.dbUrl.hostname,
      user: decodeURIComponent(this.dbUrl.username),
      password: decodeURIComponent(this.dbUrl.password),
      database: this.getDatabase(),
      port: Number(this.dbUrl.port),
      ssl : sslConfig ? JSON.parse(sslConfig) : undefined
    });
  }
```

Ref of ```ssl``` types:
```typescript
ssl?: string | SslOptions;
```

chr33s: @maruffahmed good catch, will update the pull request.

maruffahmed (approved):
I think it'll work.

packages/shopify-app-session-storage-mysql/src/mysql-connection.ts
```diff
@@ -73,6 +73,14 @@ export class MySqlConnection implements RdbmsConnection {
     return decodeURIComponent(this.dbUrl.pathname.slice(1));
   }
 
+  public getSSL(): string | mysql.SslOptions | undefined {
+    const ssl = this.dbUrl.searchParams.get('ssl');
+    if (!ssl) {
+      return undefined;
+    }
+    return JSON.parse(ssl);
```
maruffahmed, line 81: I would just do this:
```javascript
return ssl ? JSON.parse(ssl) : undefined;  
```

maruffahmed (approved):
I think it'll work now. Looking for a review from the code owner. 
@chr33s @bitkidd @macournoyer please take a look at this PR.

packages/shopify-app-session-storage-mysql/src/mysql-connection.ts
chr33s: @maruffahmed agreed my preference is also shorthand, updated.

chr33s:
@bitkidd this pull request is scoped to the ssl param, as that was my (& other's usecase, see linked issue). With the update `JSON.parse` approach as suggested by @maruffahmed this pull request now satisfies the original scope.

I think changing the initial design change warrants a new pull request?

maruffahmed:
@paulomarg please take a look at this PR.

scraper095:
Any updates or workarounds? I am facing the same issue

lizkenyon:
@chr33s can you please sign the CLA?

chr33s:
@lizkenyon I noticed the workflow run failed, I toggled the ssl param back to the default `?ssl={"rejectUnauthorized":true}`, which should resolve the issue.

reinisdy:
Hello, any updates on this issue?

chr33s:
@lizkenyon tests should now be passing, I did notice some `DrizzleSessionStoragePostgres` tests where failing but that's outside the scope of this PR.

PR #443, opened 2023-09-14
Allow all billing config overrides at request

paulomarg:

With https://github.com/Shopify/shopify-api-js/pull/985, the library will allow overriding all fields when requesting payment, not just `trialDays`.

Passing through all params to `billing.request`, rather than just `trialDays`.

## Type of change

## Checklist


byrichardpowell (approved):
Approving so you can ship without a second review.  But, should we be adding docs?

byrichardpowell:
:shipit: 👍

PR #447, opened 2023-09-20
Further improvements to Remix package docs

paulomarg:
This PR makes improvements to the docs for the remix package.


byrichardpowell (approved):
Just 2 comments, but approving so you can ship without a second review

**QuickStart**
Screenshot 2023-09-21 at 1 47 41 PM

In the quickstart guide could we just provide the CLI commands in yarn, npm and pnpm instead of linking to another tutorial, that links to another tutorial, that has the commands you need?

** Public App Proxy ** 
Screenshot 2023-09-21 at 1 56 03 PM

There are 2 sections for GraphQL: querying the GraphQL API.  Could the headigns make it clearer that one is for admin and one is for storefront?

paulomarg:
> In the quickstart guide could we just provide the CLI commands in yarn, npm and pnpm instead of linking to another tutorial, that links to another tutorial, that has the commands you need?

Fair. We should do that.

> There are 2 sections for GraphQL: querying the GraphQL API. Could the headigns make it clearer that one is for admin and one is for storefront?

Not easily, because I'm "cheating" here to pull examples from types that are not the one we're documenting here. I can improve this by adding some better examples to the app proxy type though!

PR #448, opened 2023-09-20
Token exchange prototype

rachel-carvalho:

Fixes #0000 

## Type of change

## Checklist


packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
```diff
@@ -0,0 +1,245 @@
...
+        sessionId,
+      );
+
+      if (persistedSession) {
+        logger.debug(`Reusing existing token: ${persistedSession.accessToken}`);
+
+        if (persistedSession.isScopeChanged(config.scopes)) {
+          config.sessionStorage.deleteSession(persistedSession.id);
+
+          this.redirectToInstall(request, shop);
```
rezaansyed, line 185: This is an edge case that can happen when the app changes scopes, while a merchant has an active app browser session. Please see the [discussion for more details](https://github.com/Shopify/develop-app-access/discussions/92).

packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  private handleInvalidSession(request: Request) {
+    const {api, logger, config} = this;
+
+    const isXhrRequest = request.headers.get('authorization');
+    if (!isXhrRequest) {
+      return redirectToBouncePage(new URL(request.url), {api, logger, config});
+    }
+
+    return new Response(undefined, {
```
rezaansyed, line 217: For an XHR request made via App Bridge's augmented fetch, return a `401` with the `X-Shopify-Invalid-Session` header.

packages/shopify-app-remix/src/server/authenticate/admin/helpers/handle-embedded-client-error.ts
```diff
@@ -0,0 +1,65 @@
...
+      {
+        code: error.response.code,
+        statusText: error.response.statusText,
+        body: JSON.stringify(error.response.body),
+      },
+    );
+
+    console.log('error.response', JSON.stringify(error.response));
+
+    if (error.response.code === 401) {
```
rezaansyed, line 38: When making Admin API requests, a `401` signifies that the access token is invalid. In this scenario, we want to perform token exchange. So we delete the access token and return a `401` with `X-Shopify-Invalid-Session` so that the augmented fetch can retry the request, consequently triggering token exchange.

packages/shopify-app-remix/src/server/authenticate/admin/helpers/handle-embedded-client-error.ts
```diff
@@ -0,0 +1,65 @@
...
+      // 401 unauthorized from core
+      // delete access token
+      params.config.sessionStorage.deleteSession(session.id);
+
+      const isXhrRequest = request.headers.get('authorization');
+      if (isXhrRequest) {
```
rezaansyed, line 44: In this scenario, we want to perform token exchange. So we delete the access token and return a `401` with `X-Shopify-Invalid-Session` so that the augmented fetch can retry the request, consequently triggering token exchange.

packages/shopify-app-remix/src/server/authenticate/admin/helpers/handle-embedded-client-error.ts
```diff
@@ -0,0 +1,65 @@
...
+        throw new Response(undefined, {
+          status: error.response.code,
+          headers: {
+            'Content-Type': 'application/json',
+            'X-Shopify-Invalid-Session': '1',
+          },
+        });
+      } else {
+        throw redirect(request.url);
```
rezaansyed, line 53: If an Admin API request is happening on a document load, we want to delete the access token and refresh the page in order to trigger token exchange.

packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
```diff
@@ -0,0 +1,245 @@
...
+    } else if (isExitIframe) {
+      const destination = url.searchParams.get('exitIframe')!;
+
+      logger.debug('Rendering exit iframe page', {destination});
+      throw renderAppBridge(params, request, {url: destination});
+    } else if (isAuthRequest) {
+      const shop = ensureValidShopParam(request, {api, logger, config});
+      const installUrl = `https://${shop}/admin/oauth/install?client_id=${config.apiKey}`;
+      throw redirect(installUrl);
+    }
```
byrichardpowell, line 113: Why do we need to handle auth requests in token exchange?

packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    const sessionTokenHeader = getSessionTokenHeader(request);
+    const sessionTokenParam = url.searchParams.get(SESSION_TOKEN_PARAM)!;
+    const sessionTokenString = sessionTokenHeader || sessionTokenParam;
+
+    logger.info('Authenticating admin request');
+
+    if (sessionTokenParam) {
+      await validateUrlParams(request, {api, logger, config});
```
byrichardpowell, line 122: Is validing the URL params overkill given?

1. I _think_ we can get all the same data from the session token
2. The session token can only be constructed using the API secret.
3. The URL params also use the API secret.

Worth checking with app sec? If so, we get to simplify our token exchange implementation.

packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
```diff
@@ -0,0 +1,245 @@
...
+      await ensureAppIsEmbeddedIfRequired(request, {api, logger, config});
+    }
+
+    if (sessionTokenString) {
+      try {
+        const sessionToken = await validateSessionTokenUncaught(
+          {api, logger, config},
+          sessionTokenString,
+        );
```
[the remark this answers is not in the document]
rezaansyed: We catch the invalid JWT error and either return response header `X-Shopify-Invalid-Session` for XHR requests or redirect to bounce page.

packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
rezaansyed: There is the ability on the browser to simply refresh the iFrame instead of the entire page. So after a minute if the merchant only reloads the iFrame, it'll use an expired token. We should avoid that as that same session token might be used to make a token exchange request (which fails).

packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    const sessionTokenHeader = getSessionTokenHeader(request);
+    const sessionTokenParam = url.searchParams.get(SESSION_TOKEN_PARAM)!;
+    const sessionTokenString = sessionTokenHeader || sessionTokenParam;
+
+    logger.info('Authenticating admin request');
+
+    if (sessionTokenParam) {
+      await validateUrlParams(request, {api, logger, config});
+      await ensureAppIsEmbeddedIfRequired(request, {api, logger, config});
```
byrichardpowell, line 123: This is interesting.

In auth code flow `ensureAppIsEmbeddedIfRequired` makes sense.  In token exchange `IfRequired` makes less sense, because it's always required.

When we product-ionize this code these are the kind of helpers that I'd want to be careful about.  I'd prefer to duplicate the helper and reduce coupling between authe code flow and token exchange

packages/shopify-app-remix/src/server/shopify-app.ts
```diff
@@ -75,7 +77,9 @@ export function shopifyApp<
     addDocumentResponseHeaders: addDocumentResponseHeadersFactory(params),
     registerWebhooks: registerWebhooksFactory(params),
     authenticate: {
-      admin: oauth.authenticateAdmin.bind(oauth),
+      admin: config.isEmbeddedApp
+        ? tokenExchange.authenticateAdmin.bind(tokenExchange)
+        : oauth.authenticateAdmin.bind(oauth),
```
byrichardpowell, line 82: Love this ❤️

packages/shopify-app-remix/src/server/authenticate/admin/embedded-authenticate.ts
rachel-carvalho: We're still handling the `/auth` route so that the login page where the developer can type their store url to initiate install could still work, since that's a nice tool during app development. 

Then instead of initiating auth code flow like the original `authenticate` strategy does, it just redirects the user to the shopify managed install page.

byrichardpowell: Thanks for the context 👍

packages/shopify-app-remix/src/server/authenticate/admin/helpers/handle-embedded-client-error.ts
rezaansyed: might be worth simply redirecting to `/session-token?url={request.url}` to allow the app to naturally perform token exchange without ever retrying the request.

PR #458, opened 2023-10-03
Simplify remix README file

paulomarg:

Now that we have documentation for the Remix package in [shopify.dev](https://shopify.dev/docs/api/shopify-app-remix), we shouldn't duplicate some of the content in the package README.

Changing the README to cover setting up the app / package with simpler examples, but linking out to the docs where relevant.

See https://github.com/Shopify/shopify-app-js/blob/simplify_remix_readme/packages/shopify-app-remix/README.md


byrichardpowell (approved):
:shipit:

PR #459, opened 2023-10-03
Remove node 14 support

paulomarg:
Node 14 has reached [EOL](https://endoflife.date/nodejs), and some dependencies we use no longer work with it.

Therefore, we have to drop support for that version in order to upgrade our dependencies.


byrichardpowell (approved):
:shipit: !

PR #460, opened 2023-10-04
[Feature] - Extend AppConfig To Allow Flexibility On Bot Checks

probably-not:

There are situations where the app developer does not have full control over the underlying infrastructure (including how User-Agent headers are passed to the service). In the current version, the `rejectBotRequest` function is run unconditionally. This is a major problem, since this assumes that the developer has full control over the deployment infrastructure, which is typically not the case.

This change adds two configurations:
- `extraAllowedUserAgents`: can be used to exclude specific user agents from the `isbot` check
- `skipBotCheck`: can be used to fully skip the bot check. This can be used in situations where the developer's underlying infrastructure already has bot checking at a higher layer in the stack, and the developer doesn't necessarily know what user agents will be passed through.

## Type of change

## Checklist

> I don't see any tests for this function? Tell me if there's something to add here
> Not sure which docs to change... please advise

### DIFF
--- a/packages/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
+++ b/packages/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
@@ -68,7 +68,9 @@ export class AuthStrategy<
   ): Promise<AdminContext<Config, Resources>> {
     const {api, logger, config} = this;
 
-    rejectBotRequest({api, logger, config}, request);
+    if (!config.skipBotCheck) {
+      rejectBotRequest({api, logger, config}, request);
+    }
     respondToOptionsRequest({api, logger, config}, request);
 
     const cors = ensureCORSHeadersFactory({api, logger, config}, request);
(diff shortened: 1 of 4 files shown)


probably-not:
I just signed the CLA, if someone can re-run the check, it should pass now

probably-not:
Hi @paulomarg, sorry for the delay in answering (been a bit busy on my end).

As a simple example that tripped my devs up - when using Amazon Cloudfront with an API Gateway and a Lambda answering API calls, if there is a misconfiguration somewhere along the line (i.e. one of the layers, the CDN, the Gateway not passing on all of the headers) then isbot will mark that the request is a bot no matter what (looks like all Amazon related user agents get marked from my tests).

Like I mentioned in the PR, this is a major problem when the developer is not in full control of the infrastructure. I have no problem if I just throw a machine online and point a domain name to it's public IP, but the second things get more involved (which is a certainty at companies with growing teams), then the dev has no control over the way the infrastructure reacts.

This, combined with the fact that the isbot blocking functionality is pretty much undocumented (I only came across it by looking through Github issues and the code itself), and the fact that there's 0 logging on why the request is being blocked (I even turned on full debug logs and got nothing in our application) makes this library impossible to work with as is.

byrichardpowell:
Hey 👋 

Thanks for wanting to contribute.  Since this PR had grown so stale I'm going to close this PR.

If this is still something you need then please rebase the PR or submit a new one.

PR #461, opened 2023-10-04
Future flags and webhook admin context update

paulomarg:

The format of `admin` in the return of `authenticate.webhook` is currently incorrect - it doesn't return the same `AdminApiContext` type we return elsewhere.

The main problem with updating that is that it is a breaking change. In order to avoid forcing developers to change their code to upgrade, we're introducing the concept of **future flags** to the `shopifyApp` configuration object.

Following the Remix standard, apps will be able to opt in to breaking changes ahead of time, which will enable us to continue making improvements while not breaking existing apps. Developers will also have time to update their apps before we make major releases.

The first flag we're introducing is `v3_webhookAdminContext`. This flag will change the output format of the `admin` context field in `authenticate.webhook`. If apps are using TS, they'll start getting type errors for this object as the return type is determined by the value of the flag in the config.

We're also including an example of each format to the documentation so that devs can more easily find the differences between the two.

## Type of change

## Checklist

### DIFF
--- a/.changeset/serious-nails-own.md
+++ b/.changeset/serious-nails-own.md
@@ -5,3 +5,63 @@
 Added support for `future` flags in the `shopifyApp` function, with a `v3_webhookContext` flag to have `authenticate.webhook` return a standard `admin` context, instead of a different type.
 
 Apps can opt in to the new future at any time, so this is not a breaking change (yet).
+
+<details>
+  <summary>See an example</summary>
+
+Without the `v3_webhookContext` flag, `graphql` provides a `query` function that takes the query string as the `data` param.
+When using variables, `data` needs to be an object containing `query` and `variables`.
+
+```ts
+import {json, ActionFunctionArgs} from '@remix-run/node';
+import {authenticate} from '../shopify.server';
+
+export async function action({request}: ActionFunctionArgs) {
+  const {admin} = await authenticate.webhook(request);
+
+  const response = await admin?.graphql.query<any>({
+    data: {
+      query: `#graphql
+      mutation populateProduct($input: ProductInput!) {
+        productCreate(input: $input) {
+          product {
+            id
+          }
+        }
+      }`,
+      variables: {input: {title: 'Product Name'}},
+    },
+  });
+
+  const productData = response?.body.data;
+  return json({data: productData.data});
+}
+```
+
+With the `v3_webhookContext` flag enabled, `graphql` _is_ a function that takes in the query string and an optional settings object, including `variables`.
+
+```ts
+import {ActionFunctionArgs} from '@remix-run/node';
+import {authenticate} from '../shopify.server';
+
+export async function action({request}: ActionFunctionArgs) {
+  const {admin} = await authenticate.webhook(request);
+
+  const response = await admin?.graphql(
+    `#graphql
+    mutation populateProduct($input: ProductInput!) {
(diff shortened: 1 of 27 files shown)


byrichardpowell:
We chatted in person.  2 suggestions:

1. Add a guide to shopify.dev explaining:
  - why we are using features flags 
  - what feature flags exist and what they effect
  - How to opt into feature flags
2. On shopify.dev, on the page for shopifyApp(), pull out the feature flags API so it's more obvious.

.changeset/serious-nails-own.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/shopify-app-remix': minor
+---
+
+Added support for `future` flags in the `shopifyApp` function, with a `v3_webhookContext` flag to have `authenticate.webhook` return a standard `admin` context, instead of a different type.
```
byrichardpowell, line 5: Worth adding webhook examples with the feature flag turned on and off?

packages/shopify-app-remix/src/server/__tests__/shopify-app.test.ts
```diff
@@ -45,10 +45,9 @@ describe('shopifyApp', () => {
 
   it('does not have login function when distribution is ShopifyAdmin', () => {
     // GIVEN
-    const shopify = shopifyApp({
-      ...testConfig(),
-      distribution: AppDistribution.ShopifyAdmin,
-    });
+    const shopify = shopifyApp(
+      testConfig({distribution: AppDistribution.ShopifyAdmin}),
```
byrichardpowell, line 49: 👍

packages/shopify-app-remix/src/server/authenticate/webhooks/authenticate.webhooks.doc.ts
```diff
@@ -2,7 +2,9 @@ import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
 
 const data: ReferenceEntityTemplateSchema = {
   name: 'Webhook',
-  description: 'Contains functions for verifying Shopify webhooks.',
+  description:
+    'Contains functions for verifying Shopify webhooks.' +
+    '\n\n> Note: The format of the `admin` object returned by this function changes with the `v3_webhookAdminContext` future flag. Learn more about [gradual feature adoption](/docs/api/shopify-app-remix/guide-future-flags).',
```
byrichardpowell, line 7: 👍

packages/shopify-app-remix/src/server/future/flags.ts
```diff
@@ -0,0 +1,19 @@
+export interface FutureFlags {
+  /**
+   * When enabled, returns the same `admin` context (`AdminApiContext`) from `authenticate.webhook` that is returned from `authenticate.admin`.
+   *
+   * @default false
+   */
+  v3_webhookAdminContext?: boolean;
```
byrichardpowell, line 7: Worth showing an example of with and without here?

byrichardpowell (approved):
Good work.  Approving so you can release without a second review!

packages/shopify-app-remix/src/server/future/flags.ts
paulomarg: I'll add a link to the actual docs so we don't have to maintain yet another copy of this set of examples!

PR #473, opened 2023-10-22
Fix Remix redirect on bounce

btomaj:

Remix v2 changed some instances of `ErrorResponse` to `ErrorResponseImpl`, breaking `const { redirect } = authenticate.admin(request);` redirects.

This change allows redirects to work on both `ErrorResponse` and `ErrorResponseImpl`.

## Type of change

## Checklist


btomaj:
@byrichardpowell FYI

btomaj:
@paulomarg I asked the Remix team about the difference between `ErrorResponse` and `ErrorResponseImpl`, and I was told that `ErrorResponse` is intended for stdout whereas `ErrorResponseImpl` is internal (meaning not for stdout). I'm not familiar enough with the inner workings of this codebase, so before we merge this in, I wanted to ask if this change from the Remix team requires a bigger change to this codebase?

paulomarg:
> @paulomarg I asked the Remix team about the difference between `ErrorResponse` and `ErrorResponseImpl`, and I was told that `ErrorResponse` is intended for stdout whereas `ErrorResponseImpl` is internal (meaning not for stdout). I'm not familiar enough with the inner workings of this codebase, so before we merge this in, I wanted to ask if this change from the Remix team requires a bigger change to error handling?

Thanks for chasing it down. As far as this codebase goes, this is the only place where we do this. We're trying to find a better way to handle the headers in these cases, so hopefully this code will be replaced by a "proper" check in the future.

Let's also leave the previous implementation around, as you said it helps with compatibility and shouldn't cause any problems in v2 if `ErrorResponse` never reaches this code.

PR #487, opened 2023-11-12
[App proxy and Unauthenticated storefront]: Code fixes

mcvinci:
## This PR: 
- Fixes an incorrectly referenced constant in the [App Proxy reference documentation](https://shopify.dev/docs/api/shopify-app-remix/v1/authenticate/public/app-proxy#storefront-example-group)
- Fixes some code that references `authenticated.storefront` versus `authenticate.public.appProxy(request)`  in the [Unauthenticated storefront documentation](https://shopify.dev/docs/api/shopify-app-remix/v1/unauthenticated/unauthenticated-storefront#example-graphql)
- Fixes https://github.com/Shopify/shopify-dev/issues/39079
- Fixes https://github.com/Shopify/shopify-dev/issues/38934


packages/shopify-app-remix/src/server/clients/storefront/types.ts
```diff
@@ -13,10 +13,10 @@ export interface StorefrontContext {
    * <description>Use `storefront.graphql` to make query / mutation requests.</description>
    * ```ts
    * import { ActionFunctionArgs } from "@remix-run/node";
-   * import { authenticate } from "../shopify.server";
+   * import { unauthenticated } from "../shopify.server";
```
paulomarg, line 16: I wonder if we should use `appProxy` as our "official" example here - we want to encourage folks to rely on the authenticated methods as much as we can.

paulomarg (approved):
LGTM, just had one question.

packages/shopify-app-remix/src/server/clients/storefront/types.ts
mcvinci: Thanks, @paulomarg! I just added the `appProxy` example. Would you mind taking another look to confirm all looks correct?

paulomarg (approved):

PR #501, opened 2023-11-24
Proposal oauth refactor

byrichardpowell:

1. Move some leaf functionality that will be shared between Auth code flow and token exchange into utilities.  Call those utilities from auth code flow.  This allows each stratagey to call those utilities in a way that makes sense for the strategy.  This mirrors the approach that is taken throughout the codebase, where if different authentication methods rely on the same logic, that logic is extracted into a utility.
2. Some minor changes to the auth code flow strategy.  Paulo and I thought these changes had a slight improvement on readability, but they are fairly inconsequential.
3. Change the authenticate base class to a factory function.  Since we weren't using inheritance to extend the shared authenticate functionality, we saw no good reason to keep it as a class.  Removing the class saves some ceremony (less passing of type generics, don't have to  deal with this) and it is more consistent with the rest of the codebase, which uses factory functions extensively.

## Type of change

## Checklist


packages/shopify-app-remix/src/server/authenticate/admin/strategies/auth-code-flow.ts
```diff
@@ -36,52 +40,70 @@ export class AuthCodeFlowStrategy<
...
-    if (isAuthRequest) throw await this.handleAuthBeginRequest(request, shop);
+    // If this is a valid request, but it doesn't have a session token header, this is a document request. We need to
+    // ensure we're embedded if needed and we have the information needed to load the session.
+    if (!getSessionTokenHeader(request)) {
+      const params = {api, logger, config};
 
-    if (isAuthCallbackRequest) {
-      throw await this.handleAuthCallbackRequest(request, shop);
+      await this.ensureInstalledOnShop(request);
+      await ensureAppIsEmbeddedIfRequired(params, request);
```
ragalie, line 67: I don't see why either this nor `ensureSessionTokenSearchParamIfRequired` should be in the token acquisition strategy. These are things that need to happen as part of request validation, regardless of how we get the token.

It's also imo confusing for these to live in the `respondToOauthRequests` function. These have nothing to do with OAuth.

These functions should live in with the rest of the request validation logic (which is currently housed in `authenticate.ts`.

PR #519, opened 2023-12-07
Add Github workflow to publish an experimental build from the `experimental` branch

rachel-carvalho:

We want to be able to have an experimental build published to NPM.

This creates a new Github workflow that builds the `experimental` branch,  creates an `experimental` snapshot version and publishes it to NPM under the `experimental` tag.

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/publish-experimental-build.yml
+++ b/.github/workflows/publish-experimental-build.yml
@@ -1,0 +1,36 @@
+name: Publish experimental release
+on:
+  workflow_dispatch:
+
+jobs:
+  experimental-release:
+    name: Publishing experimental release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@master
+        with:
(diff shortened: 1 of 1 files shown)


paulomarg (approved):
Makes sense to me! Just double checking that this will release _any_ changesets that are present in the experimental branch? I.e. we need to be careful with what we merge into that branch.

(that is fine, I just want to understand the setup)

rachel-carvalho:
@paulomarg that's right, any change that goes into that branch will be released in the `experimental` track when we run this workflow. 

So for instance if we make a change to `shopify-app-remix` plus another change to `shopify-app-session-storage-prisma`, it will create a release for each of these packages with a version `0.0.0-experimental-202312110927`, for instance.

This makes me realize that I left the name of the workflow referencing remix, because I was trying to get the release to be exclusive to the remix package, but that doesn't work with the current dependency graph, so I'm pushing a rename of the file.

PR #527, opened 2023-12-08
Remix: Redirect to admin install path if future flag is enabled

zzooeeyy:

Fixes: https://github.com/Shopify/develop-app-access/issues/132

- If app is embedded and the future flag is enabled for `unstable_newEmbeddedAuthStrategy`, redirect to admin install path instead of going through the app auth path

## Type of change

## Checklist

## Tophatting:

https://github.com/Shopify/shopify-app-js/assets/102243935/6c28c572-aa9d-47c4-ab53-7d293a741e1a


RyanDJLee (approved):
LGTM!

Left a comment about wanting to keep things on the Unified Admin, but it isn't a major blocker for the time being especially if we are keeping things on the experimental branch.

IMO we should try to test everything on the Unified Admin as soon as possible because all shops are technically supposed to be on it except for a very few exceptions. There is also a known, elusive [bug](https://github.com/Shopify/shopify/issues/458908) with relying on the backend middleware to keep app installation on the Unified Admin. Given how this project already has quite a few dependencies, might be worth eliminating this one.

paulomarg (approved):

packages/shopify-app-remix/src/server/authenticate/login/__tests__/login.test.ts
```diff
@@ -117,7 +118,7 @@ describe('login helper', () => {
 
         // THEN
         const expectedPath = testCaseConfig.redirectToInstall
-          ? `https://${TEST_SHOP}/admin/oauth/install?client_id=${config.apiKey}`
+          ? `https://admin.shopify.com/store/${TEST_SHOP_NAME}/oauth/install?client_id=${config.apiKey}`
```
zzooeeyy, line 121: I'll add another refactor PR to use the new utility from [this PR](https://github.com/Shopify/shopify-api-js/pull/1093) once that's ready

PR #532, opened 2023-12-11
Handle error cases for embedded apps using token exchange strategy

rezaansyed:

Fixes https://github.com/Shopify/temp-project-mover-Archetypically-20250319145332/issues/243 & https://github.com/Shopify/temp-project-mover-Archetypically-20250319145332/issues/244

- Handles invalid session token errors on request
  - Document request: redirect to bounce page
  - XHR: `401` with App Bridge retry request header 
- Handles invalid subject token error on token exchange
  - Document request: redirect to bounce page
  - XHR: `401` with App Bridge retry request header 
- Handles invalid access token on API request
  - Document request: Delete the access token and redirect to bounce page (which eventually triggers token exchange)
  - XHR: Delete access token and return `401` **without** App Bridge retry header (since part of the request may already have been processed).

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
+++ b/packages/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
@@ -82,10 +82,9 @@ export function authStrategyFactory<
       | EmbeddedAdminContext<ConfigArg, Resources>
       | NonEmbeddedAdminContext<ConfigArg, Resources> = {
       admin: createAdminApiContext<Resources>(
-        request,
         session,
         params,
-        authStrategy,
+        authStrategy.handleClientError(request),
       ),
       billing: {
         require: requireBillingFactory(params, request, session),
--- a/packages/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts
+++ b/packages/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts
@@ -9,13 +9,10 @@ import {AppConfig, AppConfigArg} from 'src/server/config-types';
 import {BasicParams, ApiConfigWithFutureFlags} from 'src/server/types';
 
 import {respondToInvalidSessionToken} from '../../helpers';
-import {triggerAfterAuthHook} from '../helpers';
+import {handleClientErrorFactory, triggerAfterAuthHook} from '../helpers';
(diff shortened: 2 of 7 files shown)


packages/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts
```diff
@@ -84,6 +88,20 @@ export class TokenExchangeStrategy<Config extends AppConfigArg>
...
 
+  public async handleInvalidAccessTokenError({
+    request,
+    session,
+  }: HandleInvalidAccessTokenOptions): Promise<void> {
+    const {config, api, logger} = this;
+
+    config.sessionStorage.deleteSession(session.id);
+
+    respondToInvalidSessionToken({
```
paulomarg, line 99: Should this be throwing?

packages/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
```diff
@@ -68,16 +68,18 @@ export function authStrategyFactory<
...
-      admin: createAdminApiContext<Resources>(request, session, {
-        api,
-        logger,
-        config,
-      }),
+      admin: createAdminApiContext<Resources>(
+        request,
+        session,
+        params,
+        authStrategy,
```
paulomarg, line 81: I like the solution of having the strategy determine how we handle errors, but I'm a little on the fence about passing in the entire auth strategy as a parameter for creating the context, it feels like we're mixing concerns a little bit.

What if, instead of passing the strategy in to `handleClientErrorFactory`, we have the strategy export its own flavour of `handleClientErrorFactory`, something like

```ts
  public async handleClientError({
    request,
    session,
  }: StrategyHandleClientErrorOptions): Promise<HandleAdminClientError> {
    const {config, api, logger} = this;
    return handleClientErrorFactory({
      onError: // Do token exchange things
    });
  }
```

That way, each strategy can determine how it handles the errors and they all subscribe to a common interface without the strategy having to be shared around.

IIRC `handleClientErrorFactory` is pretty much tied to the strategy already anyway, so it shouldn't really affect anything else. Perhaps it makes more sense for the strategy to just export a callback for its own error handling.

Does that sound sensible?

PR #561, opened 2024-01-08
Update header for retry invalid session

zzooeeyy:
Update retry invalid session header so that it won't be stripped


rezaansyed (approved):

rezaansyed:
> Will this change break pre-existing apps that upgrade to this version?

Nope, this is only part of token exchange so no one is using this header yet.

paulomarg:
True, I missed that :)

Thought it was the other `X-` header.

VishnuKP11:
https://hdshipkart.myshopify.com

Issue #560, opened 2024-01-07
[Feat Request] new Session adapter `@shopify/shopify-app-session-storage-drizzle` 


blanklob:
## Overview

Will be great to build `drizzleorm` session adapter since its much more compatible with Cloudflare D1 Db whiel Prisma isn't thanks. 

- DrizzleORM https://github.com/drizzle-team/drizzle-orm

lizkenyon:
Hi there 👋 

Thanks for reaching out with the request! The team will review this.

We may not get to this right away.

We do also accept community contributions for new session storage adaptors (Like this [dynmodb])(https://github.com/Shopify/shopify-app-js/tree/main/packages/shopify-app-session-storage-dynamodb) one! So if you want to put up a PR for this the team would be able to help with reviewing it.

doitalldev:
It doesn't tackle D1 though does it? I was getting a db.insert error when trying to use the package with D1

PR #587, opened 2024-01-18
feat(adapters): Add Drizzle

notrab:

[Drizzle](https://orm.drizzle.team/) is popular TypeScript ORM that would benefit Shopify App developers.

This PR adds the ability to use Drizzle as your storage adapter.

Some things aren't clear from a Drizzle side (it doesn't look to be possible or similar to Prisma). There doesn't appear to be a way that I can find in the Drizzle docs to get the schema from a db instance so that's why the constructor requires 2 args &mdash; `db` and `sessions`.

I tested this using a local SQLite file (via [libSQL](https://docs.turso.tech/libsql)) and Drizzle following the documentation in the `README.md`.

If there are any changes or things incorrect, please let me know.

## Type of change

- Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/shopify-app-session-storage-drizzle/README.md
+++ b/packages/shopify-app-session-storage-drizzle/README.md
@@ -2,7 +2,7 @@
 
 This package implements the `SessionStorage` interface that works with an instance of [Drizzle](https://orm.drizzle.team).
 
-There are 3 adapters for Drizzle: `DrizzleSessionStoragePostgres`, `DrizzleSessionStorageSQLite` and `DrizzleSessionStorageMySQL`. 
+There are 3 adapters for Drizzle: `DrizzleSessionStoragePostgres`, `DrizzleSessionStorageSQLite` and `DrizzleSessionStorageMySQL`.
 
 Session storage for Drizzle requires a `schema.ts` with a `session` table with at-least the columns as in the example. Make sure to create `session` table and apply changes to the database before using this pakacge.
 
--- a/packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-mysql.adapter.ts
+++ b/packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-mysql.adapter.ts
@@ -109,7 +109,7 @@ export class DrizzleSessionStorageMySQL implements SessionStorage {
   }
 
   private rowToSession(row: InferSelectModel<MySQLSessionTable>): Session {
-    const sessionParams: {[key: string]: boolean | string | number} = {
+    const sessionParams: Record<string, boolean | string | number> = {
       id: row.id,
       shop: row.shop,
       state: row.state,
--- a/packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-sqlite.adapter.ts
+++ b/packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-sqlite.adapter.ts
@@ -102,7 +102,7 @@ export class DrizzleSessionStorageSQLite implements SessionStorage {
   }
 
   private rowToSession(row: InferSelectModel<SQLiteSessionTable>): Session {
-    const sessionParams: {[key: string]: boolean | string | number} = {
+    const sessionParams: Record<string, boolean | string | number> = {
       id: row.id,
       shop: row.shop,
       state: row.state,
--- a/packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-postgres.adapter.ts
+++ b/packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-postgres.adapter.ts
@@ -104,7 +104,7 @@ export class DrizzleSessionStoragePostgres implements SessionStorage {
   }
 
   private rowToSession(row: InferSelectModel<PostgresSessionTable>): Session {
-    const sessionParams: {[key: string]: boolean | string | number} = {
+    const sessionParams: Record<string, boolean | string | number> = {
       id: row.id,
       shop: row.shop,
       state: row.state,
--- a/packages/shopify-app-session-storage-drizzle/tsconfig.json
+++ b/packages/shopify-app-session-storage-drizzle/tsconfig.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "./build/ts",
-    "rootDir": "src",
+    "rootDir": "src"
   },
   "include": ["src/**/*.ts"],
   "exclude": ["**/*.test.ts", "**/*.test.tsx", "**/test/*", "**/__tests__/*"]


notrab:
I have signed the CLA!

notrab:
Awesome, thanks @matteodepalo. I'm new to Drizzle but happy to change anything, just let me know.

blanklob:
This will close #560

realmikesolo:
Hello, @notrab and @matteodepalo . I am an active contributor to Drizzle ORM and have some improvements that I'd like to implement for this adapter. These include enhancing type-safety and optimizing SQL operations for better correctness and performance, which I believe will make them more suitable for Drizzle ORM. I suggest we postpone this pull request until I have made these improvements based on this code. What are your thoughts on this?

realmikesolo:
@notrab, thank you! I'm starting to work on this. For any questions or discussions, you can contact me on Discord at `@solo`.

realmikesolo:
I have signed the CLA!

packages/shopify-app-session-storage-drizzle/LICENSE.md
```diff
@@ -0,0 +1,9 @@
+MIT License
+
+Copyright (c) 2022-present, Shopify Inc.
```
paulomarg, line 3: Total nit
```suggestion
Copyright (c) 2024-present, Shopify Inc.
```

packages/shopify-app-session-storage-drizzle/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "graphql",
+    "rest",
+    "webhook",
+    "Admin API",
+    "Storefront API",
+    "session storage",
+    "Drizzle"
+  ],
+  "dependencies": {
+    "drizzle-orm": "^0.29.3",
```
paulomarg, line 34: If we expect folks to also include `drizzle-orm` directly in their apps, we should consider making it a peer dependency instead to prevent multiple instances being installed.

packages/shopify-app-session-storage-drizzle/README.md
```diff
@@ -0,0 +1,41 @@
...
+There are 3 adapters for Drizzle: `DrizzlePgAdapter`, `DrizzleSQLiteAdapter` and `DrizzleMySQLAdapter` which is in development. 
+
+Session storage for Drizzle requires a `schema.ts` with a `session` table. Make sure to create `session` table and apply changes to the database before using this pakacge.
+
+## Example with SQLite
+
+```ts
+import {sqliteTable, text, integer, blob} from 'drizzle-orm/sqlite-core';
+
+export const session = sqliteTable('session', {
```
paulomarg, line 14: Nit: could we call this `sessionTable` instead? We usually use `session` in other places in the app and this might get a little confusing.

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-sqlite.adapter.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export class DrizzleSQLiteAdapter<
+  T extends BaseSQLiteDatabase<'sync' | 'async', TRunResult, TSchema>,
+  TRunResult,
+  TSchema extends {[key: string]: unknown} = {[key: string]: unknown},
+> implements SessionStorage
+{
+  constructor(
+    private readonly db: T,
+    private readonly sessionTable: SQLiteSessionTable,
```
paulomarg, line 22: I assume this means the user can override the session table if they want to, right? That'd be a good thing to have.

packages/shopify-app-session-storage-drizzle/src/schemas/pg.schema.ts
```diff
@@ -0,0 +1,14 @@
...
+  shop: text('shop').notNull(),
+  state: text('state').notNull(),
+  isOnline: boolean('isOnline').default(false).notNull(),
+  scope: text('scope'),
+  expires: timestamp('expires', {precision: 3, mode: 'string'}),
+  accessToken: text('accessToken'),
+  userId: bigint('userId', {mode: 'bigint'}),
+});
+
+export type PgSessionTable = typeof session;
```
paulomarg, line 14: Question: if the app is setting their own table schema, will this type be compatible with their schema?

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-pg.adapter.ts
```diff
@@ -0,0 +1,139 @@
...
+import {
+  PgDatabase,
+  QueryResultHKT as PgDatabaseQueryResultHKT,
+} from 'drizzle-orm/pg-core';
+
+import {PgSessionTable} from '../schemas/pg.schema';
+
+export class DrizzlePgAdapter<
+  T extends PgDatabase<PgDatabaseQueryResultHKT, TSchema>,
+  TSchema extends {[key: string]: unknown} = {[key: string]: unknown},
```
paulomarg, line 19: Should this extend the schema type instead?
```suggestion
  TSchema extends PgSessionTable,
```

If yes, we should probably type `sessionTable` as `TSchema` in the constructor, right?

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-pg.adapter.ts
```diff
@@ -0,0 +1,139 @@
...
+      expires: session.expires
+        ? new Date(session.expires).toDateString()
+        : null,
+      accessToken: session.accessToken,
+      userId: session.onlineAccessInfo?.associated_user.id as unknown as bigint,
+    };
+  }
+
+  private rowToSession(row: InferSelectModel<PgSessionTable>): Session {
+    const sessionParams: {[key: string]: boolean | string | number} = {
```
paulomarg, line 114: This works, but it should also be possible to just create a `Session` with the required values, and then assign the rest to that object.

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-pg.adapter.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (row.scope) {
+      sessionParams.scope = row.scope;
+    }
+
+    if (row.accessToken) {
+      sessionParams.accessToken = row.accessToken;
+    }
+
+    if (row.userId) {
+      sessionParams.onlineAccessInfo = String(row.userId);
```
paulomarg, line 134: This would need to be a `OnlineAccessInfo` object, something like (untested):
```suggestion
      sessionParams.onlineAccessInfo = { associated_user: { id: row.userId } };
```

It seems some of the other adapters are also doing this, which is probably leading to some interesting scenarios. Something for us to fix later :)

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-pg.adapter.ts
realmikesolo: No, this is not possible with Drizzle as I know

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-pg.adapter.ts
realmikesolo: Honestly, I don't know what should I do here, because then we use `Session.fromPropertyArray(Object.entries(sessionParams));`
where entries have this type `entries: [string, string | number | boolean][]`. So it will be incompatible

paulomarg: Yeah, I think this is a wider problem and it would probably take a bigger change anyway. This is fine as is because it's not limited to this adapter.

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-pg.adapter.ts
realmikesolo: Could you please tell a bit more about this, because I didn't get it, probably some example? Firstly, I declare `sessionParams` with required values and then add optional values.

paulomarg: I was just thinking that instead of creating an object here, we could do something like:

```ts
const session = new Session({ <mandatory fields> });

if (row.<field>) {
  session.<field> = row.<field>;
}
```

That would enable us to ensure that the types for the session are being respected as we set the values.

paulomarg: Yeah I figured this might happen - I think you can leave this as is and we'll find a broader solution to all of the packages later on.

realmikesolo:
Hi, @paulomarg. I am trying to run tests, but have encountered an issue that jest cannot process ES module syntax, as it seems to be using default configuration parameters. Could you provide more details on how to set up the test workflow to allow Jest to properly handle ES module code?

paulomarg:
Hmm, it should be set up to work with that already 🤔, loom handles that config so I'd be surprised if it failed that way just for this package. Are you able to run the tests for other packages in this repo?

packages/shopify-app-session-storage-drizzle/README.md
```diff
@@ -0,0 +1,42 @@
...
+});
+```
+
+You can then instantiate and use `DrizzleSQLiteAdapter` like so:
+
+```js
+import {shopifyApp} from '@shopify/shopify-app-express';
+import {DrizzleSQLiteAdapter} from '@shopify/shopify-app-session-storage-drizzle';
+
+import {db} from './db.server';
```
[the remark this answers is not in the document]
notrab: I can tackle this tomorrow @paulomarg @realmikesolo and update the README

realmikesolo:
@paulomarg Apologies for the delayed response. I've added a test for PostgreSQL in the new pull request and renamed the classes as you suggested. Once we resolve the issue with Jest and ESM, I will proceed to add a test for SQLite as well. Additionally, I will make any necessary adjustments to the PostgreSQL test if required.

paulomarg:
Thanks! I'm going to be a little bit slow to respond this week, but I'll try to carve out some time to try and see if I can figure this error out :)

realmikesolo:
Hello, @paulomarg! Thank you very much for the help; everything worked out. I updated the loom config and made some small changes. Next, I am going to work on SQLite tests and the MySQL adapter.

packages/shopify-app-session-storage-drizzle/package.json
```diff
@@ -31,7 +31,9 @@
     "Drizzle"
   ],
   "dependencies": {
-    "tslib": "^2.6.2"
+    "pg": "^8.11.3",
```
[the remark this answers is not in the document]
notrab: Isn't `pg` just used for tests here? We could most it to `devDependencies` perhaps?

paulomarg:
Hey! I played around with this error a bit, and strangely I don't get it on `main` - but I was able to get it to work (not 100% reliably, unfortunately) by adding resolutions for the jest and jest-circus packages to the root dir's `package.json` file. I'd be OK with that as a solution but I haven't found the silver bullet yet :(

paulomarg:
Quick update: this seems to work for me, I got the tests to pass reliably by adding this to the root package.json:

```diff
@@ -26,6 +26,7 @@
     "eslint": "^8.55.0",
     "eslint-plugin-prettier": "^5.1.3",
     "jest": "^29.1.0",
+    "jest-environment-jsdom": "^29.7.0",
     "jest-fetch-mock": "^3.0.3",
     "jest-runner-eslint": "^2.1.2",
     "prettier": "^3.2.4",
@@ -34,6 +35,11 @@
     "tslib": "^2.6.2",
     "typescript": "5.3.3"
   },
+  "resolutions": {
+    "jest": "^29.1.0",
+    "jest-circus": "^29.1.0",
+    "ts-jest": "^29.1.0"
+  },
   "dependencies": {},
   "workspaces": [
     "packages/*"
```

paulomarg:
Would you mind doing it and running locally to see if that fixes it for you? That should tell us whether it'll work in CI :)

notrab:
Hey @paulomarg @realmikesolo 

I've merged upstream and all tests are passing.

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-mysql.adapter.ts
paulomarg, line 1: Also, something just occurred to me while checking this out: the code for all the adapters seems to be very similar.

I may have asked this before, but is there no way to have a generic type for the `db` param that we can share across all of the adapters, so we only need to implement them once? Probably ok if there's no easy way to do that, but it would make this adapter easier to maintain in the future.

paulomarg (commented):
I tested this out in an app and it worked like a charm! The tests are passing, but we had a few linting errors, could you please fix those so we can merge?

Also: 🎉

packages/shopify-app-session-storage-drizzle/src/adapters/drizzle-mysql.adapter.ts
paulomarg: Hmm, I think that might end up complicating things a bit more. We can re-evaluate to see if people are running into problems with the way things are now.

blanklob:
Keep us updated when this merged and ready for use.

PR #593, opened 2024-01-24
Fix dev dependencies across all projects

paulomarg:

Currently, there's a lot of repetition on our dependencies across the multiple packages in this repo, which makes it way harder to maintain them than it has to be. The vast majority of the problematic dependencies are development only so they shouldn't have any effect on production.

Removing any dependencies from the individual packages that were already added at the top level.

## Type of change

## Checklist


package.json
```diff
@@ -22,12 +22,16 @@
     "@shopify/loom-plugin-prettier": "^2.0.1",
     "@shopify/prettier-config": "^1.1.2",
     "@shopify/typescript-configs": "^5.1.0",
+    "@types/jest": "^29.5.1",
     "eslint": "^8.55.0",
+    "eslint-plugin-prettier": "^4.2.1",
     "jest": "^29.1.0",
```
lizkenyon, line 28: Looks like we were at `"@types/jest": "^29.5.1",` in shopify-app-remix. Though not that, that should cause us any problems

lizkenyon (approved):

PR #594, opened 2024-01-24
Update Typescript to v5

paulomarg:
This PR updates the dev dependency on `typescript` to the latest version, and fixes a couple of configurations that required updates.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -5,7 +5,8 @@
     "declaration": true,
     "declarationMap": true,
     "pretty": true,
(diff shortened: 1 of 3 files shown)


tsconfig.base.json
```diff
@@ -5,7 +5,8 @@
     "declaration": true,
     "declarationMap": true,
     "pretty": true,
-    "moduleResolution": "node16",
+    "module": "Node16",
```
paulomarg, line 8: TS complained about this not being set.

matteodepalo (approved):
💯

PR #634, opened 2024-02-06
[Feature] Add check function to billing API

lizkenyon:

Fixes [How to handle a free tier with the billing API?](https://github.com/Shopify/shopify-app-template-remix/issues/431)

* This PR exposes the `check` function from the API layer on the billing API
* Developers want an easy way to determine what plans a merchant has active without throwing an error.
* This will allow developers to implement a free tier for their app plan.

## Type of change

## Checklist


packages/shopify-app-remix/src/server/authenticate/admin/billing/__tests__/check.test.ts
```diff
@@ -0,0 +1,203 @@
...
+const BILLING_CONFIG: Shopify['config']['billing'] = {
+  [responses.PLAN_1]: {
+    amount: 5,
+    currencyCode: 'USD',
+    interval: BillingInterval.Every30Days,
+  },
+};
+
+describe('Billing check', () => {
+  it('returns true when there is payment', async () => {
```
paulomarg, line 35: Total nit, but we're not actually returning `true` here. Tiny but I went in expecting something else :)

packages/shopify-app-remix/src/server/authenticate/admin/billing/types.ts
```diff
@@ -155,6 +163,61 @@ export interface BillingContext<Config extends AppConfigArg> {
     options: RequireBillingOptions<Config>,
   ) => Promise<BillingCheckResponseObject>;
 
+  /**
+   * Checks if the shop has an active payment for any plan defined in the `billing` config option.
+   *
+   * @returns A promise that resolves to an object containing the active purchases for the shop.
+   *
+   * * @example
```
paulomarg, line 171: Nit: this will break indentation in the docs.

packages/shopify-app-remix/src/server/authenticate/admin/billing/types.ts
```diff
@@ -155,6 +163,61 @@ export interface BillingContext<Config extends AppConfigArg> {
...
+   * <description>Use billing.check if you want to determine which plans are in use. Unlike `require`, `check` does not
+   * throw an error if no active billing plans are present. </description>
+   * ```ts
+   * // /app/routes/**\/*.ts
+   * import { LoaderFunctionArgs } from "@remix-run/node";
+   * import { authenticate, MONTHLY_PLAN } from "../shopify.server";
+   *
+   * export const loader = async ({ request }: LoaderFunctionArgs) => {
+   *   const { billing } = await authenticate.admin(request);
+   *   const { resp } = await billing.check({
```
paulomarg, line 182: Do we return an object like `{resp: {hasActivePayment, ...}}`? That doesn't sound ideal - should we unpack that in this package and just return `{hasActivePayment, ...}`?

paulomarg (approved):
Just nits, but this is 🔥

packages/shopify-app-remix/src/server/authenticate/admin/billing/types.ts
lizkenyon: No it doesn't, and I realized this when writing test but forgot to update the docs. good catch

PR #635, opened 2024-02-07
Update `unauthenticated` session loading to throw specific `SessionNotFoundError`

paulspringett:

It's currently awkward to catch this specific error, requiring matching on the error message string.

#### Before

```typescript
import {unauthenticated} from '~/shopify.server';

try {
  const {admin} = await unauthenticated.admin(shop);
} catch (err) {
  if (err.message.includes('Could not find a session for shop')) {
    // do something specific
  }
}
```

#### After

```typescript
import {SessionNotFoundError} from '@shopify/shopify-app-remix';
import {unauthenticated} from '~/shopify.server';

try {
  const {admin} = await unauthenticated.admin(shop);
} catch (err) {
  if (err instanceof SessionNotFoundError) {
    // do something specific
  }
}
```

Defines and exports a new error class extending `ShopifyError`, meaning it should be backwards compatible for anyone already doing `err instanceof ShopifyError`.

Throws `SessionNotFoundError` instead of `ShopifyError` if the session is missing.

## Type of change

## Checklist

### DIFF
--- a/.changeset/stupid-peas-lay.md
+++ b/.changeset/stupid-peas-lay.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-app-remix": minor
+---
+
+Add SessionNotFoundError for unauthenticated requests
--- a/packages/shopify-app-remix/src/server/errors.ts
+++ b/packages/shopify-app-remix/src/server/errors.ts
@@ -1,0 +1,3 @@
+import {ShopifyError} from '@shopify/shopify-api';
+
+export class SessionNotFoundError extends ShopifyError {}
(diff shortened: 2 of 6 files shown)


CaptainDiego:
Examples in Subscriptions Remix App and SFN Remix App
- https://github.com/Shopify/subscriptions-app-remix/pull/923
- https://github.com/Shopify/sfn/blob/a72a38fa1754a5f5015d03f6cd03b6ee78cef374/app/utilities/error/shopifyError/types.ts#L12-L18

paulomarg (approved):
🔥 

Can't see any other places where we might need to call this, either!

PR #639, opened 2024-02-09
Return 403 in non-embedded XHR requests

### LINKED ISSUE #470: [`shopify-app-express`]: validateAuthenticatedSession for non-embedded apps that make XHR requests for data?
## Overview

👋 Thanks for supporting this robust SDK that does a lot of the heavy lifting for us!

We are building a straightforward application that leverages Shopify APIs on the backend.

After the initial installation and scope grant, we want to make subsequent calls to the API that are instantiated by a user interaction in the frontend. For this purpose, we want to use a short-lived (online) access token as described by the [docs](https://shopify.dev/docs/apps/auth/oauth/access-modes#online-access):

> Online access is meant to be used when a user is interacting with your app through the web, [...]
>
> Example use case(s): Your app's security requirements specify short-lived access to a store. Tokens with online access mode expire either when the user logs out or after 24 hours.

To illustrate the use-case:

1. User installs Shopify app, accepts grant, and is taken to our product landing page (self hosted; not embedded).
2.  On the product landing page, the user can subscribe to a _Billing Plan_ of their choosing.
3. The "Subscribe" action (triggered by the user) makes a call to our backend (e.g. `/api/pay?store={name}`).
4. The backend endpoint validates that the session is valid:
    4a. If valid, the route handlers makes the M2M call to Shopify's billing API, and respond accordingly
    4b. If not valid, the route handler results in a redirection to begin the authn/authz flow. Once authorized, they will then be redirected back to our product page, which should then automatically trigger the `/api/pay?store={name}` request again (this time with valid credentials).

For (4), on the surface, it would appear that this is specifically what the [validateAuthenticatedSession](https://github.com/Shopify/shopify-app-js/blob/bdfd30bea4b743d61e8b3589a48021b6c0411399/packages/shopify-app-express/src/middlewares/validate-authenticated-session.ts#L13) is designed to do. The [documentation](https://github.com/Shopify/shopify-app-js/blob/bdfd30bea4b743d61e8b3589a48021b6c0411399/packages/shopify-app-express/docs/reference/validateAuthenticatedSession.md) suggests it should behave correctly for both embedded and non-embedded apps.

In the case where the user is authorized, the middleware works great. However, in the case where the session is invalid, the [redirectOutOfApp](https://github.com/Shopify/shopify-app-js/blob/main/packages/shopify-app-express/src/redirect-out-of-app.ts#L9) function will result in a `302` to the auth endpoint. This tells the browser (who initialized the `fetch()` request) to redirect. The first redirect to the auth endpoint is fine, but then _that_ route handler redirects to `/admin/oauth/authorize` on the shopify store domain. Since this is a cross-origin redirect, the browser fails due to CORS:

```
Request URL:
http://localhost:8445/api/integration/shopify/pay?shop={shop-name}.myshopify.com&plan=Plus%20Annually
Request Method:
GET
Status Code:
302 Found
--> Location: /api/integration/shopify/auth-v2?shop={shop-name}.myshopify.com

Request URL:
http://localhost:8445/api/integration/shopify/auth-v2?shop={shop-name}.myshopify.com
Request Method:
GET
Status Code:
302 Found
--> Location: https://{shop-name}.myshopify.com/admin/oauth/authorize?client_id={client-id}&scope=write_themes%2Cread_products%2Cwrite_pixels%2Cread_customer_events&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fintegration%2Fshopify%2Fauth-v2%2Fcallback&state=147754441580872&grant_options%5B%5D=

Request URL:
https://{shop-name}.myshopify.com/admin/oauth/authorize?client_id={client-id}&scope=write_themes%2Cread_products%2Cwrite_pixels%2Cread_customer_events&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fintegration%2Fshopify%2Fauth-v2%2Fcallback&state=147754441580872&grant_options%5B%5D=
--> ERROR
```
> Access to fetch at 'https://{shop-name}.myshopify.com/admin/oauth/authorize?client_id={client-id}&scope=write_themes%2Cread_products%2Cwrite_pixels%2Cread_customer_events&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fintegration%2Fshopify%2Fauth-v2%2Fcallback&state=147754441580872&grant_options%5B%5D=' (redirected from 'http://localhost:8445/api/integration/shopify/pay?shop={shop-name}.myshopify.com&plan=Plus%20Annually') from origin 'http://localhost:8445' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

-------------------------

To work around this, I have a custom middleware that is essentially a copy/paste of [validateAuthenticatedSession](https://github.com/Shopify/shopify-app-js/blob/bdfd30bea4b743d61e8b3589a48021b6c0411399/packages/shopify-app-express/src/middlewares/validate-authenticated-session.ts#L13). The only difference is that it responds with an HTTP `401` or `403` (with the `redirectUri` in the response body) as opposed to a `302`. This gives the XHR call a chance to react and set the `window.location` to the `redirectUri`, because there’s no mechanism for `fetch()` to capture a 3xx redirect response and handle it manually (as it would violate [atomic HTTP redirect handling](https://fetch.spec.whatwg.org/#atomic-http-redirect-handling) in the case where the URI exposes sensitive data).

-------------------------

For the case of non-embedded apps that make XHR calls, it would be nice to have an out-of-the-box middleware, especially since the current middleware is 90% there.

Besides the need to be able to change the `window.location`, we would also need some special redirect handling (e.g. using `/my-home-page` vs `/my-home-page?just-authorized` as the `callbackPath`), so that once the auth completes, the product page would then know to trigger the request initiated by step (3) "Subscribe" above.

-------------------------

I would also have expected this to be a fairly straightforward / common case (i.e. making credentialed requests via XHR in a non-embedded app), so I am surprised there is not something more well established. It would seem the App Bridge capability is tied to embedded apps w/ session tokens. This leads me to believe I may be missing something simple.

Thoughts?

paulomarg:

Closes #470 

As pointed out in the issue, we're returning a 302 that runs into CORS errors, making the `validateAuthenticatedSession` middleware essentially broken for non-embedded apps.

Returning the same 403 response headers we return for XHR requests for embedded apps. It'll be up to the app to catch and handle those requests, but it'll be a valid response, instead of a 302 that will always fail.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-express/src/redirect-out-of-app.ts
+++ b/packages/shopify-app-express/src/redirect-out-of-app.ts
@@ -8,9 +8,10 @@ export function redirectOutOfApp({
   config,
 }: ApiAndConfigParams): RedirectOutOfAppFunction {
   return function redirectOutOfApp({req, res, redirectUri, shop}): void {
-    if (!api.config.isEmbeddedApp && isFetchRequest(req)) {
-      appBridgeHeaderRedirect(config, res, redirectUri);
-    } else if (req.headers.authorization?.match(/Bearer (.*)/)) {
+    if (
+      (!api.config.isEmbeddedApp && isFetchRequest(req)) ||
+      req.headers.authorization?.match(/Bearer (.*)/)
+    ) {
       appBridgeHeaderRedirect(config, res, redirectUri);
     } else if (req.query.embedded === '1') {
       exitIframeRedirect(config, req, res, redirectUri, shop);
--- a/packages/shopify-app-express/src/redirect-out-of-app.ts
+++ b/packages/shopify-app-express/src/redirect-out-of-app.ts
@@ -1,13 +1,16 @@
 import {Request, Response} from 'express';
 
 import {AppConfigInterface} from './config-types';
-import {RedirectOutOfAppFunction, RedirectOutOfAppParams} from './types';
+import {RedirectOutOfAppFunction, ApiAndConfigParams} from './types';
 
 export function redirectOutOfApp({
+  api,
   config,
-}: RedirectOutOfAppParams): RedirectOutOfAppFunction {
+}: ApiAndConfigParams): RedirectOutOfAppFunction {
   return function redirectOutOfApp({req, res, redirectUri, shop}): void {
-    if (req.headers.authorization?.match(/Bearer (.*)/)) {
+    if (!api.config.isEmbeddedApp && isFetchRequest(req)) {
+      appBridgeHeaderRedirect(config, res, redirectUri);
+    } else if (req.headers.authorization?.match(/Bearer (.*)/)) {
       appBridgeHeaderRedirect(config, res, redirectUri);
     } else if (req.query.embedded === '1') {
       exitIframeRedirect(config, req, res, redirectUri, shop);
@@ -69,4 +72,8 @@ function serverSideRedirect(
   );
 
   res.redirect(redirectUri);
+}
+
+function isFetchRequest(req: Request) {
+  return req.xhr || req.headers['sec-fetch-dest'] === 'empty';
 }
--- a/.changeset/curvy-lizards-exist.md
+++ b/.changeset/curvy-lizards-exist.md
@@ -1,0 +1,10 @@
+---
+"@shopify/shopify-app-express": patch
+---
+
+Return a 403 with X-Shopify headers on XHR requests for non-embedded apps, instead of a 302. The 302 ran into CORS errors and always failed.
+
+These requests will return the following headers:
+
+- `X-Shopify-Api-Request-Failure-Reauthorize`: `1`
+- `X-Shopify-Api-Request-Failure-Reauthorize-Url`: <URL>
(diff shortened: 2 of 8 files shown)


packages/shopify-app-express/src/redirect-out-of-app.ts
```diff
@@ -1,13 +1,16 @@
...
 
 export function redirectOutOfApp({
+  api,
   config,
-}: RedirectOutOfAppParams): RedirectOutOfAppFunction {
+}: ApiAndConfigParams): RedirectOutOfAppFunction {
   return function redirectOutOfApp({req, res, redirectUri, shop}): void {
-    if (req.headers.authorization?.match(/Bearer (.*)/)) {
+    if (!api.config.isEmbeddedApp && isFetchRequest(req)) {
+      appBridgeHeaderRedirect(config, res, redirectUri);
```
matteodepalo, line 12: Since this is doing the same thing as the branch below, could we merge them with an `||`? Or would you prefer to keep them separate for clarity?

matteodepalo (approved):
Looking good! Just a minor comment.

PR #654, opened 2024-02-19
Add `cause` when throwing `MissingSessionTableError`

paulspringett:

Our app had an issue over the weekend where it could no longer see the `Session` table – the database itself was unaffected, so this seems to be a connection issue between the app and database.

We saw a lot of `Prisma session table does not exist` errors during the incident.

We believe the issue was caused by the container reaching its memory usage limit, however, it would be really useful to see the upstream cause of the `MissingSessionTableError`.

1. Adds the caught error as the `cause` for the thrown `MissingSessionTableError`
2. Adds `es2020` to the TypeScript `lib` options

## Type of change

## Checklist

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -31,6 +31,7 @@
       "es2018",
       "es2019",
       "es2020",
+      "es2022",
       "esnext.asynciterable"
     ],
     "baseUrl": ".",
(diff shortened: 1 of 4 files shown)


tsconfig.base.json
```diff
@@ -31,6 +31,7 @@
       "es2018",
       "es2019",
       "es2020",
+      "es2022",
```
paulspringett, line 34: It looks like the lowest version of Node we support is 16, and 16.13+ does support `Error.cause` so I think this is ok? 

See: https://node.green/#ES2022-features-Error-cause-property-Error-has-cause

paulomarg (approved):
Nice, thank you.

PR #682, opened 2024-03-03
update drizzle README file with postgresql & mysql examples, add links for dr…

realmikesolo:
Update readme file


paulomarg (commented):
This is great, thank you! Since we're only updating documentation here, we don't need to trigger a new release for it, so you can run `yarn changeset --empty` to create a changeset that doesn't alter the version!

realmikesolo:
Hello, @paulomarg! Thank you for your help. I updated readme file again, deleted the previous changeset and created the empty one.

paulomarg (approved):
Thanks!

PR #686, opened 2024-03-06
Catch Prisma P2002 error

zzooeeyy:
Fixes - https://github.com/Shopify/shopify-app-template-remix/issues/536

Catch and handle Prisma P2002 error but throw other errors. 

This is the[ suggested method from Prisma](https://www.prisma.io/docs/orm/reference/prisma-client-reference#unique-key-constraint-errors-on-upserts), however, in this case, we won't re-try since the data _should_ be identical. 

## Type of change

## Checklist

## Tophatting:
1. Manually modified the local Remix package to call `storeSesssion` twice concurrently during token exchange
06-18-xvmrx-vfr8r

2. See that the operation fails before the fix
error2

3. See that the operation succeeds, and the error is caught after the fix from a local Prisma package
caught

### DIFF
--- a/packages/shopify-app-session-storage-prisma/src/__tests__/prisma.test.ts
+++ b/packages/shopify-app-session-storage-prisma/src/__tests__/prisma.test.ts
@@ -84,6 +84,7 @@ describe('Prisma throws P2002 on upsert', () => {
     jest.spyOn(prisma.session, 'upsert').mockImplementation(() => {
       throw new Prisma.PrismaClientKnownRequestError('error message', {
         code: 'P2002',
+        clientVersion: '0',
       });
     });
   });
@@ -100,6 +101,29 @@ describe('Prisma throws P2002 on upsert', () => {
   it('Returns true after handling Prisma P2002 errors', async () => {
     const result = await storage.storeSession(session);
     expect(result).toBe(true);
+  });
+
+  it('Throws other errors not P2002', async () => {
+    const expectedError = new Prisma.PrismaClientKnownRequestError(
+      'error message',
+      {
+        code: 'P2003',
+        clientVersion: '0',
+      },
+    );
+    jest.clearAllMocks();
+    jest.spyOn(prisma.session, 'upsert').mockImplementation(() => {
+      throw expectedError;
+    });
+
+    try {
+      await storage.storeSession(session);
+
+      // This should never be reached
+      expect(true).toBe(false);
+    } catch (actualError) {
+      expect(actualError).toStrictEqual(expectedError);
+    }
   });
 });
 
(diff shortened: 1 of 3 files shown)


packages/shopify-app-session-storage-prisma/src/__tests__/prisma.test.ts
```diff
@@ -58,6 +59,50 @@ describe('PrismaSessionStorage when with no database set up', () => {
...
+  afterAll(async () => {
+    await prisma.session.deleteMany();
+    jest.clearAllMocks();
+  });
+
+  it('handles Prisma P2002 errors', async () => {
+    await expect(storage.storeSession(session)).resolves.not.toThrow();
+  });
+
+  it('Returns true after handling Prisma P2002 errors', async () => {
```
paulomarg, line 100: Could we also add a test that we leave errors other than this one alone?

packages/shopify-app-session-storage-prisma/src/prisma.ts
```diff
@@ -38,11 +39,24 @@ export class PrismaSessionStorage<T extends PrismaClient>
...
+      });
+    } catch (error) {
+      if (
+        error instanceof Prisma.PrismaClientKnownRequestError &&
+        error.code === 'P2002'
+      ) {
+        console.log(
+          'Caught PrismaClientKnownRequestError P2002, Session already exists',
+        );
+        return true;
```
paulomarg, line 56: Should we trigger an update here just in case? This is a race condition so quite unlikely that it'll be 2 different concurrent operations, but maybe it can happen?

packages/shopify-app-session-storage-prisma/src/prisma.ts
```diff
@@ -38,11 +39,24 @@ export class PrismaSessionStorage<T extends PrismaClient>
 
     const data = this.sessionToRow(session);
 
-    await this.getSessionTable().upsert({
-      where: {id: session.id},
-      update: data,
-      create: data,
-    });
+    try {
```
paulomarg, line 42: I wonder if instead of catching the error, it might make more sense to run upsert / delete operations inside [transactions](https://www.prisma.io/docs/orm/prisma-client/queries/transactions#the-transaction-api)?

paulomarg: Well, assuming the DBMS supports locking writes to the table, transactions should do the trick. However you're right in that we should probably not assume that will be the case for every system supported by prisma, right?

packages/shopify-app-session-storage-prisma/src/prisma.ts
```diff
@@ -38,11 +41,29 @@ export class PrismaSessionStorage<T extends PrismaClient>
...
+      if (
+        error instanceof Prisma.PrismaClientKnownRequestError &&
+        error.code === UNIQUE_KEY_CONSTRAINT_ERROR_CODE
+      ) {
+        await this.getSessionTable().upsert({
+          where: {id: session.id},
+          update: data,
+          create: data,
+        });
+        console.log(
```
paulomarg, line 60: Totally nit-picking, but we should probably log before the retry, just in case it blows up somehow 😄

paulomarg (approved):

PR #707, opened 2024-03-20
Begin process to replace Loom with Turbo

paulomarg:

We want to revamp our build process for the packages in this repo for a few reasons:
- Building both ESM and CJS outputs to make it easier for more apps to import the package
- Building with a target fitting for Node 16 (which we still technically support) to make stack traces clearer
- Building with turbo + rollup instead of loom (which uses rollup internally) to make it easier for folks to contribute

These changes should have no effect on the final packages, other than the ones stated above.


packages/shopify-app-session-storage-test-utils/rollup.config.js
```diff
@@ -0,0 +1,66 @@
...
+      noEmit: true,
+      emitDeclarationOnly: true,
+      outDir: './dist/ts',
+    }),
+  ];
+}
+
+const packageName = pkg.name.substring(1);
+export const bannerConfig = {
+  banner: `/*! ${packageName}@${pkg.version} -- Copyright (c) 2023-present, Shopify Inc. -- license (MIT): https://github.com/Shopify/shopify-app-js/blob/main/LICENSE.md */`,
```
lizkenyon, line 29:
```suggestion
  banner: `/*! ${packageName}@${pkg.version} -- Copyright (c) 2024-present, Shopify Inc. -- license (MIT): https://github.com/Shopify/shopify-app-js/blob/main/LICENSE.md */`,
```
?

paulomarg:
> Will any of these changes resolve the down compilation of the async await, like mentioned [here](https://github.com/Shopify/shopify-api-js/issues/657)?

Yes, we'll start building with newer targets and move away from `es5`, which should hopefully create much more readable transpiled code to use more modern features instead.

paulomarg:
> and why turbo? 

We've been using turbo successfully on `shopify-api-js` for a while now, it gives us a good mix of flexibility and ease of use for now. Since this is already going to be a massive change where we're touching every package, I wanted to err on the safe side and not introduce a whole new tool.

We can always migrate off of turbo later once everything is in place, but for now I'd rather focus on getting better output transpiled code without messing with the toolset.

PR #722, opened 2024-03-27
Build up to the Prisma adapter with turbo

paulomarg:

The changes in #707 allowed us to start building packages with turbo in this repo. Following that, we'll need to get all of the packages building with the new process.

As a first step, this PR is changing all of the packages needed to build the Prisma session adapter to use turbo, and making the configurations for rollup a little more generic since most of it will be shared anyway.

Following this, I'll set up a PR to build the Remix package with the new setup, which will be a little more work since that package has some special configs. Following that, we'll build everything the new way and we'll be ready to merge the repos.


config/rollup/rollup.config.js
paulomarg, line 1: This is _mostly_ copying the file from its previous location in the session-storage package since this will apply to basically everything.

sle-c (approved):

PR #727, opened 2024-03-29
Add optional `poolConfig` to Postgres session options

thosakwe:
When using PostgreSQL for sessions, it's not possible to use a host that requires a specific SSL certificate.

For example, DigitalOcean managed databases come with a CA certificate needed for authentication.

The reason you can't use these hosts is because the PostgreSQLSessionStorage constructor has no way to provide SSL options.

I added an optional `pg.PoolConfig` option.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-session-storage-postgresql/src/postgres-connection.ts
+++ b/packages/shopify-app-session-storage-postgresql/src/postgres-connection.ts
@@ -5,12 +5,21 @@ export class PostgresConnection implements RdbmsConnection {
   sessionStorageIdentifier: string;
   private ready: Promise<void>;
   private pool: pg.Pool;
+  /**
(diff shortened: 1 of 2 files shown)


thosakwe:
I'm actually not sure this is necessary.

It turns out that `pg` accepts the same environment variables as libpq and psql:

> node-postgres uses the same [environment variables](https://www.postgresql.org/docs/9.1/static/libpq-envars.html) as libpq and psql to connect to a PostgreSQL server. Both individual clients & pools will use these environment variables.

https://node-postgres.com/features/connecting

So in my case, I could export `PGSSLMODE=require` and `PGSSLROOTCERT=<.ca file path>`

If this PR turns out to not be necessary, that's okay, this post will at least be on the Internet to help the next person who has this problem.

PR #729, opened 2024-04-02
Update shopify-api dependency

lizkenyon:

Bumps @shopify/shopify-api to 9.6.2

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-app-session-storage-test-utils/src/battery-of-tests.ts
+++ b/packages/shopify-app-session-storage-test-utils/src/battery-of-tests.ts
@@ -63,7 +63,7 @@ export function batteryOfTests(storageFactory: () => Promise<SessionStorage>) {
           id: sessionId,
           shop: 'shop',
           state: 'state',
-          isOnline: false,
+          isOnline: true,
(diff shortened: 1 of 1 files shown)


packages/shopify-app-session-storage-test-utils/src/battery-of-tests.ts
```diff
@@ -137,25 +138,25 @@ export function batteryOfTests(storageFactory: () => Promise<SessionStorage>) {
         id: `${prefix}_1`,
         shop: 'find-shop1-sessions.myshopify.io',
         state: 'state',
-        isOnline: true,
```
lizkenyon, line 140: Many of our tests were set up with invalid data. The current flow of code cannot create session data that looks like this. The new update to the session class in the shopify-api-js can no longer handle this invalid data setup.

paulomarg: Agreed, these just seem like invalid test cases.

paulomarg (approved):

PR #735, opened 2024-04-03
Build all packages with turbo

paulomarg:
This PR follows up #723, and updates all packages to build with the new tooling, thus making the repo ready for merging.


packages/shopify-app-express/jest.config.ts
```diff
@@ -0,0 +1,10 @@
+import type {Config} from 'jest';
+
+import baseConfig from '../../config/tests/jest.config';
+
+const config: Config = {
+  ...baseConfig,
+  testTimeout: 35000,
```
lizkenyon, line 7: Any reason for increasing the testTimeout for these ones?

lizkenyon (approved):

packages/shopify-app-express/jest.config.ts
paulomarg: I remember I had to do that, let me see if we can not.

PR #738, opened 2024-04-05
Return a bounce response on invalid tokens

paulomarg:

Closes https://github.com/Shopify/shopify-app-template-remix/issues/528
Closes https://github.com/Shopify/shopify-app-template-remix/issues/606

There are some scenarios in which Vite's HMR triggers a full page reload in the app (I believe one example is when you edit a page that is _not_ the current one).

If that happened more than 1 minute after the page loaded, the `id_token` for the request would have expired. Since Vite would replay the same request, it would then fail, in which case we simply threw a 401 response back, which causes the page load to blow up.

Changing that response so it returns a bounce request to get app bridge to request a new token before the reload instead of returning a 401. That way, the full page reload will use a new token and go through.

## Type of change

## Checklist

### DIFF
--- a/.changeset/famous-ducks-switch.md
+++ b/.changeset/famous-ducks-switch.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-app-remix": patch
+---
+
+Fixed an issue where full page reloads broke HMR on pages loaded for more than a minute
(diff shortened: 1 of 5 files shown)


.turbo/daemon/b00cdf3e7de80319-turbo.log.2024-03-27
sle-c, line 1: accidental commit?

sle-c (approved):
LGTM

PR #742, opened 2024-04-09
Rearranging packages folder

paulomarg:
After merging the repos, the structure of the `packages` folder was a bit too complicated. There were too many packages in a single folder, which made it harder to discover what was in there.

This PR moves files around, but doesn't necessarily fix any issues caused by the move - these are almost all just moved files.


paulomarg:
> But I guess that you would only use that, in an app context.

That was the thinking - the package is geared towards writing apps, so it's more of an app package than a client.

PR #782, opened 2024-04-11
Improve build process parallelization

paulomarg:

After joining the repos, CI was taking longer than ideal, because we were limiting how much parallelization we allowed. 

Reconfigure the turbo commands so we can make the most of parallelization, by ensuring tests can run in parallel without messing with building - this creates an implicit dependency on building before being able to run tests (which we do in CI).


matteodepalo (approved):
👍

PR #800, opened 2024-04-16
Unify rollup configs

paulomarg:

As part of joining the repos, we also started using rollup across the board. However, to make the changes smaller, I chose not to change the existing build pipelines for the previous shopify-api-js repo packages, which meant we had a bit of duplication.

Using the same `rollup-utils` functions to build all of the packages.


lizkenyon (approved):
I tested with the session-storage-prisma built it and linked it do a development app. Verified works as expected.

lizkenyon:
@paulomarg tested with shopify-api!

PR #807, opened 2024-04-19
Require ids in `find` methods

### LINKED ISSUE #756: Rest resource .find returns last seen object when passing id = NaN
# Issue summary

- [X] Upgraded to the latest version of the package
  - `@shopify/shopify-api` version: 9.5.1
  - Node version: 20.11.0
  - Operating system: Mac OS 14.0
- [X] Set `{ logger: { level: LogSeverity.Debug } }` when calling [`shopifyApi()`](https://github.com/Shopify/shopify-api-js/blob/main/packages/shopify-api/docs/reference/shopifyApi.md#logger), if applicable
- [X] Found a reliable way to reproduce the problem that indicates it's a problem with the package

## Expected behavior

Passing an invalid ID should throw, not silently return an object.

## Actual behavior

If you pass id: NaN to a .find method, the sdk seems returns the last object of that type that was dealt with.

```
    const product = new shopify.rest.Product({ session });

    product.title = "Burton Custom Freestyle 151";
    product.body_html = "<strong>Good snowboard!</strong>";
    product.vendor = "Burton";
    product.product_type = "Snowboard";
    product.status = "draft";

    await product.save({
      update: true,
    });

    const productThatShouldntBeReturned = await shopify.rest.Product.find({ session, id: NaN });
```

With the above code, I would be expecting that .find throws or at least returns null, but it's returning a fully formed object that happens to be the product created right before.

If you remove the product creation and only keep the .find part then the SDK seems to return the first product ordered by ID? It should throw as well.

paulomarg:

Closes #756 

When all ids evaluated to falsey values in the REST resources' `find()` methods, we could end up defaulting to the non-id specific versions of a request - e.g. `/products/<id>.json` became `/products.json` because we treated these ids as optional.

However, typically in REST APIs, whenever we are GET-ing a specific item you'll need to supply its id, so it didn't make sense for `find()` to do an id-less query.

Making it possible to trigger an error if we're missing an id by passing in a parameter to `baseFind`, and updating all `find()` methods to pass that parameter in.

I'm considering this a bug fix because those requests were never supposed to be fired in the first place and would cause unpredictable behaviour.

## Type of change

## Checklist

### DIFF
--- a/.changeset/short-pillows-own.md
+++ b/.changeset/short-pillows-own.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-api": patch
+---
+
+Fixing REST resource `find()` methods to fail when missing all ids, instead of defaulting to the same behaviour as `all()`.
--- a/packages/apps/shopify-api/lib/clients/admin/__tests__/resources/base.test.ts
+++ b/packages/apps/shopify-api/lib/clients/admin/__tests__/resources/base.test.ts
@@ -4,7 +4,7 @@ import {
 } from '../../../../__tests__/test-helper';
 import {testConfig} from '../../../../__tests__/test-config';
 import {Session} from '../../../../session/session';
-import {HttpResponseError} from '../../../../error';
+import {HttpResponseError, RestResourceError} from '../../../../error';
 import {PageInfo} from '../../../types';
 import {ApiVersion, LATEST_API_VERSION} from '../../../../types';
 import {shopifyApi} from '../../../../index';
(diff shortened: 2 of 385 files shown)


paulomarg:
Note: I search-replaced the `requireIds` param in the actual classes, so they should all be the same. Just skimming that code should be enough.

sle-c (approved):
LGTM, confirmed that `.all` methods have requireIds: false 🙇

PR #813, opened 2024-04-22
Alias v10_lineItemBilling to v11_lineItemBilling

lizkenyon:

We want to keep the `lineItemBilling` flag until the v11 release

* This PR allows you to use the flag either `v10_lineItemBilling` or `v11_lineItemBilling` to use a billing config with lineItems

## Type of change

## Checklist


packages/apps/shopify-api/future/flags.ts
```diff
@@ -36,10 +42,15 @@ export function logDisabledFutureFlags(
...
-  if (!config.future?.v10_lineItemBilling) {
+  if (!config.future?.v11_lineItemBilling) {
     logFlag(
-      'v10_lineItemBilling',
+      'v11_lineItemBilling',
       'Enable this flag to use the new billing API, that supports multiple line items per plan.',
     );
   }
+  if (config.future?.v10_lineItemBilling) {
+    logger.info(
```
lizkenyon, line 52: Should we use info or debug here?

packages/apps/shopify-api/lib/billing/types.ts
```diff
@@ -121,7 +121,9 @@ export type BillingConfigItem<
 > =
   FeatureEnabled<Future, 'v10_lineItemBilling'> extends true
```
lizkenyon, line 122: I attempted to create a type that would allow you only to set `v10_lineItemBilling` or `v11_lineItemBilling` but things got quite complicated. So taking suggestions on if/how we want to do that if we deem necessary.

packages/apps/shopify-api/lib/clients/admin/__tests__/rest_client.test.ts
```diff
@@ -594,7 +594,7 @@ describe('REST client', () => {
     await client.get({path: '/url/path'});
 
     // first call to .log is .debug with package and runtime info during initialization
-    expect(shopify.config.logger.log).toHaveBeenCalledTimes(2);
+    expect(shopify.config.logger.log).toHaveBeenCalledTimes(3);
```
lizkenyon, line 597: All these tests had to be updated because we added another log. We have the options of either updating these log counts or we could turn off the `v10_lineItemBilling` for these tests. (These tests don't test billing)

paulomarg: It'd be great if we didn't depend on the count here at all, but turning the flag off sounds reasonable too, because then we won't have to change them back later.

packages/apps/shopify-api/future/flags.ts
paulomarg: Maybe `deprecated` set to v11? Also worth pointing out that if we're not calling `logFlag` here we should include the flag name in the message :)

paulomarg (approved):
Just nits, LGTM

PR #814, opened 2024-04-23
Fix back stack for embedded apps

henrytao-me:

Resolves part of https://github.com/Shopify/shopify-app-bridge/issues/313

This PR fixes Link Component to work with back stack in embedded apps. 

## Type of change

## Checklist

### DIFF
--- a/.changeset/curvy-apes-beam.md
+++ b/.changeset/curvy-apes-beam.md
@@ -2,4 +2,4 @@
 "@shopify/shopify-app-remix": patch
 ---
 
-Fix back stack for embedded apps
+Fix a bug that prevents using the browser's back stack for embedded apps
--- a/.changeset/curvy-apes-beam.md
+++ b/.changeset/curvy-apes-beam.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-app-remix": patch
+---
+
+Fix back stack for embedded apps
--- a/packages/apps/shopify-app-remix/src/react/components/RemixPolarisLink.tsx
+++ b/packages/apps/shopify-app-remix/src/react/components/RemixPolarisLink.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import {Link} from '@remix-run/react';
+import React, {useEffect, useRef, useState} from 'react';
+import {Link, useNavigate} from '@remix-run/react';
 // Polaris' LinkProps doesn't match this type, so we need to import it directly
 import type {
   LinkLikeComponent,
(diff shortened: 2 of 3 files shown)


.changeset/curvy-apes-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopify/shopify-app-remix": patch
+---
+
+Fix back stack for embedded apps
```
paulomarg, line 5: Mind if we just add a little more colour here? This is the main way for folks to catch up on what we're doing, so I like to be very explicit in the changelog.
```suggestion
Fix a bug that prevents using the browser's back stack for embedded apps
```

packages/apps/shopify-app-remix/src/react/components/RemixPolarisLink.tsx
paulomarg, line 1: Could / should we have some tests for these changes? We could create a folder around this or add them to the existing AppProvider tests (which is the only place that will use this).

packages/apps/shopify-app-remix/src/react/components/RemixPolarisLink.tsx
```diff
@@ -9,6 +9,83 @@ import type {
...
+          } else {
+            forwardedRef.current = anchor;
+          }
+        }
+      }}
+    />
+  );
+}) as LinkLikeComponent;
+
+function parseUrl(url: string | undefined | null) {
```
paulomarg, line 62: Total nit, but does `undefined` vs `null` matter here? It seems like we could just use `undefined` all around?

paulomarg (commented):
Just some nits, but holding off on approving until we figure out whether tests make sense for this :)

packages/apps/shopify-app-remix/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": "./dist/ts",
-    "rootDir": "src"
+    "rootDir": "src",
+    "types": ["@cloudflare/workers-types", "node", "jest", "jsdom"]
```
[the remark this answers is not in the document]
henrytao-me: Production build uses `tsconfig.build.json`, doesn't it?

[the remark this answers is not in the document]
paulomarg: But it extends that one. We could also override the build config to set all types but jsdom, that would work :)

MitchLillie (approved):
🎩 via https://github.com/Shopify/app-bridge-next/pull/345

Agree on adding unit tests :)

henrytao-me:
Close as we are working on a solution that doesn't involve update in the template

PR #818, opened 2024-04-24
Add ability to encrypt session data

paulomarg:

Currently, when storing sessions in a database, we don't provide any utility from `@shopify/shopify-api` to encrypt the access token, which is easy to rotate for apps but might be annoying for developers if their databases are ever leaked.

To add an extra layer of security to the way apps store tokens, this PR introduces the ability to serialize and deserialize sessions in a way that handles the encryption for apps, given a `CryptoKey`.

If a row was previously unencrypted, these functions will still be able to load it and update the row when saving it, which means that the migration path for session storages is simply:
- `loadSession`
- `storeSession`

for every row in their database, if they don't want to wipe it clean.

I'll add documentation and a changelog entry to this PR once we're aligned on what it will do.

## Type of change

## Checklist

### DIFF
--- a/packages/apps/shopify-api/docs/guides/session-storage.md
+++ b/packages/apps/shopify-api/docs/guides/session-storage.md
@@ -1,22 +1,16 @@
 # Storing sessions
 
-As of v6 of the library, there are no `SessionStorage` implementations included and the responsibility for implementing session storage is now delegated to the application.
+In order to be able to load user data without using cookies, your app will need to store its access tokens and other relevant data.
 
-The previous implementations of `SessionStorage` are now available in their own packages, the source of which is available in the [respective directory](../../../session-storage#readme).
+This package provides a `SessionStorage` interface that makes it easy to plug in a new storage strategy to your app.
+You can use of one the [packages we provide](../../../session-storage/README.md), or implement your own following the instructions in this page.
 
-|                      Package                      |  Session storage object  | Notes                                    |
-| :-----------------------------------------------: | :----------------------: | ---------------------------------------- |
-|   `@shopify/shopify-app-session-storage-memory`   |   MemorySessionStorage   |                                          |
-|  `@shopify/shopify-app-session-storage-mongodb`   |  MongoDBSessionStorage   |                                          |
-|   `@shopify/shopify-app-session-storage-mysql`    |   MySQLSessionStorage    |                                          |
-| `@shopify/shopify-app-session-storage-postgresql` | PostgreSQLSessionStorage |                                          |
-|   `@shopify/shopify-app-session-storage-redis`    |   RedisSessionStorage    |                                          |
-|   `@shopify/shopify-app-session-storage-sqlite`   |   SQLiteSessionStorage   |                                          |
-|  `@shopify/shopify-app-session-storage-dynamodb`  |  DynamoDBSessionStorage  |                                          |
-|     `@shopify/shopify-app-session-storage-kv`     |     KVSessionStorage     |                                          |
-|      `@shopify/shopify-app-session-storage`       |      SessionStorage      | Abstract class used by the classes above |
+In this page, you'll file:
 
-## Basics
+- [What data is in a `Session` object?](#what-data-is-in-a-session-object)
+- [Save a session to storage](#save-a-session-to-storage)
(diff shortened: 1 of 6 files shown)


packages/apps/shopify-api/lib/session/session.ts
```diff
@@ -134,6 +146,48 @@ export class Session {
...
+
+    return `${Session.CIPHER_PREFIX}${asBase64(iv)}${cipher}`;
+  }
+
+  private static async decryptValue(value: string, key: CryptoKey) {
+    if (!value.startsWith(Session.CIPHER_PREFIX)) {
+      return value;
+    }
+
+    const keyString = value.slice(Session.CIPHER_PREFIX.length);
```
lizkenyon, line 184: Nit: Would it make sense to put getting the IV and cipher into the encrypt helpers files? Like how we have generateIV()

packages/apps/shopify-api/lib/session/session.ts
```diff
@@ -259,19 +313,42 @@ export class Session {
   /**
    * Converts the session into an array of key-value pairs.
    */
-  public toPropertyArray(
+  public toPropertyArray(returnUserData = false): SessionParamsArray {
+    return this.flattenProperties(this.toObject(), returnUserData);
+  }
+
+  /**
+   * Converts the session into an array of key-value pairs, encrypting sensitive data.
```
lizkenyon, line 321: We could explicitly say it is the access token that is getting encrypted.

packages/apps/shopify-api/runtime/crypto/__tests__/encrypt.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  const iv = cryptoLib.getRandomValues(new Uint8Array(12));
+  const key = await cryptoLib.subtle.generateKey(
+    {name: 'AES-GCM', length: 256},
+    true,
+    ['encrypt', 'decrypt'],
+  );
+
+  // WHEN
+  const encryptedValue = await encryptString('Test encrypted value', {key, iv});
```
lizkenyon, line 16: We could put 'Test encrypted value' in a variable

packages/apps/shopify-api/docs/guides/session-storage.md
```diff
@@ -294,4 +292,14 @@ const session = Session.fromPropertyArray(sessionProperties);
 >
 > The existing [SQL-based implementations](../../../session-storage#readme), i.e., MySQL, PostgreSQL and SQLite, convert it from seconds from storage. The remaining implementations do not change the retrieved `expires` property.
 
+### Encrypting data for storage
+
+If you want to encrypt your access tokens before storing them, the `Session` class provides two methods: `fromEncryptedPropertyArray` and `toEncryptedPropertyArray`.
+
+These behave the same as their non-encrypted counterparts, and take in a `CryptoKey` object. If a session is currently not encrypted in storage, these methods will still load it normally, and will encrypt them when saving them.
+Currently, only the `accessToken` is encrypted.
```
lizkenyon, line 300: Question: Should we be making it flexible to allow folks to encrypt more than just the accessToken?

packages/apps/shopify-api/lib/session/session.ts
paulomarg: Yeah I guess that would make sense - I don't see us encrypting anything other than sessions but it would keep the session clean. Let's do it.

packages/apps/shopify-api/docs/guides/session-storage.md
paulomarg: I thought of that too, but haven't thought of _how_ the app would convey that information. I think it would need to be a direct argument to `to`/`fromEncryptedArray` that the app would have to pass in to the session storage.

That would mean the app could do something like:

```ts
// Evergreen way of encrypting every column we allow
new PrismaSessionStorage(prisma, {encryptionKey: true, encryptColumns: 'all'});

// Explicitly select which columns
new PrismaSessionStorage(prisma, {encryptionKey: true, encryptColumns: ['accessToken', 'userId']});
```

paulomarg: It is fairly easy to implement though, I'm going to timebox it and see if I can get it off the ground.

PR #823, opened 2024-04-25
Fix package minification process

### LINKED ISSUE #802: @shopify/storefront-api-client not working via CDN
# Issue summary

  - Node version:
  - Operating system:
- [x] Set `{ logger: { level: LogSeverity.Debug } }` in my configuration, when applicable
- [x] Checked that this isn't an issue with a Shopify API

 I try the example using CDN as follows:

```
<script src="https://unpkg.com/@shopify/storefront-api-client@0.2.3/dist/umd/storefront-api-client.min.js"></script>

<script>
  const client = ShopifyStorefrontAPIClient.createStorefrontApiClient({...});
</script>
```

Note that I've also tried bumping to the latest version rather than the older version from the docs.

## Expected behavior

I should be seeing the same results as when using the NPM package. Here is a screenshot of it working fine when I tried this way.

Screenshot 2024-04-17 at 11 29 19 AM

## Actual behavior

I get weird results like the following:

**client.fetch()**
Screenshot 2024-04-17 at 8 30 26 AM

**client.request()**
Screenshot 2024-04-17 at 8 36 23 AM

Instantiate a client using the CDN package.

## Debug logs
See screenshots included above

paulomarg:

Closes #802 

As reported, client requests behaved strangely on the minified version of the SFAPI client, because the `fetch` function was being mangled and led to it being somehow overridden in the final build.

Ensuring that `terser` leaves the `fetch` function alone when minifying the code, so that requests still work. We already had a rule to preserve it as a function name, but it still failed when it was passed as a parameter, probably because terser couldn't know that was a function.

## Type of change

## Checklist

### DIFF
--- a/.changeset/gentle-apricots-work.md
+++ b/.changeset/gentle-apricots-work.md
@@ -1,0 +1,6 @@
+---
+"@shopify/graphql-client": patch
+"@shopify/storefront-api-client": patch
+---
+
+Fixed the minified build process to not mangle the `fetch` function, which led to requests failing in the final package.
--- a/config/rollup/rollup-utils.js
+++ b/config/rollup/rollup-utils.js
@@ -37,7 +37,14 @@ export function getPlugins({
       declaration: false,
       moduleResolution: 'Bundler',
     }),
-    ...(minify === true ? [terser({keep_fnames: new RegExp('fetch')})] : []),
+    ...(minify === true
+      ? [
+          terser({
+            keep_fnames: new RegExp('fetch'),
+            mangle: {reserved: ['fetch']},
+          }),
+        ]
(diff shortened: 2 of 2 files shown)


lizkenyon (approved):

andershagbard:
@paulomarg This is also an issue when just using the Storefront API client and webpack.

A better fix could be to rename the `fetch` function, so we dont have to add `keep_fnames` and `mangle`?

paulomarg:
I think that's a good callout - I just tested renaming our internal `fetch` objects to `fetchFn` and the UMD version works normally without the extra configs. I'll set up a PR with the change!

PR #846, opened 2024-04-30
Alias default to is_default for customer addresses

### LINKED ISSUE #754: Inconsistent default Attribute usage in CustomerAddress Type
`@shopify/shopify-api 9.0.2
`

`CustomerAddress` type is shared between different classes/instances but with different `default` attribute types:

`default` is a **property boolean type**
```ts
async getCustomerAddresses(customerId: number): Promise<CustomerAddress[]> {
    const r = await shopify.rest.CustomerAddress.all({
      session: customShopifyAppSession,
      customer_id: customerId,
    });
    return r.data;
 }

	const addresses = await getCustomerAddresses()
    console.log(addresses[0].default) // true or false
```
In this case, a item from r.data would be a `CustomerAddress` type, where the attribute default is defined as a method, but it's actually a boolean property. However, this is also possible:

`default` is a **method**.
```ts
  const customer = new shopify.rest.CustomerAddress({session});
  customer.customer_id = customerId;
  customer.id = existingAddress.id;
  await customer.default({});
```

The clearest case is: 

```ts
  const customer = new shopify.rest.CustomerAddress({session});
	customer.save();
	// customer.default will be a method
	
    customer.saveAndUpdate()
   // customer.default will be a boolean.
```

paulomarg:

Closes #754 

As pointed out in the issue, we currently have 2 instances of `default` in the `CustomerAddress` resource:
1. The method to set an address as default
2. The response field indicating whether an address is the default

Aliasing the read-only field to `is_default` so they don't overlap and cause breakages in apps.

Question: is this actually a breaking change? Are we breaking apps that expect `default` to be the boolean value in certain contexts?

## Type of change

## Checklist

### DIFF
--- a/.changeset/kind-vans-add.md
+++ b/.changeset/kind-vans-add.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-api": patch
+---
+
+Added an `is_default` field to `CustomerAddress` so it doesn't overlap with the existing `default()` method we provide in the class.
--- a/.changeset/kind-vans-add.md
+++ b/.changeset/kind-vans-add.md
@@ -3,3 +3,25 @@
 ---
 
 Added an `is_default` field to `CustomerAddress` so it doesn't overlap with the existing `default()` method we provide in the class.
+
+Before:
+
+```ts
+const address = await shopify.rest.CustomerAddress.find({ session, id: 1234 });
+// Boolean
+console.log(address.default);
+// Error - not a function
+await address.default();
+```
+
+After:
+
+```ts
+const address = await shopify.rest.CustomerAddress.find({ session, id: 1234 });
+// Boolean
+console.log(address.is_default);
+// Function
+await address.default();
+```
+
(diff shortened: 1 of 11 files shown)


.changeset/kind-vans-add.md
```diff
@@ -0,0 +1,7 @@
+---
+"@shopify/shopify-api": patch
+---
+
+Added an `is_default` field to `CustomerAddress` so it doesn't overlap with the existing `default()` method we provide in the class.
```
lizkenyon, line 5: Let's add an example to clarify

To access the default boolean field of the CustomerAddress
```
const customer = new shopify.rest.CustomerAddress({session});
console.log(customer.is_default) // false
```

lizkenyon (approved):
In the next API release do we want to make this the default functionality?

paulomarg:
> In the next API release do we want to make this the default functionality?

Yes I think so - the current behaviour is technically broken, even if only some scenarios.

PR #849, opened 2024-05-01
Authenticate customer account extension requests

lizkenyon:

Introduces a new API to authenticate requests from[ Customer Account Extensions](https://shopify.dev/docs/apps/customer-accounts).

* Customer account extensions can use networks access to make HTTP requests to apps, and other external sources. 
* These requests can use [session tokens to be authenticated](https://shopify.dev/docs/api/customer-account-ui-extensions/unstable/apis/session-token#examples) 

Because Checkout UI extensions and Customer Account extensions are authenticated the same way, this refactors the the authenticateCheckoutFactory to be useable for both.

TODO
* ~generate docs~
* ~generate changelog~
* ~test in dev app~

## Type of change

## Checklist

### DIFF
--- a/.changeset/kind-vans-add.md
+++ b/.changeset/kind-vans-add.md
@@ -1,0 +1,27 @@
+---
+"@shopify/shopify-api": patch
+---
+
+Added an `is_default` field to `CustomerAddress` so it doesn't overlap with the existing `default()` method we provide in the class.
+
+Before:
+
+```ts
+const address = await shopify.rest.CustomerAddress.find({ session, id: 1234 });
+// Boolean
+console.log(address.default);
+// Error - not a function
+await address.default();
+```
+
+After:
+
+```ts
+const address = await shopify.rest.CustomerAddress.find({ session, id: 1234 });
+// Boolean
+console.log(address.is_default);
+// Function
+await address.default();
+```
+
+To prevent breaking existing apps, this only happens when the `customerAddressDefaultFix` flag is enabled.
--- a/packages/apps/session-storage/shopify-app-session-storage-drizzle/package.json
+++ b/packages/apps/session-storage/shopify-app-session-storage-drizzle/package.json
@@ -54,7 +54,7 @@
     "@shopify/shopify-app-session-storage-test-utils": "^2.0.5",
     "@types/better-sqlite3": "^7.6.9",
     "better-sqlite3": "^9.5.0",
-    "drizzle-kit": "^0.20.12",
+    "drizzle-kit": "^0.20.17",
     "drizzle-orm": "^0.29.3",
     "mysql2": "^3.9.7",
     "pg": "^8.11.5"
--- a/packages/apps/shopify-api/future/flags.ts
+++ b/packages/apps/shopify-api/future/flags.ts
@@ -15,6 +15,12 @@ export interface FutureFlags {
    * Enable line item billing, to make billing configuration more similar to the GraphQL API.
    */
   lineItemBilling?: boolean;
+
+  /**
+   * Change the CustomerAddress classes to expose a `is_default` property instead of `default` when fetching data. This
+   * resolves a conflict with the default() method in that class.
+   */
+  customerAddressDefaultFix?: boolean;
 }
 
 /**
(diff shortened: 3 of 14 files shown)


packages/apps/shopify-app-remix/src/server/authenticate/public/customer-account/authenticate.public.customerAccount.doc.example.offers.ts
```diff
@@ -0,0 +1,34 @@
+// Most apps would load this from their database
+export function getOffers(shop: string, customerID: string) {
+    const offers: Record<string, any[]> = {
```
paulomarg, line 3: Nit, but is the tab width right here?

packages/apps/shopify-app-remix/src/server/authenticate/public/customer-account/authenticate.public.customerAccount.doc.example.ts
```diff
@@ -0,0 +1,18 @@
+import type {ActionFunctionArgs, LoaderFunctionArgs} from '@remix-run/node';
+import {json} from '@remix-run/node';
+
+import {authenticate} from '../shopify.server';
+import {getOffers} from '../offers.server';
+
+// The loader responds to preflight requests from Shopify
+export const loader = async ({request}: LoaderFunctionArgs) => {
+  await authenticate.public.checkout(request);
```
paulomarg, line 9: Wrong function name :)

packages/apps/shopify-app-remix/src/server/authenticate/public/customer-account/types.ts
```diff
@@ -0,0 +1,65 @@
...
+   * <caption>Using the decoded session token.</caption>
+   * <description>Get store-specific data using the `sessionToken` object.</description>
+   * ```ts
+   * // app/routes/public/my-route.ts
+   * import { LoaderFunctionArgs, json } from "@remix-run/node";
+   * import { authenticate } from "../shopify.server";
+   * import { getMyAppData } from "~/db/model.server";
+   *
+   * export const loader = async ({ request }: LoaderFunctionArgs) => {
+   *   const { sessionToken } = await authenticate.public.checkout(
```
paulomarg, line 33: Wrong method here too.

packages/apps/shopify-app-remix/src/server/authenticate/public/factory.ts
```diff
@@ -45,5 +49,5 @@ export function authenticatePublicFactory<
   authenticatePublic.checkout = authenticateCheckout;
   authenticatePublic.appProxy = authenticateAppProxy;
 
-  return authenticatePublic;
+  return authenticatePublic as AuthenticatePublic<Future>;
```
paulomarg, line 52: Why do we need to cast the type here? I would think that we're returning the appropriate type given the definitions.

packages/apps/shopify-app-remix/src/server/authenticate/public/extension/authenticate.ts
```diff
@@ -0,0 +1,50 @@
...
+  respondToBotRequest,
+  respondToOptionsRequest,
+  getSessionTokenHeader,
+  validateSessionToken,
+  ensureCORSHeadersFactory,
+} from '../../helpers';
+
+import {AuthenticateExtension, ExtensionContext} from './types';
+
+export function authenticateExtensionFactory(
```
paulomarg, line 13: I wonder if we can touch up the types a little bit so:
- we don't have this little bit of duplication
- we don't have to cast the types

Maybe we can have something like:

```suggestion
export function authenticateExtensionFactory<Options, Context>(): (Request, Options) => Promise<Context> { ... }
```

that way we wouldn't need the generic empty types, and the types returned by this function could be used without casting.

packages/apps/shopify-app-remix/src/server/authenticate/public/factory.ts
lizkenyon: We do also cast  the return above,
 I don't know if typescript was just struggling with evaluating the FF type in this scenario, but I was getting an error.

PR #936, opened 2024-05-24
Remove support for node 16

lizkenyon:

Fixes #0000 

## Type of change

## Checklist

### DIFF
--- a/packages/apps/shopify-api/tsconfig.json
+++ b/packages/apps/shopify-api/tsconfig.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": ".",
-    "rootDir": "."
+    "rootDir": ".",
   },
   "include": [
     "lib/**/*.ts",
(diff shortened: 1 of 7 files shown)


lizkenyon:
Looks like this is the issue causing our tests to fail.
https://github.com/microsoft/TypeScript/issues/52331

packages/apps/shopify-api/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "baseUrl": ".",
     "outDir": ".",
-    "rootDir": "."
+    "rootDir": ".",
+    "useDefineForClassFields": false
```
lizkenyon, line 7: Our tests were failing before adding this.
We can either use this option or update our test code.
https://github.com/microsoft/TypeScript/issues/52331

tsconfig.base.json
```diff
@@ -7,7 +7,7 @@
     "pretty": true,
     "module": "Node16",
     "moduleResolution": "Node16",
-    "target": "ES2021",
+    "target": "ES2022",
```
lizkenyon, line 10: Updated to this because of [this recommendation ](https://www.npmjs.com/package/@tsconfig/node18)

PR #947, opened 2024-05-28
[Chore] Increasing both test timeouts and concurrency

lizkenyon:
* The CI has been quite flaky because of the test timeouts
* Increasing the timeouts makes the test pass more often
* Increasing the concurrency too much is causing the tests to hang
* This seems to be a happy medium

Fixes #0000 

## Type of change

## Checklist


packages/apps/session-storage/shopify-app-session-storage-postgresql/src/__tests__/postgresql.test.ts
```diff
@@ -57,7 +57,7 @@ describe('PostgreSQLSessionStorage', () => {
         }
         return true;
       },
-      {interval: 500, timeout: 20000},
+      {interval: 500, timeout: 40000},
```
paulomarg, line 60: We may want to bump this a bit higher, still getting failures here.

PR #951, opened 2024-05-28
Remove v3_webhookAdminContext future flag

lizkenyon:

Remove v3_webhookAdminContext future flag and enable the feature by default.

## Type of change

## Checklist


packages/apps/shopify-app-remix/src/server/authenticate/webhooks/authenticate.ts
```diff
@@ -75,39 +71,16 @@ export function authenticateWebhookFactory<
...
+      params,
+      session,
+      handleClientError: handleClientErrorFactory({request}),
+    });
 
     return {
       ...webhookContext,
       session,
-      admin: admin as WebhookAdminContext<Future, Resources>,
+      admin: admin as WebhookAdminContext<Resources>,
```
paulomarg, line 83: Do we still need to cast this?

packages/apps/shopify-app-remix/src/server/authenticate/webhooks/types.ts
```diff
@@ -140,25 +139,10 @@ export interface WebhookContextWithoutSession<Topics = string | number | symbol>
...
-}
-
-export type WebhookAdminContext<
-  Future extends FutureFlagOptions,
-  Resources extends ShopifyRestResources,
-> =
-  FeatureEnabled<Future, 'v3_webhookAdminContext'> extends true
-    ? AdminApiContext<Resources>
-    : LegacyWebhookAdminApiContext<Resources>;
+export type WebhookAdminContext<Resources extends ShopifyRestResources> =
```
paulomarg, line 142: Do we still need this type? Could we just use AdminApiContext instead?

lizkenyon: Yes, we can return `AdminApiContext`, I wasn't sure if we wanted to keep the more specific type name.

paulomarg: Personally, I'd prefer not to alias types unless we had to for backward compatibility.

PR #957, opened 2024-05-29
Remove old server import path

lizkenyon:

With the release of version 4.0.0 we are removing the deprecated old import path.

## Type of change

## Checklist


packages/apps/shopify-app-remix/docs/MIGRATION_V3.md
```diff
@@ -32,3 +32,84 @@ export async function loader({request}: LoaderFunctionArgs) {
...
+With the addition of React component to this package, we've separated the exports between server and react code, as in:
+
+```diff
+import '@shopify/shopify-app-remix/server/adapters/node';
++ import {shopifyApp} from '@shopify/shopify-app-remix/server';
+- import {shopifyApp} from '@shopify/shopify-app-remix';
+import {AppProvider} from '@shopify/shopify-app-remix/react';
+```
+
+## Use the AppProvider component
```
lizkenyon, line 54: Also moved this here. We discussed having it in the template but we currently don't have any migration guides there. Would we still like to document it there?

paulomarg (approved):

packages/apps/shopify-app-remix/package.json
```diff
@@ -16,12 +16,6 @@
   "module": "./dist/esm/server/index.mjs",
   "types": "./dist/ts/server/index.d.ts",
   "exports": {
-    ".": {
```
lizkenyon, line 19: Is this all we needed to change? 

I tested on a dev app, and it was now giving me an error when I tried to import without `/server`

paulomarg: Yes, that is it! Unfortunately I couldn't leave a comment because this is a JSON file, but that's all we had to change.

PR #1046, opened 2024-06-13
fix(api-codegen-preset): strip extension from type import

thetrevdev:

@shopify/graphql-codegen doesn't expect a file extension when it is provided the importTypes option.  If it does it will produce:
`import type * as StorefrontTypes from './storefront.types.ts';`
when we really want
`import type * as StorefrontTypes from './storefront.types';`

This pull requests just leaves off the extension when passing the types file to the presetConfig

## Type of change

## Checklist

### DIFF
--- a/.changeset/nice-countries-collect.md
+++ b/.changeset/nice-countries-collect.md
@@ -1,0 +1,5 @@
+---
+'@shopify/api-codegen-preset': patch
+---
+
+type imports in generated codegen files will not include the extension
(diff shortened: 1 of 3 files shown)


paulomarg:
Hi, thanks for contributing this! Just so I can understand the issue better, would you mind sharing a configuration that leads into this case?

As a bit of extra context, the way both this and `@shopify/graphql-codegen` are set up, they'll read whether your output file for the `generates` step creates a `.ts` or a `.d.ts` file and produce the output accordingly, so I don't think we want to remove that logic.

If you found a config that breaks that behaviour though, we definitely want to fix it :)

thetrevdev:
```
shopifyApiProject({
      apiType: ApiType.Storefront,
      apiVersion: '2024-04',
      declarations: false,
    })
```
This will correctly generate a `storefront.generated.ts` and a `storefront.types.ts`
however 
`storefront.generated.ts`
will have 
`import * as StorefrontTypes from './storefront.types.ts';` 
Instead of
`import * as StorefrontTypes from './storefront.types';`

The unit tests show this behavior and I updated their expected output as well after this change

paulomarg (approved):
I see it now, thanks for this!

PR #1051, opened 2024-06-14
Loosen topic type check

gonzaloriestra:

Related to https://github.com/Shopify/develop-app-inner-loop/issues/1504

Use string as the type for topics instead of the list coming from the config or mandatory topics, so that we can use declarative webhooks in the Remix template.

I have tested this locally with a Remix app, and the topic is now a string, so the switch below doesn't complain when using a topic not registered in the code:
```
const { topic, shop, session, admin } = await authenticate.webhook(request);

switch (topic) {
  case "TOPIC_REGISTERED_IN_TOML":
```

## Type of change

## Checklist


packages/apps/shopify-app-remix/src/server/types.ts
```diff
@@ -257,10 +257,7 @@ interface Authenticate<Config extends AppConfigArg> {
    * };
    * ```
    */
-  webhook: AuthenticateWebhook<
-    RestResourcesType<Config>,
-    keyof Config['webhooks'] | MandatoryTopics
-  >;
+  webhook: AuthenticateWebhook<RestResourcesType<Config>, string>;
```
paulomarg, line 260: Does this leave behind an unused import to `MandatoryTopics`?

paulomarg (approved):

packages/apps/shopify-app-remix/src/server/types.ts
gonzaloriestra: No, it's still used here: 
https://github.com/Shopify/shopify-app-js/blob/0154c33e2a0d0ba681eed74fada0f0df2b34cf57/packages/apps/shopify-app-remix/src/server/authenticate/webhooks/authenticate.ts#L18

gonzaloriestra: Right, I think we can do it now: https://github.com/Shopify/shopify-app-js/pull/1051/commits/0b5a666acb020c94aefdfa1a575bab307b38ab63

paulomarg (approved):

PR #1085, opened 2024-06-20
Skip OAuth for Merchant Custom Apps

lizkenyon:
* Merchant custom apps or app in the Shopify Admin do not use OAuth
* In these scenarios we need to skip the OAuth flow, and create a session with the configured access token.

Questions:
* Should we make session storage optional for merchant custom apps?
* Should we make session storage optional for all app types using token exchange?
* What should we do when we receive a 401 from Shopify
* Should we run the afterAuth hook? If so when? If we were to run it in the authenticate message it would be running on every request??

## Type of change

## Checklist


packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
```diff
@@ -140,11 +140,6 @@ export function authStrategyFactory<
         shop,
       });
 
-      logger.debug('Request is valid, loaded session from session token', {
```
lizkenyon, line 143: In each Auth strategy we are logging if the request is valid. 
I don't think we need this as we aren't always loading session from session token?

packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/merchant-custom-app.ts
```diff
@@ -0,0 +1,62 @@
...
+    return session;
+  }
+
+  public handleClientError(request: Request): HandleAdminClientError {
+    // const {api, config, logger} = this;
+    return handleClientErrorFactory({
+      request,
+      onError: async ({error}: OnErrorOptions) => {
+        if (error.response.code === 401) {
+          // What should we do if we receive a 401 error?
```
lizkenyon, line 56: What should we do here if we receive a 401?

packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
sle-c: seems like we throw an exception in case the auth process fails, so I agree we don't need this log

sle-c (approved):
minor comments but otherwise LGTM 

Cartoon Thumbs Up GIF by SpongeBob SquarePants

sle-c:
Sorry I didn't see the questions. I don't have concrete answers, only some thoughts.

> Should we make session storage optional for merchant custom apps?
> Should we make session storage optional for all app types using token exchange?

If we want to make it optional, I'd vote to make it optional for all app types. Otherwise, if some requires session storage, some don't, it'll be best to require for all. IMO, all or nothing is easier to follow than some are required, some don't.

> Should we run the afterAuth hook? If so when? If we were to run it in the authenticate message it would be running on every request??

Seems like afterAuth is commonly used for registering webhooks, with declarative webhooks, do we still need to run webhook registrations?

packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
paulomarg: No objections here!

packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/__tests__/merchant-custom/admin-client.test.ts
```diff
@@ -0,0 +1,132 @@
...
+            apiVersion: '2021-01' as ApiVersion,
+            headers: {custom: 'header'},
+            tries: 2,
+          },
+        ),
+    },
+  ])(
+    '$testGroup re-authentication',
+    ({testGroup: _testGroup, mockRequest, action}) => {
+      it('returns 401 when receives a 401 response on fetch requests', async () => {
```
paulomarg, line 63: I agree with this approach, there is no way we can automatically recover from this scenario as the token would have to have changed in the admin side.

packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/merchant-custom-app.ts
paulomarg: Agreed, these requests should explode to remind the developer to update their token. Maybe even mention that in the error message?

paulomarg (approved):
Approving since it's mostly nits or things we agree on.

packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/merchant-custom-app.ts
lizkenyon: image

image

Throws an error and prints a log with a link to the token rotation docs.

packages/apps/shopify-app-remix/src/server/shopify-app.ts
```diff
@@ -71,12 +72,22 @@ export function shopifyApp<
...
   const params: BasicParams = {api, config, logger};
+
+  let strategy;
+  if (config.distribution === AppDistribution.ShopifyAdmin) {
+    strategy = new MerchantCustomAuth(params);
+  } else if (
+    config.future.unstable_newEmbeddedAuthStrategy &&
+    config.isEmbeddedApp
+  ) {
+    strategy = new TokenExchangeStrategy(params);
```
paulomarg, line 83: I'm surprised making the storage optional didn't require any changes in the pre-existing strategies - is that type incorrect in those?

paulomarg (approved):

PR #1087, opened 2024-06-21
[Feature]: allow responses from after auth hook bubble up to the loader

### LINKED ISSUE #1057: [shopify-app-remix] Managed installation and parallel route loading
# Issue summary

Hi,

I'm not sure if this is strictly a bug or if I simply need some advice how to handle the new managed installation flow correctly. 
I'm using the default remix template created by the Shopify CLI.
After the client has installed the app through managed installation, Shopify redirects the client to the app on route `/app`. Since it's the first request after installation, the `shopifyApp()` starts the *Authenticating admin request* flow and creates a new session and stores it on the session storage. This process is executed multiple times by each `loader()` function matching the current route. In this case, there are two loaders being executed for the request `/app`: 
- `/app/routes/app.tsx` 
- `/app/routes/app._index.tsx`. 

Both loaders run in parallel due to Remix's parallel route loading. Then, after *Authenticating admin request* has finished the `afterAuth()` hook is called. This hook would actually be called twice because of each loader. However, the managed installation via the `TokenExchangeStrategy` dedupes this call so it only runs once:

https://github.com/Shopify/shopify-app-js/blob/7045fe4e3a9c39b858ca8e8337978ee42d496b82/packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts#L159-L172

I use the `afterAuth` hook to create a database entities (user, shop, etc) in my own system. All other authenticated routes need this information and query this API in their respective loaders. Now the problem is that the managed installation through `TokenExchangeStrategy` doesn't provide a sequential authentication flow like with the previous `AuthCodeFlowStrategy`. The authentication in a managed installation seem to be started by multiple loaders in parallel, and the "fastest" loader calls the `afterAuth` hook. That means there are multiple routes rely on the information being created in the `afterAuth` hook but it may not be created **yet**. So I end up with lots of *not found* errors in loaders because they run before the `afterAuth` has finished. The errors disappear after a reload.

My first idea was to throw a redirect at the end of `afterAuth` to force the entire route to reload when the entity was created. But `redirect` is not exposed on the params of `afterAuth` and any error thrown inside `afterAuth` it caught by an outer [`try-catch`](https://github.com/Shopify/shopify-app-js/blob/ed8c30e016d061558dd4321accbcd492524bf569/packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts#L81-L93) anyway.

What is the right way to handle this issue? I think doing API calls inside `afterAuth` after installation sounds like a normal thing, right?

- [X] Upgraded to the latest version of the relevant packages
  - Node version:
  - Operating system:
- [X] Set `{ logger: { level: LogSeverity.Debug } }` in my configuration, when applicable
- [X Found a reliable way to reproduce the problem that indicates it's a problem with the package

## Expected behavior

After the managed installation, all loaders should wait for the `afterAuth` hook to be finished before returning.

## Actual behavior

The `-> app.tsx loader` and `-> app._index.tsx loader` are started in parallel. Both loaders start the *Authenticating admin request* flow. The first loaders calls the `-> afterAuth` hook while the second `loader` finishes with `<- app._index.tsx loader` before `<- afterAuth` and `<- app.tsx loader` have finished.

```sh
14:45:48 │ remix         │ [shopify-api/INFO] version 11.0.0, environment Remix
14:45:48 │ remix         │ -> app.tsx loader
14:45:48 │ remix         │ -> app._index.tsx loader
14:45:48 │ remix         │ [shopify-app/INFO] Authenticating admin request
14:45:48 │ remix         │ [shopify-app/INFO] Authenticating admin request
14:45:48 │ remix         │ [shopify-app/INFO] No valid session found
14:45:48 │ remix         │ [shopify-app/INFO] Requesting offline access token
14:45:48 │ remix         │ [shopify-app/INFO] No valid session found
14:45:48 │ remix         │ [shopify-app/INFO] Requesting offline access token
14:45:49 │ remix         │ [shopify-api/INFO] Creating new session | {shop: nebula-etl-test.myshopify.com, isOnline: false}
14:45:49 │ remix         │ [shopify-app/INFO] Running afterAuth hook
14:45:49 │ remix         │ -> afterAuth
14:45:49 │ remix         │ Creating user
14:45:49 │ remix         │ [shopify-api/INFO] Creating new session | {shop: nebula-etl-test.myshopify.com, isOnline: false}
14:45:49 │ remix         │ <- app._index.tsx loader
14:45:49 │ remix         │ No user found
14:45:50 │ remix         │ User created
14:45:50 │ remix         │ [shopify-api/INFO] Registering webhooks | {shop: nebula-etl-test.myshopify.com}
14:45:50 │ remix         │ <- afterAuth
14:45:50 │ remix         │ <- app.tsx loader
```

Repo: https://github.com/zirkelc/shopify-managed-installation

`pnpm shopify app dev`

zirkelc:

Fixes #1057  

This PR adds a `redirect` function to the `afterAuth()` hook so that we have more control of the request flow after authentication. If the `afterAuth()` hook throws a `Response`, this `Response` is allowed to bubble up to the respective Remix `loader` which started the auth process via `admin.authenticate()`. 

## Type of change

## Checklist

- [X] I have used `yarn changeset` to create a draft changelog entry (do NOT update the `CHANGELOG.md` files manually)

---

I tested this change on a dev store for the new the managed installation (TokenExchangeStrategy) and the old auth flow (AuthCodeFlowStrategy). I throw a `redirect` inside `afterAuth()` to redirect the user to the `/app/additional` page after installation. In case of managed installation, the redirect works fine. 

In case of the old auth flow, there are probably parameters missing on the request (query or header) that cause the app to do an additional roundtrip to be embedded. The logs indicate this with `[shopify-app/DEBUG] App is not embedded, redirecting to Shopify`. I assume that a `Response` thrown from `afterAuth` must go through this function: 
https://github.com/Shopify/shopify-app-js/blob/0bd38310c2686664faee8bd9145eca27c66ab9a0/packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/auth-code-flow.ts#L214-L218
which would change redirect URL:
https://github.com/Shopify/shopify-app-js/blob/0bd38310c2686664faee8bd9145eca27c66ab9a0/packages/apps/shopify-app-remix/src/server/authenticate/admin/helpers/redirect-to-shopify-or-app-root.ts#L16-L18

I could investigate the second case (auth flow) further if I wanted to, but I'll stop here for now.


packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.ts
```diff
@@ -140,14 +140,10 @@ export function authStrategyFactory<
         shop,
       });
 
-      logger.debug('Request is valid, loaded session from session token', {
-        shop: session.shop,
-        isOnline: session.isOnline,
-      });
-
```
zirkelc, line 147: This debug statement won't be printed if `strategy.authenticate` throws a `Redirect` from `afterAuth`, so I moved it to the `TokenExchangeStrategy`.

packages/apps/shopify-app-remix/src/server/authenticate/admin/helpers/trigger-after-auth-hook.ts
```diff
@@ -23,6 +25,7 @@ export async function triggerAfterAuthHook<
         params,
         authStrategy.handleClientError(request),
       ),
+      redirect: config.isEmbeddedApp ? redirectFactory(params, request) : remixRedirect,
```
zirkelc, line 28: This always passes a `redirect` function to `afterAuth`.
I tried to make the type similar to the `EmbeddedAdminContext` and `NonEmbeddedAdminContext` returned from `authenticate.admin()`, so that `redirect` is only available for embedded apps. But this would require lots of other type changes to the `AppConfig` to make the generics work. 

Alternatively, `redirect` could also be `undefined` for non-embedded apps.

packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts
```diff
@@ -78,14 +78,23 @@ export class TokenExchangeStrategy<Config extends AppConfigArg>
         newSession = onlineSession;
       }
 
+      logger.debug('Request is valid, loaded session from session token', {
+        shop: newSession.shop,
+        isOnline: newSession.isOnline,
+      });
```
zirkelc, line 84: Moved here from `authenticate.ts` to be logged if `handleAfterAuth()` throws

matteodepalo:
Hi @zirkelc thank you for this PR, we've put it in our backlog.

packages/apps/shopify-app-remix/src/server/authenticate/admin/helpers/trigger-after-auth-hook.ts
```diff
@@ -23,6 +25,9 @@ export async function triggerAfterAuthHook<
         params,
         authStrategy.handleClientError(request),
       ),
+      redirect: config.isEmbeddedApp
```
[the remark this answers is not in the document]
zirkelc: Sure, I removed the `redirect` function

packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts
zirkelc: I added the log to `strategies/auth-code-flow.ts`.

packages/apps/shopify-app-remix/src/server/config-types.ts
```diff
@@ -262,6 +262,7 @@ interface HooksConfig {
    * @param context - An object with context about the request that triggered the hook.
    * @param context.session - The session of the merchant that installed your app. This is the output of sessionStorage.loadSession in case people want to load their own.
    * @param context.admin - An object with access to the Shopify Admin API's.
+   * @param context.redirect - A function that redirects the user to a new page, ensuring that the appropriate parameters are set for embedded apps. Non-embedded apps will receive Remix's default redirect function.
```
paulomarg, line 265: Last nit: we should also remove this :)

paulomarg (approved):
I think this is ready to be merged! If you could please

- rebase / merge `main`
- remove this last nit

I'll merge this! Thanks!

zirkelc:
Thank you! :-)

paulomarg:
Thanks for the updates, and for helping us make the package better!

PR #1183, opened 2024-07-10
[FIX] Remix Polaris Link not allowing target _blank

### LINKED ISSUE #1153: [shopify-app-remix] AppProvider breaks Polaris Button component external links
# Issue summary

- [X] Upgraded to the latest version of the relevant packages
  - `@shopify/*` package and version: @shopify/shopify-app-remix 3.0.0
  - Node version: 20.12.2
  - Operating system: OSX
- [ ] Set `{ logger: { level: LogSeverity.Debug } }` in my configuration, when applicable
- [X] Found a reliable way to reproduce the problem that indicates it's a problem with the package

When an embedded app using the provided `<AppProvider>` component renders a Polaris `<Button>` component with the `url` and `external` properties set, the rendered button does not open in a new tab/window.

`<AppProvider>` overrides the LinkComponent, which is fine for links, but the Polaris [Button](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/Button/Button.tsx) component actually renders an [UnstyledLink](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/UnstyledLink/UnstyledLink.tsx) component when the `url` property is provided. The UnstyledLink normally sets the `target` on the rendered `<a>` when `external` is provided, but the intercepting `RemixPolarisLink` does not.

## Expected behavior

For the following Polaris react component:

image

The rendered html should include `target="_blank"`:

image

## Actual behavior

For the same Polaris react component as above, the rendered html does not include `target="_blank"`:

image

1. Create a new remix app using the shopify-app-remix-template template
2. Replace the `<script>` tag in `app/routes/app.tsx` with `<AppProvider>`, as instructed in the [migration docs](https://github.com/Shopify/shopify-app-js/blob/%40shopify/shopify-app-remix%403.0.0/packages/apps/shopify-app-remix/docs/MIGRATION_V3.md#use-the-appprovider-component)
3. Add a Polaris button component to the route, providing both the `url` and `external` properties

admirsaheta:

Fixes #1153 

By default, the AppProvider component from @shopify/shopify-app-remix overrides the standard Link component used for routing.
Polaris' Button component, when used with a url prop and set to open externally, relies on an internal UnstyledLink component.
This RemixPolarisLink component acts as an intermediary between Remix's Link and Polaris' Button to ensure external links open in a new tab.

## Type of change

## Checklist

### DIFF
--- a/.changeset/lovely-buses-glow.md
+++ b/.changeset/lovely-buses-glow.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-remix': major
+---
+
+Fixes RemixPolarisLink blocking \_blank targets
--- a/packages/apps/shopify-app-remix/src/react/components/RemixPolarisLink.tsx
+++ b/packages/apps/shopify-app-remix/src/react/components/RemixPolarisLink.tsx
@@ -9,6 +9,24 @@ import type {
 export const RemixPolarisLink = React.forwardRef<
   HTMLAnchorElement,
   LinkLikeComponentProps
->((props, ref) => (
-  <Link {...props} to={props.url ?? props.to} ref={ref} />
-)) as LinkLikeComponent;
+>((props, ref) => {
+  const {url, external, ...otherProps} = props;
+
+  // Handle external links with target="_blank"
+  if (url && external) {
+    return (
+      <a
+        href={url}
+        target="_blank"
+        rel="noopener noreferrer"
+        {...otherProps}
+        ref={ref}
+      >
+        {props.children}
+      </a>
+    );
+  }
+
+  // Use Remix Link for internal routing
+  return <Link {...otherProps} to={url ?? props.to} ref={ref} />;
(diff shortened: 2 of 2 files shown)


paulomarg:
Instead of using `external`, is there a reason not to use `target="_blank"`, as per [the Button docs](https://polaris.shopify.com/components/actions/button)?

I tested that in an embedded app and it opened in a new tab, as expected.

paulomarg:
Hey, thanks for contributing! If I'm understanding this correctly, the goal is to fix the `Button` component, right? I think we can just use `target=_blank` in the button for that effect, since it seems that the `external` field is being deprecated, at least from the link.

This worked as expected for me in an embedded app:

```tsx
<Button url="https://google.com" target="_blank">Google</Button>
```

paulomarg:
Awesome! Do you feel like this PR is still needed, or should we go ahead and close it? :)

PR #1199, opened 2024-07-11
[Docs] Better describe billing config useage amount

lizkenyon:

Fixes #0000 

## Type of change

## Checklist


.changeset/spotty-pandas-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/shopify-api': patch
```
lizkenyon, line 2: I put this as a patch because I technically changed the type, but in practice it is unchanged. Is the changeset necessary?

paulomarg: I think we could do without the changeset - the public API of the package hasn't changed. That being said, we have other releases lined up so this will go out either way :)

paulomarg (approved):
LGTM!

PR #1210, opened 2024-07-15
[Chore] Rename internal fetch GQL client functions

paulomarg:

As pointed out in https://github.com/Shopify/shopify-app-js/pull/823#issuecomment-2228319316, we might still the issue from #802 in certain setups.

Since we don't particularly care about the `fetch` name here, we can just rename the internal function so we don't confuse rollup.

## Type of change

## Checklist


lizkenyon (approved):

lizkenyon:
Can we remove the the rollup config?
```
keep_fnames: new RegExp('fetch'),
mangle: {reserved: ['fetch']},
```

PR #1236, opened 2024-07-18
Allow not filtering by plans in billing checks

paulomarg:

Shopify has introduced [managed pricing](https://shopify.dev/docs/apps/launch/billing/managed-pricing) for apps, so they don't need to set up the billing plans in their configs any more. They can configure billing directly when publishing the app, and use the billing API to look for subscriptions.

However, the current code in these packages forced apps to pass in a set of plans to filter by when querying the API, and those plans needed to match the configuration object.

Since no config object is needed when using managed billing, we should make it possible to call `check` without a config.

## Type of change

## Checklist

### DIFF
--- a/.changeset/perfect-countries-tickle.md
+++ b/.changeset/perfect-countries-tickle.md
@@ -2,8 +2,8 @@
 '@shopify/shopify-api': minor
 ---
 
-Added a new future flag `unstable_managedPricingSupport`, which will:
+Added a new future flag `unstable_managedPricingSupport` to support apps using [Shopify managed pricing](https://shopify.dev/docs/apps/launch/billing/managed-pricing), which will:
 
 - Change `billing.check` to always return an object.
 - Change `billing.check` and `billing.subscription` to work without a billing config.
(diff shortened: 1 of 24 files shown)


sle-c (commented):
I tried to tophat by creating a new shopify app with the remix template and managed billing. In the `app._index.tsx` route, I called 
```javscript
const billingCheck = await admin.billing.check();

  console.log("billingCheck", billingCheck);`
```

I got this error which I think should only shows if I didn't use managed billing 🤔 

19-44-li1lw-35qfe

I'm using the local shopify-app-remix package with this branch checked out

```json
# package.json

    "@shopify/shopify-app-remix": "/Users/user/src/github.com/Shopify/shopify-app-js/packages/apps/shopify-app-remix",
```

paulomarg:
I'm making some improvements to normalize this method a little bit more since it had 3 different return formats. I'll post the steps I took to test it when I'm done!

sle-c (approved):
Realize my mistake was that I didn't link the shopify-api package to my local package. I tried again and was able to run the `check` method successfully 🎉 

Screenshot 2024-07-22 at 10 21 44 AM

PR #1269, opened 2024-07-26
Fix dependabot auto-changeset workflow

paulomarg:
This PR fixes the workflow for automatically adding a changeset to dependabot PRs which wasn't checking out the repo, and therefore couldn't run the script.

### DIFF
--- a/.github/workflows/dependabot_auto_merge.yml
+++ b/.github/workflows/dependabot_auto_merge.yml
@@ -15,19 +15,26 @@ jobs:
         uses: dependabot/fetch-metadata@v1
         with:
           github-token: '${{ secrets.GITHUB_TOKEN }}'
-      - name: Create changeset if not dev dependency
+
+      - name: Checkout (not dev dep)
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot_auto_merge.yml
```diff
@@ -15,19 +15,26 @@ jobs:
         uses: dependabot/fetch-metadata@v1
         with:
           github-token: '${{ secrets.GITHUB_TOKEN }}'
-      - name: Create changeset if not dev dependency
+
+      - name: Checkout (not dev dep)
+        uses: actions/checkout@master
+        if: ${{ steps.metadata.outputs.update-type != 'version-update:semver-major' && steps.metadata.outputs.dependency-type != 'direct:development' }}
```
matteodepalo, line 21: Can we not checkout and create in the same step? It feels strange to have to repeat the `if`. If it gets messy disregard this comment.

matteodepalo (approved):
:+1: minor comment only.

PR #1286, opened 2024-07-30
Remove checkout call from workflow

paulomarg:
Another attempt at getting the workflow just right for changesets.


lizkenyon (approved):
I am assuming we are already on the correct branch?

paulomarg:
Yes, the checkout action pulls `github.head_ref`, which is the source branch for the PR.

PR #1288, opened 2024-07-30
Fix actions bot email address

paulomarg:
Now that PRs are being created properly, we just need to make sure the user configs are right for the actions bot.


lizkenyon (approved):
👋  🎲 🎲

PR #1368, opened 2024-08-14
Scopes API: Refactor the interface

pobed2:
Fixes https://github.com/Shopify/develop-app-runtime-primitives/issues/459

Updates the Scopes API to match the [API that was designed with the help of the UI TAG](https://github.com/Shopify/ui-api-design/blob/main/apis/Scopes/README.md)

## Type of change
Marking this as a patch given that it updates the API for an unreleased feature.

## Checklist

### DIFF
--- a/.changeset/dependabot-1366.md
+++ b/.changeset/dependabot-1366.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-session-storage-dynamodb': patch
+---
+
+Updated `@aws-sdk/client-dynamodb` dependencies
--- a/.changeset/dependabot-1367.md
+++ b/.changeset/dependabot-1367.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-session-storage-dynamodb': patch
+---
+
+Updated `@aws-sdk/util-dynamodb` dependencies
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@shopify/prettier-config": "^1.1.2",
     "@shopify/typescript-configs": "^5.1.0",
     "@types/jest": "^29.5.12",
-    "@types/node": "^20.14.12",
+    "@types/node": "^22.3.0",
     "@types/supertest": "^6.0.2",
     "eslint": "^8.57.0",
     "eslint-plugin-jest": "^28.8.0",
@@ -52,7 +52,7 @@
     "ts-jest": "^29.1.3",
     "ts-node": "^10.9.2",
     "tslib": "^2.6.3",
-    "turbo": "^2.0.12",
+    "turbo": "^2.0.13",
     "typescript": "5.5.4"
   },
   "dependencies": {},
--- a/packages/apps/session-storage/shopify-app-session-storage-dynamodb/package.json
+++ b/packages/apps/session-storage/shopify-app-session-storage-dynamodb/package.json
@@ -44,8 +44,8 @@
     "AWS"
   ],
   "dependencies": {
-    "@aws-sdk/client-dynamodb": "^3.598.0",
-    "@aws-sdk/util-dynamodb": "^3.598.0"
+    "@aws-sdk/client-dynamodb": "^3.631.0",
+    "@aws-sdk/util-dynamodb": "^3.631.0"
   },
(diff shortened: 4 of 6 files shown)


packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/query.ts
```diff
@@ -17,38 +17,33 @@ export function queryScopesFactory(
...
-    granted: {
-      required: grantedRequired,
-      optional: grantedOptional,
-    },
-    declared: {
-      required: declaredRequired.toArray(),
-    },
+    granted: granted.toArray(),
+    required: required.toArray(),
+    optional: optional.toArray(),
```
pobed2, line 47: I updated the return value of this method to match the ["ScopesDetail" return value](https://github.com/Shopify/ui-api-design/blob/main/apis/Scopes/README.md#check-read_customers-scope)

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/revoke.ts
```diff
@@ -37,8 +37,10 @@ export function revokeScopesFactory(
       });
     }
 
-    const scopesInfo = await fetchScopeInformation(admin);
-    return mapFetchScopeInformation(scopesInfo);
+    const scopesDetail = await fetchScopeDetail(admin);
+    return {
+      detail: mapFetchScopeDetail(scopesDetail),
```
pobed2, line 42: I updated the return value of `revoke` to match the [one designed with the tag](https://github.com/Shopify/ui-api-design/blob/main/apis/Scopes/README.md#revoke-write_orders-and-write_themes)

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/request.ts
```diff
@@ -4,7 +4,7 @@ import type {BasicParams} from '../../../types';
 import {redirectToInstallPage} from '../helpers/redirect-to-install-page';
 import {AdminApiContext} from '../../../clients';
 
-import {fetchScopeInformation} from './client/fetch-scopes-information';
+import {fetchScopeDetail} from './client/fetch-scopes-details';
 
 export function requestScopesFactory(
```
pobed2, line 9: The return value for `request` wasn't updated since this implementation simply redirects to the install page.

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/request.ts
```diff
@@ -14,7 +14,7 @@ export function requestScopesFactory(
   return async function requestScopes(scopes: string[]) {
     const {logger} = params;
 
-    logger.debug('Request optional scopes: ', {
+    logger.debug('Requesting optional scopes: ', {
       shop: session.shop,
       ...{scopes},
     });
```
paulomarg, line 20: Nit: can we simplify the context here?
```suggestion
    logger.debug('Requesting optional scopes: ', {shop: session.shop, scopes});
```

packages/apps/shopify-app-remix/src/server/types-contexts.ts
```diff
@@ -39,6 +43,9 @@ export type AdminContext<App = DefaultApp> = IAdminContext<
   ConfigComponents<ShopifyConfig<App>>['resources']
 >;
 
+export type ScopesContext<_App = DefaultApp> = IScopesContext;
+export type ScopesDetail = IScopesDetail;
```
paulomarg, line 47: Nice! Good to make sure we're exposing these type helpers out of the package too.

paulomarg (approved):
🔥

packages/apps/shopify-app-remix/src/server/types-contexts.ts
paulomarg: The goal of exporting these types here is to make it possible for apps to create functions that take those, like:

```ts
function myFn(scopes: ScopesContext<typeof shopify>) {
  scopes.revoke(...);
}
```

and have TS sort out the types correctly from that `shopify` instance's configs.

PR #1370, opened 2024-08-15
Fixed admin.redirect issue with relative URLs

paulomarg:

In the admin context `redirect` helper, we build a full app URL from a relative one, but we were using `request.url` instead.

When tunnelling, we might get an HTTP URL in `request.url`, which would trigger a protocol switch, which Safari blocks.

Using `config.appUrl` as the base URL instead.

## Type of change

## Checklist

### DIFF
--- a/packages/apps/shopify-app-remix/src/server/authenticate/admin/helpers/__tests__/redirect.test.ts
+++ b/packages/apps/shopify-app-remix/src/server/authenticate/admin/helpers/__tests__/redirect.test.ts
@@ -76,6 +76,28 @@ describe('Redirect helper', () => {
     );
   });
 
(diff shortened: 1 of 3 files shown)


packages/apps/shopify-app-remix/src/server/authenticate/admin/helpers/__tests__/redirect.test.ts
```diff
@@ -173,59 +213,52 @@ describe('Redirect helper', () => {
       );
     });
 
-    it('parses shopify admin routes', async () => {
+    it('returns an app bridge script on bounce requests', async () => {
```
paulomarg, line 216: I just moved a couple of tests around because they were in the "wrong" `describe` block, but they're untouched.

lizkenyon (approved):

PR #1376, opened 2024-08-16
[Feature] Create Usage Records for Usage Based Billing

lizkenyon:
Allow users to create ussage records for their usage based plans
* If a developer passes in a subscription line item ID, we will attempt to make the record on that subscription line it
* Otherwise we will query the subscriptions for the active useage billing
* We allow passing in an idempotency key if the developer would like. This is forwarded on to the API request

Fixes #0000 

## Type of change

## Checklist


packages/apps/shopify-api/lib/billing/check.ts
```diff
@@ -133,6 +133,7 @@ const HAS_PAYMENTS_QUERY = `
         id
         name
         test
+        status
```
lizkenyon, line 136: We were not returning the status from this method. So we had no way to know which subscription was actually active.

packages/apps/shopify-api/lib/billing/__tests__/create-usage-record.test.ts
```diff
@@ -0,0 +1,178 @@
...
+    const response = await shopify.billing.createUsageRecord({
+      session,
+      description,
+      price,
+      subscriptionLineItemId: subscriptionId,
+    });
+
+    expect(response).toEqual(
+      JSON.parse(Responses.USAGE_RECORD_CREATE_RESPONSE).data
+        .appUsageRecordCreate.appUsageRecord,
```
paulomarg, line 63: Nit: if we'll be using this response to test more than a couple of tests, we could consider exporting it as a separate const to make these `expect` calls a bit easier to read.

packages/apps/shopify-api/lib/billing/create-usage-record.ts
```diff
@@ -0,0 +1,172 @@
...
+
+interface CreateUsageRecordVariables {
+  description: string;
+  price: Money;
+  subscriptionLineItemId: string;
+  idempotencyKey?: string;
+}
+
+const CREATE_USAGE_RECORD_MUTATION = `
+mutation appUsageRecordCreate($description: String!, $price: MoneyInput!, $subscriptionLineItemId: ID!) {
```
paulomarg, line 28: Should we name this in a more "unique" way for tracking purposes? Could be a follow-up if we're not in other methods.

packages/apps/shopify-app-remix/src/server/authenticate/admin/billing/create-usage-record.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    try {
+      return await api.billing.createUsageRecord({
+        session,
+        description: options.description,
+        price: options.price,
+        isTest: options.isTest,
+        idempotencyKey: options.idempotencyKey,
+        subscriptionLineItemId: options.subscriptionLineItemId,
+      });
```
paulomarg, line 27: Any particular reason not to pass the options through?

```suggestion
      return await api.billing.createUsageRecord({
        ...options,
        session,
      });
```

packages/apps/shopify-api/docs/guides/billing.md
```diff
@@ -284,4 +284,16 @@ const canceledSubscription = await shopify.billing.cancel({
...
+| ---------------------  | ---------------------------- | :-------: | :-----------: | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `session`              | `Session`                    |    Yes    |       -       | The session to be used for this request                                                                                                                                          |
+| `description`          | `string`                     |    Yes    |       -       | The description of the usage record.                                                                                                                  |
+| `price`                | `Money`                      |    Yes    |       -       | The amount and currency to be charged                                                                                              |
```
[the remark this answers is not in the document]
lizkenyon: Yeah I can update the documentation, to clearly specify amount and currency.

> to pass the objects we use for the billing API in general through
With this are you suggesting we don't use the Money type, in our types?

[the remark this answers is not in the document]
paulomarg: I think we use it in responses, but not in any of the params for the methods, so this one stands out a bit.

PR #1396, opened 2024-08-20
Unpin graphql-config dependency

paulomarg:
As of v5.1.2, `graphql-config` no longer uses `minimatch` @ v10, which dropped support for node 18-. We can now unpin the dependency so we can continue to update it.


.changeset/chatty-dingos-sneeze.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
gonzaloriestra, line 2: Is this intended?

gonzaloriestra (approved):

.changeset/chatty-dingos-sneeze.md
paulomarg: Yeah, this won't trigger a release of this package, but I guess we should. Let me replace it.

lizkenyon (approved):

PR #1400, opened 2024-08-21
Expose billing request override types

paulomarg:

Fixes https://github.com/Shopify/shopify-app-template-remix/issues/817

We already had support for overriding values when requesting payment, but those types weren't being exposed to apps.

Tweaking the `request` method types so we're also exposing `lineItems` overrides, which were already supported.

## Type of change

## Checklist


packages/apps/shopify-app-remix/src/server/index.ts
```diff
@@ -10,6 +10,7 @@ export {
   LogSeverity,
   DeliveryMethod,
   BillingInterval,
+  BillingReplacementBehavior,
```
paulomarg, line 13: We didn't export this before, so it wasn't easy to access it from the remix template when needed.

PR #1430, opened 2024-08-27
Update revoke operation to return revoked scopes instead of all scopes details

JonathanJChang:
Currently on a successful revoke, we make a query fetch and return the entire scopes detail. However, we decided that it is better to return the same values as that of the graphql for consistency; which are the scopes that were successfully revoked.

In the revoke operation we return a list of revoked scopes instead of fetching the scope details and returning that.

## Type of change

## Checklist

### DIFF
--- a/.changeset/three-dryers-wave.md
+++ b/.changeset/three-dryers-wave.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-remix': patch
+---
+
+Return revoked scopes instead of querying for scopes after revoking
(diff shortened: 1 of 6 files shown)


paulomarg (approved):
This makes sense to me - I'm partial to just passing the GraphQL return through, as that makes it easier for folks to go from the documentation in .dev to their code 👍

bachhoant:
af83432d2890e82be7bbc6a092f6c9d3f7a68897

bachhoant (approved):

PR #1445, opened 2024-09-02
[Feature] Support per-shop dynamic access scopes for embedded apps.

mikek-ok:

We have a large and complex app, with various optional features. Our aim is to keep the access scopes for our end users (shops) to a minimally required set. As shops request access to more features within our app, we dynamically increase the access scopes required for that shop. Currently, scopes are set up globally per app, these changes provide a way to have different sets of scopes per shop.

- Change the `scopes` parameter on an app config to optionally take a callback of the form `(shop: string) => Promise<AuthScopes>`
- When a callback is used, scopes will be resolved passing in the current shop for a given session/flow, allowing for database lookups or otherwise constructing an appropriate set of access scopes

## Type of change

## Checklist

- [X] I have used `pnpm changeset` to create a draft changelog entry (do NOT update the `CHANGELOG.md` files manually)
  - _(I added in the callback sig against apps/shopify-api's readme, happy to add in extra docs against the remix/express readmes with a more detailed explanation as well)_
  
(cc: @henrytao-me @owlyowl)

### DIFF
--- a/.changeset/early-jars-remain.md
+++ b/.changeset/early-jars-remain.md
@@ -1,0 +1,7 @@
+---
+'@shopify/shopify-app-express': minor
+'@shopify/shopify-app-remix': minor
+'@shopify/shopify-api': minor
+---
+
+Support per-shop dynamic access scopes for embedded apps.
--- a/packages/apps/shopify-api/lib/__tests__/config.test.ts
+++ b/packages/apps/shopify-api/lib/__tests__/config.test.ts
@@ -2,6 +2,7 @@ import * as ShopifyErrors from '../error';
 import {validateConfig} from '../config';
 import {ConfigParams} from '../base-types';
 import {ApiVersion, LATEST_API_VERSION, LogSeverity} from '../types';
+import {AuthScopes} from '../auth';
 
 let validParams: ConfigParams;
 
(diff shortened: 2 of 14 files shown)


mikek-ok:
I have signed the CLA!

lizkenyon (commented):
Thank you for this! 

If you run `pnpm build-docs` it will regenerate our generated docs that will become the [Remix docs](https://shopify.dev/docs/api/shopify-app-remix/v3/entrypoints/shopifyapp).

We can also update the scope types in the [reference](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-api/docs/reference/shopifyApi.md#scopes) for the Shopify API package

lizkenyon:
Hi!

First thank you for taking the time to put up this PR. We appreciate your efforts to make the libraries better for the entire community.

I spoke with the team that handles App Auth, and they are going to be releasing the optional scopes feature for apps that use [Shopify managed app installation](https://shopify.dev/docs/apps/build/authentication-authorization/app-installation) soon, aiming for November. 

Because this will be the recommended way to accomplish this, we don't want to introduce this implementation of this feature into the library at this time.

Looking forward to any future contributions you may have.

owlyowl:
Hi Liz!
Thanks for taking the time to have a look at this.

Optional scopes is a very different use-case to dynamic scopes per merchant.

We're big on progressive enhancement of permissions as per the built for
Shopify guidelines and this is a feature that is at the core of our entire
auth and app flow, hence the reason for making this PR and having to fork
the current repo in the first place.

Will the intended changes for managed install support the functionally
proposed in the PR because we were under the impression from discussions
with our partner manager and Riley that it wouldn't.

Would love to arrange a video call at some stage to discuss further.

lizkenyon:
Hey @owlyowl! 

I am going to reflag this with the team that owns App Auth, to get you a more detailed response.

owlyowl:
Thanks so much Liz! Really appreciate it!

PR #1459, opened 2024-09-04
[Feature] Adds support for signal request option to the graphql-client

ecbrodie:

This PR adds support for a `signal`  request option to the graphql-client. This new option will simply serve as a pass-through for [the corresponding option on the fetch api](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#signal).

The initial consumer of this new option will be an internal library, but this change should serve the overall public community well too.

Nothing to add beyond what's stated above.

## Type of change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "@babel/plugin-transform-runtime": "^7.24.7",
     "@babel/preset-env": "^7.25.3",
     "@changesets/cli": "^2.27.7",
-    "@cloudflare/workers-types": "^4.20240815.0",
+    "@cloudflare/workers-types": "^4.20240903.0",
     "@jest/types": "^29.6.3",
     "@rollup/plugin-babel": "^6.0.3",
     "@rollup/plugin-commonjs": "^26.0.1",
@@ -52,7 +52,7 @@
     "ts-jest": "^29.1.3",
     "ts-node": "^10.9.2",
     "tslib": "^2.6.3",
-    "turbo": "^2.0.14",
+    "turbo": "^2.1.1",
     "typescript": "5.5.4"
   },
   "dependencies": {},
--- a/packages/apps/session-storage/shopify-app-session-storage-kv/package.json
+++ b/packages/apps/session-storage/shopify-app-session-storage-kv/package.json
@@ -51,7 +51,7 @@
     "@shopify/shopify-app-session-storage": "^3.0.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^4.20240815.0",
+    "@cloudflare/workers-types": "^4.20240903.0",
     "@shopify/shopify-app-session-storage-test-utils": "^3.0.4",
     "miniflare": "^3.20240806.1"
   },
--- a/packages/apps/shopify-api/package.json
+++ b/packages/apps/shopify-api/package.json
@@ -91,7 +91,7 @@
     "uuid": "^10.0.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^4.20240815.0",
+    "@cloudflare/workers-types": "^4.20240903.0",
     "@swc/core": "^1.7.11",
     "@types/express": "^4.17.13",
     "@types/jsonwebtoken": "^9.0.6",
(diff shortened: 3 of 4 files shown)


packages/api-clients/graphql-client/src/graphql-client/types.ts
```diff
@@ -86,10 +89,8 @@ export interface ClientConfig {
   readonly retries: Required<ClientOptions>['retries'];
 }
 
-export interface RequestOptions {
+export interface RequestOptions extends Omit<CustomRequestInit, 'body'> {
   variables?: OperationVariables;
-  url?: string;
```
melissaluu, line 91: Do you mean to change the `RequestOptions` interface or just add `signal` to it? This would change the interface since it'll remove `url` and add `method` that's from `CustomRequestInit`.

ecbrodie: Oh shoot, didn't mean to exclude url. Missed that it is not included by the new supertype. Will fix.

.changeset/khaki-scissors-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/graphql-client': patch
```
melissaluu, line 2: Since we're adding support for `signal`, would this make more sense to be a `minor` update?

ecbrodie: I originally had it as minor actually, but decided against it upon further thought. Since this is a type-only change, this PR is actually a complete no-op in terms of the code that we ship to NPM.

There is a fair argument to be made that this should be minor because we are changing the public API, although technically the same outcome with passing through `signal` could have already been done by JS consumers of this library anyway.

melissaluu:
@ecbrodie - it would be beneficial if we had some tests for passing in the `signal` to ensure client's `fetch`, `request` and `requestStream` functions correctly takes in and passes the `signal` through to the `http-fetch` at the end. ATM, the functions available in the client are wrappers around the standard HTTP fetch API and we do extra processing and we don't blindly pass through the `RequestOptions`  - I believe you'll need to update the code [here](https://github.com/Shopify/shopify-app-js/blob/main/packages/api-clients/graphql-client/src/graphql-client/graphql-client.ts#L109-L114) to correctly bring the `signal` object through to the final HTTP fetch

.changeset/khaki-scissors-look.md
ecbrodie: Given we have since debunked that there won't be a non-types change to ship, this is now being changed to a minor version bump.

ecbrodie:
@melissaluu thanks for catching that. I missed that these changes were done in an internal POC that was done for this change.

Latest push includes the required code changes to pass forward `signal` and what I believe is sufficient testing.

packages/api-clients/graphql-client/README.md
```diff
@@ -84,12 +84,14 @@ const client = createGraphQLClient({
...
+| headers?   | `Record<string, string \| string[]>` | Additional and/or replacement headers to be used in the request                                                                                                                                                                                      |
+| retries?   | `number`                             | Alternative number of retries for the request. Retries only occur for requests that were abandoned or if the server responds with a `Too Many Request (429)` or `Service Unavailable (503)` response. Minimum value is `0` and maximum value is `3`. |
+| signal?    | `AbortSignal                         | If this option is set, the request can be canceled by calling `abort()` on the corresponding  `AbortController`.                                                                                                                                     |
```
matteodepalo, line 93:
```suggestion
| signal?    | `AbortSignal`                         | If this option is set, the request can be canceled by calling `abort()` on the corresponding  `AbortController`.                                                                                                                                     |
```

matteodepalo (approved):
LGTM, only a small typo 👍

.changeset/khaki-scissors-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/graphql-client': minor
+---
+
+Adds support for signal request option
```
melissaluu, line 5:
```suggestion
Adds support for `signal` request option
```

ecbrodie: Oh, didn't know we have markdown formatting in changesets...will fix

melissaluu (approved):
Code looks good 👍  but let's update the PR description to match the current changes

PR #1506, opened 2024-09-13
[WIP] add `shopify-app-remix` scopes API documentation

RyanDJLee:
Addresses: https://github.com/Shopify/develop-app-runtime-primitives/issues/496

Spin url: https://shopify-dev.shopify-dev-86xc.ryan-d-lee.us.spin.dev/docs/api/shopify-app-remix/v3/apis/scopes

image
image
image
image

## Type of change
**Documentation**

## Checklist

### DIFF
--- a/packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.admin.doc.ts
+++ b/packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.admin.doc.ts
@@ -32,6 +32,7 @@ const data: ReferenceEntityTemplateSchema = {
     'EmbeddedAdminContext',
     'AdminApiContext',
     'BillingContext',
+    'ScopesApiContext',
   ],
   related: [
     {
--- a/packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/authenticate.admin.scopes.doc.ts
+++ b/packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/authenticate.admin.scopes.doc.ts
@@ -1,0 +1,28 @@
+import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
+
+const data: ReferenceEntityTemplateSchema = {
+  name: 'Scopes',
+  description:
+    'Contains functions used to manage optional scopes for your app.\n\nThis object is returned on authenticated Admin requests.',
+  category: 'APIs',
+  type: 'object',
+  isVisualComponent: false,
+  definitions: [
+    {
+      title: 'scopes',
+      description:
+        'Provides utilities that apps can use to manage optional scopes for the app using the Admin API.',
+      type: 'ScopesApiContext',
+    },
+  ],
+  jsDocTypeExamples: ['ScopesApiContext'],
+  related: [
+    {
+      name: 'Admin context',
+      subtitle: 'Authenticate requests from Shopify Admin.',
+      url: '/docs/api/shopify-app-remix/authenticate/admin',
+    },
+  ],
+};
+
+export default data;
--- a/packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
+++ b/packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
@@ -1,14 +1,226 @@
+/**
+ * The Scopes API enables embedded apps and extensions to request merchant consent for access scopes.
+ */
 export interface ScopesApiContext {
+  /**
+   * Queries Shopify for the scopes for this app on this shop
+   *
+   * @returns The scope details.
+   *
+   * @example
+   * <caption>Query for scopes details.</caption>
+   * <description>Call `scopes.query` to get scope details.</description>
+   * ```ts
+   * // /app/routes/app.query.tsx
+   *import type { ActionFunctionArgs } from "@remix-run/node";
+   *import { json } from "@remix-run/node";
+   *
+   *import { authenticate } from "../shopify.server";
+   *
+   *export async function action({
+   *  request,
+   *}: ActionFunctionArgs) {
+   *  const { scopes } = await authenticate.admin(request);
+   *
+   *  const response = await scopes.query();
+   *  return json({scopesDetail : response});
+   *}
+   * ```
+   * ```ts
+   * // /app/routes/**\/*.tsx
+   *const handleQueryClick = () => {
+   *  fetcher.submit({}, {
+   *    action: "query",
+   *    method: "POST",
+   *  });
+   *};
+   *
+   *return(
+   *  <Button onClick={handleQueryClick}>
+   *    Query
+   *  </Button>
+   *);
+   * ```
+   * ```ts
+   * // /app/shopify.server.ts
+   * import { LATEST_API_VERSION, shopifyApp } from "@shopify/shopify-app-remix/server";
+   *
+   * const shopify = shopifyApp({
+   *   future: {
+   *     unstable_optionalScopesApi: true,
+   *   },
+   *   // ...etc
+   * });
+   * export default shopify;
+   * export const authenticate = shopify.authenticate;
+   * ```
+   */
   query: () => Promise<ScopesDetail>;
-  request: (scopes: string[]) => Promise<void>;
-  revoke: (scopes: string[]) => Promise<RevokeResponse>;
+
+  /**
+   * Requests the merchant to grant the provided scopes for this app on this shop
+   *
+   * // TODO: I think we can omit the other error cases as they are pre-reqs (i.e. scopes deployed to Shopify, malformed access scope,.. )
+   *    We will add discoverable references to managing your scopes section to ensure developers understand what this feature and the pre-reqs are in case they discover it first via this page 
+   *    
+   * **Warning** This method performs a server-side redirect.
+   * // TODO: should we link to managing your scopes section to ensure they deployed?
+   * We recommend you do your own validation to ensure that the optional scopes being requested have not been granted.
+   *
+   * @example
+   * <caption>Request for optional scopes.</caption>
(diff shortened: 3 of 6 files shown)


packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
```diff
@@ -1,14 +1,222 @@
...
+   *  </Button>
+   *);
+   * ```
+   * ```ts
+   * // /app/shopify.server.ts
+   * import { LATEST_API_VERSION, shopifyApp } from "@shopify/shopify-app-remix/server";
+   *
+   * const shopify = shopifyApp({
+   *   future: {
+   *     unstable_optionalScopesApi: true,
```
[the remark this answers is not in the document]
allenchazhoor: We should also remove it from Step 2, which would make it more of a one-step process. Maybe no point in a numbered list then?

packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.admin.doc.ts
```diff
@@ -32,6 +32,7 @@ const data: ReferenceEntityTemplateSchema = {
     'EmbeddedAdminContext',
     'AdminApiContext',
     'BillingContext',
+    'ScopesApiContext',
```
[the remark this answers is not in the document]
kbav: +1, this page is problematic.
+1, not necessarily something we need to fix here, but would be great to rethink this page & its purpose

packages/apps/shopify-app-remix/docs/staticPages/index.doc.ts
```diff
@@ -166,6 +166,7 @@ const data: LandingTemplateSchema = {
         "\n\nIt takes advantage of [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)" +
         " to handle automatic app installations and scope updates, while using" +
         " [token exchange](https://shopify.dev/docs/apps/auth/get-access-tokens/token-exchange) to get an access token for the logged-in user." +
+        "\n\n If you wish to learn about Scopes Management and APIs, please read through [Managing Access Scopes](https://shopify.dev/docs/apps/build/authentication-authorization/app-installation/managing-access-scopes)" +
```
kbav, line 169:
```suggestion
        "\n\n If you wish to learn about scopes management and APIs, please read through [Manage access scopes](https://shopify.dev/docs/apps/build/authentication-authorization/app-installation/managing-access-scopes)" +
```

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/authenticate.admin.scopes.doc.ts
```diff
@@ -0,0 +1,28 @@
+import {ReferenceEntityTemplateSchema} from '@shopify/generate-docs';
+
+const data: ReferenceEntityTemplateSchema = {
+  name: 'Scopes',
+  description:
+    'Contains functions used to manage scopes for your app.\n\nThis object is returned on authenticated Admin requests.',
```
kbav, line 6:
```suggestion
    'Contains methods used to manage scopes for your app.\n\nThis object is returned on authenticated Admin requests.',
```

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
```diff
@@ -1,14 +1,212 @@
...
-export interface RevokeResponse {
-  revoked: string[];
+  /**
+   * Requests the merchant to grant the provided scopes for this app on this shop
+   *
+   * **Warning** This method performs a server-side redirect.
+   *
+   * @example
+   * <caption>Request consent from the merchant to grant the provided scopes for this app.</caption>
+   * <description>Call `scopes.request` to get request optional scopes.</description>
```
kbav, line 64:
```suggestion
   * <description>Call `scopes.request` to request optional scopes.</description>
```

kbav (commented):
Love seeing this and all the work you've put into it, Ryan! Thank you for working on this 🙌 

I'm in process of reviewing this and I have a dumb question... but where does the code for the examples on the right live?

For example, for the `query` method, I was going to propose a few changes
- rather than set up a Resource Route we POST to for retrieving data through `action`, we could show using a Resource Route for retrieving data through `loader`. The [`useRouteLoaderData` hook](https://remix.run/docs/en/main/hooks/use-route-loader-data) can be used for scenarios like this to then pull in data from other routes
- `request` method example was a bit convoluted (may be clearer to just have an `action` and button on the same page, instead of using a Resource Route) and not clear that it leads to a full page redirect.

I was about to write "similar to the above" for the `revoke` method but now I'm wondering if it would just be overall better to demonstrate these being used in `loader`s and `action`s right on the same page. It’s a bit less mental overhead for looking at the examples, and would change the examples from having 3 tabs/files to 2 tabs/files — importantly, with all the relevant details in the first/visible tab.

What do you think about that suggestion?

allenchazhoor:
Put up the recent changes here: https://shopify-dev.shopify-dev-3h44.allen-chazhoor.us.spin.dev/docs/api/shopify-app-remix/v3/apis/scopes

allenchazhoor:
There is also this [comment](https://github.com/Shopify/extensibility/pull/81#pullrequestreview-2340972820) from the App Bridge PR that I believe is relevant here

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
```diff
@@ -1,14 +1,166 @@
...
+   *
+   *import { authenticate } from "../shopify.server";
+   *
+   *export async function loader({
+   *  request,
+   *}: LoaderFunctionArgs) {
+   *  const { scopes } = await authenticate.admin(request);
+   *
+   *  const response = await scopes.query();
+   *  return json({scopesDetail : response});
```
kbav, line 26: idea: we can reinforce that `ScopesDetail` is what is returned here

```suggestion
   *  const scopesDetail = await scopes.query();
   *  return json({scopesDetail});
```

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
```diff
@@ -1,14 +1,166 @@
...
+   * <caption>Request consent from the merchant to grant the provided scopes for this app.</caption>
+   * <description>Call `scopes.request` to request optional scopes.</description>
+   * ```ts
+   * // /app/routes/app.request.tsx
+   *import type { LoaderFunctionArgs } from "@remix-run/node";
+   *import { json } from "@remix-run/node";
+   *
+   *import { authenticate } from "../shopify.server";
+   *import { AuthScopes } from "@shopify/shopify-api";
+   *export async function loader({
```
kbav, line 57: From the context below, it looks like this is being set up more specifically as an `action`: `request.formData()`, etc.

I think the `query` makes sense in a `loader`, but `request` perhaps best makes sense as part of a resource route's `action`, and perhaps the same for `revoke`. Worth testing for those

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
```diff
@@ -1,14 +1,166 @@
...
+   * export default shopify;
+   * export const authenticate = shopify.authenticate;
+   * ```
+   */
+  request: (scopes: Scope[]) => Promise<void>;
+
+  /**
+   * Revokes the provided scopes from this app on this shop
+   *
+   * **Warning:** This method throws an [error](https://shopify.dev/docs/api/admin-graphql/unstable/objects/AppRevokeAccessScopesAppRevokeScopeError) if the provided optional scopes contains a required scope.
```
kbav, line 87: If other API’s don't use this pattern of adding a bolded "Warning", I'd consider dropping the word entirely

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
```diff
@@ -1,14 +1,166 @@
...
+   *    const revokedResponse = await scopes.revoke(scopesToRevoke.toArray());
+   *    return json(revokedResponse);
+   *  } catch (e) {
+   *    console.log(JSON.stringify(e));
+   *    return json({});
+   *  }
+   *}
+   * ```
+   * ```ts
+   * // /app/shopify.server.ts
```
kbav, line 118: As it stands, I don't know what value these app/shopify.server.ts files are providing. (What are they saying that is different from the `shopify.server.ts` they get at app install? Are the first tabs sufficient to show how to set up & access the API?)

packages/apps/shopify-app-remix/src/server/authenticate/admin/scope/types.ts
```diff
@@ -1,14 +1,166 @@
...
+   * <caption>Revoke optional scopes.</caption>
+   * <description>Call `scopes.revoke` to revoke optional scopes.</description>
+   * ```ts
+   * // /app/routes/app.revoke.tsx
+   *import type { LoaderFunctionArgs } from "@remix-run/node";
+   *import { json } from "@remix-run/node";
+   *import { authenticate } from "../shopify.server";
+   *import { AuthScopes } from "@shopify/shopify-api";
+   *
+   *export async function loader({
```
kbav, line 99: see above comment re: using `loader` in the `request` example... I think this should be `action`

kbav (commented):
Sorry if I steered you wrong re: converting all to use `loader`; if I did, I meant to just request that for the `query` example. I think the other two could leverage a [resource route](https://remix.run/docs/en/main/guides/api-routes#resource-routes) pattern, much like Ryan & co. leveraged for making the optional scopes test app

PR #1563, opened 2024-09-25
fix: enable `returnUserData` when storing/loading sessions

MarcMogdanz:

We have a use case where we need to email address of the logged in Shopify user, that's why we've enabled the `useOnlineTokens: true` option when setting up our shopify remix app.

```typescript
const shopify = shopifyApp({
	// all other options
	useOnlineTokens: true,
	sessionStorage: new RedisSessionStorage(client),
});
```

Using online tokens populates the `onlineAccessInfo` field on the session which also includes the `associated_user` object. But currently the redis session adapter (and to my knowledge all other session adapters too) do not save this information. For the redis adapter, only the user id of the associated user is saved in the current setup.

The underlaying helper functions `toPropertyArray` and `fromPropertyArray` of the `@shopify/shopify-api` package (which is being used by the `@shopify/shopify-api`) already have support for handling the associated user data properly if the `returnUserData` parameter is set.

I've set the `returnUserData` to `true` for all session interactions which properly saves the user data to storage if present. Enabling this but using offline tokens (which is the default) does not impact the functionality and it behaves the same as it does already. I've also added a few tests to make sure for both cases the expected data is being saved and loaded.

Note: It may also make sense to switch the default of `returnUserData` in the underlaying `@shopify/shopify-api` so it's enabled for all session adapters, but since I can't really properly test all of them since we're only using the redis one I've defaulted to keep the change inside the redis adapter only.

## Type of change

## Checklist


.changeset/smart-phones-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/shopify-app-session-storage-redis': patch
```
matteodepalo, line 2: Would you mind making this a minor change instead of patch?

matteodepalo (commented):
Only a minor comment, but looks good 👍

matteodepalo (approved):
Thank you!

PR #1590, opened 2024-10-02
Simplify pollForTable and surface underlying error

surma:
The error `"Prisma ${this.tableName} table does not exist."` is very misleading when the error could be a plethora of things, a common one being providing the wrong credentials.

It’s crucial for us to surface the underlying error, at least as a console.log.

(Also I simplified the polling function.)


surma:
I'm so sorry. I did not notice that this PR wasn't passing. Please never hesitate to ping me on stuff like this.

Regarding your comment, `pollForTable` is async, so it definitely returns a promise, so that shouldn't be an issue.

Looking at the error, it seems to imply that the table name is expected to be `main.Session` instead of just `Session`, which seems surprising to me. Curious how my PR broke this test 🤔  I'll take a look.

surma:
Oh no the test is failing because I changed the error message inadvertantly 🙄

PR #1597, opened 2024-10-03
Use rollup json module instead of with keyword

KevinGrandon:

This repository currently uses the with keyword for importing JSON. This keyword is deprecated and causes problems on the latest LTS node versions in strict mode. 

Instead of using the with keyword, leverage the json rollup plugin for the same effect.
## Type of change

- [x ] Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
+++ b/packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
@@ -2,9 +2,6 @@ import {
   AppProvider as PolarisAppProvider,
   AppProviderProps as PolarisAppProviderProps,
 } from '@shopify/polaris';
-// This leads to some TS errors, but it does compile to something that works
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
 import englishI18n from '@shopify/polaris/locales/en.json';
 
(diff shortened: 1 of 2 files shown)


KevinGrandon:
I have signed the CLA!

packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
```diff
@@ -5,7 +5,7 @@ import {
 // This leads to some TS errors, but it does compile to something that works
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
```
matteodepalo, line 7: Is this @ts-ignore still needed after this change?

matteodepalo (commented):
Thank you for the PR! Looks good to me just one minor comment.

packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
KevinGrandon: Good call, it should not be. Removing.

KevinGrandon:
> Thank you for the PR! Looks good to me just one minor comment.

Thanks, I've addressed the comment.

PR #1651, opened 2024-10-17
Attach cause to error when validating store domain. Add documentation for storefront api client use with React Native

andy-liuu:

A merchant faced a bug that was difficult to track due to the error handling of our store domain url validation in our api client libraries. An unexpected error happened when using a client library in a React Native app, and the error was hidden from view.

Attaching the actual thrown error in the error.cause field will make it easier to debug unexpected issues. The original error message is preserved and users are provided with more information

- attach the actual error alongside the error message during validation
- add to the storefront api readme about this potential gotcha
- relax unit tests to only verify that the error message contains a string

## Type of change

## Checklist

### DIFF
--- a/packages/api-clients/graphql-client/src/api-client-utilities/validations.ts
+++ b/packages/api-clients/graphql-client/src/api-client-utilities/validations.ts
@@ -9,7 +9,7 @@ export function validateDomainAndGetStoreUrl({
 }) {
   try {
     if (!storeDomain || typeof storeDomain !== 'string') {
-      throw new Error(`A valid store domain must be provided`);
+      throw new Error();
     }
 
     const trimmedDomain = storeDomain.trim();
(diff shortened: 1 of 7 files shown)


andy-liuu:
Calling it a patch for the package whose errors are actually changed. No changelog for only documentation change or unit test relaxing

packages/api-clients/graphql-client/src/api-client-utilities/validations.ts
```diff
@@ -25,6 +25,7 @@ export function validateDomainAndGetStoreUrl({
   } catch (_error) {
     throw new Error(
       `${client}: a valid store domain ("${storeDomain}") must be provided`,
+      {cause: _error},
```
sle-c, line 28: can we remove the `_` in `_error` because it's used now. `_` indicates that it's not used.

sle-c (approved):
minor comment, otherwise LGTM

andy-liuu:
> wait it auto merged even though there's a comment 😿

dang. I'll sneak it in my next PR.

lizkenyon:
I also had a comment. We should add the details in the gotcha.


> Let's put the exact error that folks were seeing, so if they google it, this may come up.

```
Error: Storefront API Client: a valid store domain ("http://shop.myshopify.com/") must be provided
```

PR #1705, opened 2024-10-29
Remove future flag for released feature

zzooeeyy:
Closes:
- https://github.com/Shopify/develop-app-runtime-primitives/issues/540

When optional scopes feature is released, we have the option to either:
* Enable scopes future flag by default for new apps and communicate to apps to enable this future flag to use `Scopes` API OR
* Remove the future flag so that apps can use `Scopes` API automatically. 

I think removing the future flag makes more sense since we're adding new feature and not modifying existing behaviour of the library.

Removes future flag `wip_optionalScopesApi`.

### Top hatting
I tested by linking my app to a local version of the remix library. I can use the Scopes API without adding the future flag.

## Type of change

## Checklist

### DIFF
--- a/.changeset/four-bugs-fix.md
+++ b/.changeset/four-bugs-fix.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-remix': minor
+---
+
+Add Scopes API to Remix library by removing the future flag that enabled this feature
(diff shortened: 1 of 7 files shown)


lizkenyon:
Just to confirm the docs for this feature is in this [PR](https://github.com/Shopify/shopify-app-js/pull/1506)?

kbav (approved):
Thank you Zoey!

PR #1927, opened 2025-01-07
Feature: Extend GraphQL Client Package Types

### LINKED ISSUE #1813: Suggestion: Enhance Type Definitions for GQLExtensions and FetchResponseBody
### Background  
We are currently working on enhancing our GraphQL infrastructure and are evaluating the integration of this package. During our review, we encountered a few areas where the type definitions could better align with common usage patterns and improve type safety.

### Problem Description  

1. **`GQLExtensions` Type**  
   - Currently, `GQLExtensions` is typed as `Record<string, any>`.  
   - While this provides flexibility, it lacks specificity and omits fields like `cost`, which we believe should be included by default.  
   - Explicitly including common extension fields like `cost` would improve the developer experience and reduce ambiguity.

2. **`FetchResponseBody` Extensions Field**  
   - The `extensions,data,errors,headers` fields in `FetchResponseBody` are optional.  
   - However, in our testing, this field is always present in successful responses.  
   - Making it required would improve type safety and prevent unnecessary null checks.

### Proposed Solution  

- **Enhance the `GQLExtensions` Type**  
  - Include common fields like `cost` while maintaining flexibility to accommodate additional fields.

- **Update the `FetchResponseBody` Type**  
- Introduce a `SuccessClientResponse` type for responses containing `data` and `extensions`.  
- Introduce an `ErrorClientResponse` type for responses containing `errors`.  

This separation would allow for more explicit handling of success and error scenarios, aligning with common GraphQL response patterns.

### Question  
Would these changes align with your roadmap for the package? We believe this improvement could enhance the package’s usability for GraphQL clients.

Thank you for your consideration!

admirsaheta:

Fixes #1813  
These changes are intended to enhance type safety and provide more clarity in the codebase by using readonly where applicable. Additionally, an extended GQLExtensions type has been introduced to support more detailed cost-related metadata for GraphQL queries.

This pull request introduces the use of readonly in several places to prevent mutation of properties that should remain unchanged. 
Specifically, it includes:
- ```readonly``` applied to properties in interfaces such as ClientOptions, ClientConfig, RequestOptions, etc.
- An extension to the GQLExtensions type to provide more detailed cost-related information for GraphQL queries, including the requested and actual query costs, throttle status, and associated metadata.

## Type of change

## Checklist


byrichardpowell:
Hey @admirsaheta  thanks for your PR.

I like your changes and I'm just checking with some people internally that they are correct in 100% of use cases.  Also worth noting this may end up a major change, since code today could reassign RequestOptions or use a HTTP method that is not in the union.

byrichardpowell:
Hey again 👋 

So, we chatted internally and are happy to accept these changes with some caveats:

* This is a breaking change since code that reassigns, or code that uses a different HTTP method would fail.
* I think HTTP Request method should be scoped down to `'GET' | 'POST'`, since these are the only methods for GraphQL.

If you are happy to make those changes, we should be good to accept this PR.  Alternatively, if you prefer we are happy to take on these changes.  

To fairly set expectations, this is a relatively low priority for us, so it may be a while.

Suleimanlatrsh:
Thank you for this PR and for spotting the gap, the direction here (tighter, safer types for the GraphQL client) is still something we want. Since this has been open for a while and now conflicts with main, and there are a few things to work through (this needs to be a major/breaking change, the `OperationVariables` generic isn't wired through anywhere, the narrowed `graphQLErrors` type drops the standard `locations`/`path` fields, and `GQLExtensions` loses its top-level flexibility for non-cost extension keys), we're going to redo this on a fresh branch and land it with the next graphql-client major. Closing this one out, but genuinely appreciate the contribution and the idea.

PR #1980, opened 2025-01-17
Group related packages together for dependabot

Arkham:

Makes it so dependabot doesn't always create multiple separate PRs for related projects, i.e. one for @aws-sdk/client-dynamodb and one for @aws-sdk/util-dynamodb

Current groups:
- aws dynamodb packages
- remix-run packages
- prisma packages


lizkenyon (approved):
thanks for this improvement! ⭐

PR #2053, opened 2025-01-31
[Feature]: added support client credentials in shopify-api-js

fwaadahmad1:

 - Resolves [https://github.com/Shopify/develop-app-access/issues/617](https://github.com/Shopify/develop-app-access/issues/617)
 - Resolves [https://github.com/Shopify/develop-app-access/issues/608](https://github.com/Shopify/develop-app-access/issues/608)

 - No UI/UX Changes
 - Added support for expiry for offline tokens. (Still works for Offline Tokens with no expiry)
 - Introduces Client credentials token acquisition flow to `shopify-api` library.

### What should the reviewers focus on?

 - Existing `token_exchage` and `oauth` flows both use `create_session` file, see that changes do not affect the current behaviour of the library.
 - `create_session` now supports expiration for both online and offline tokens.

### Setup Instructions

1. Setup a Spin instance with constellation `partners:dev`
    - Run `spin up partners:dev --wait`
1. clone the [client-creds-tophat](https://github.com/fwaadahmad1/client-creds-tophat) app

#### Video

[Tophatting Client Credentials token acquisition flow](https://drive.google.com/file/d/1t54rZD5ZfCfoo-ngVs-c3X5Olyq2gnLj/view?usp=sharing)

  - if video is not enough, we can pair on tophatting together
#### Instructions

1. Make sure the future flag in `appConfig` in `shopify.server.js` file looks like this
```
...
future: {
    unstable_newEmbeddedAuthStrategy: false,
    removeRest: true,
  },
...
```
2. Install the app to your store, and check the database for an `access_token` with no expiry.
3. Click the button `initiate background job`, in your app
    - This button will trigger a background job on the remix server that will initiate the `client_credentials` flow from `shopify-api` library and then save it to the database.
    - The logs for this api call can be seen in the app's console
4. check the database for an `access_token` with expiry.
5. Change the future flag in `appConfig` in `shopify.server.js` file as follows
    - No need to run `shopify app dev` again, the app hot reloads
```
...
future: {
    unstable_newEmbeddedAuthStrategy: true,
    removeRest: true,
  },
...
```
7. Delete the token record from the database
8. Click `Generate a Product` button in the app to trigger a server action which will in turn trigger the token acquisition flow as token is missing from db.
    - this flow will follow `auth_code` flow as config was updated as such
9. Check the database for an `access_token` with no expiry.

## Type of change

## Checklist


packages/apps/shopify-api/lib/auth/oauth/client-credentials.ts
```diff
@@ -0,0 +1,62 @@
...
+const ClientCredentialsGrantType = 'client_credentials';
+
+export type ClientCredentials = (
+  params: ClientCredentialsParams,
+) => Promise<{session: Session}>;
+
+export function clientCredentials(config: ConfigInterface): ClientCredentials {
+  return async ({shop}: ClientCredentialsParams) => {
+    const cleanShop = sanitizeShop(config)(shop, true);
+    if (!cleanShop) {
```
lizkenyon, line 24: Is there a scenario when we are expecting this to happen? Or will `sanitizedShop` always throw `throw new InvalidShopError('Received invalid shop argument');`

packages/apps/shopify-api/lib/auth/oauth/__tests__/oauth.test.ts
```diff
@@ -596,6 +595,102 @@ describe('callback', () => {
     expect(responseCookies.shopify_app_session).toBeUndefined();
   });
 
+  test('callback throws an error when the request is done by a bot', async () => {
```
lizkenyon, line 598: Looks like we got some duplicated tests in this file.

lizkenyon (commented):
In general looks good! ⭐ 

Next step will be to add some documentation to the [guides](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-api/docs/guides/oauth.md) and [references](https://github.com/Shopify/shopify-app-js/tree/main/packages/apps/shopify-api/docs/reference/auth).

packages/apps/shopify-api/lib/auth/oauth/client-credentials.ts
fwaadahmad1: This code was left in after I was trying different approaches, Thanks for pointing this out! 
sanitizedShop will alway throw new InvalidShopError('Received invalid shop argument');

byrichardpowell (approved):

lizkenyon (approved):

PR #2064, opened 2025-02-04
[Feature] Add the ability to override 'keepalive' on fetch requests

mdentremont:
- Optional (backwards compatible) change which allows additional customizability on fetch requests

- Using keepalive will help use cases where we want a request to be made before a navigation occurs ([see here](https://github.com/Shopify/portable-wallets/issues/1891))

- Add "keepalive" to the request options and passes it into the "fetch" function

## Type of change

## Checklist

- [X] I have used `pnpm changeset` to create a draft changelog entry (do NOT update the `CHANGELOG.md` files manually)

### DIFF
--- a/.changeset/heavy-seas-shave.md
+++ b/.changeset/heavy-seas-shave.md
@@ -1,0 +1,21 @@
+---
+'@shopify/graphql-client': minor
+---
+
+Make fetch's keepalive configurable when making requests
+
+Example:
+```typescript
+const shopQuery = `
+  query ShopQuery {
+    shop {
+      name
+      id
+    }
+  }
+`;
+
+const {data, errors, extensions} = await client.request(shopQuery, {
+  keepalive: true,
+});
+```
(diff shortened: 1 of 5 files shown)


packages/api-clients/graphql-client/src/graphql-client/types.ts
```diff
@@ -79,6 +80,7 @@ export interface ClientOptions {
   headers: HeadersObject;
   url: string;
   customFetchApi?: CustomFetchApi;
+  keepalive?: boolean;
```
lizkenyon, line 83: I believe we  don't need this, we just need to add to the Request?

packages/api-clients/graphql-client/README.md
```diff
@@ -229,6 +230,24 @@ const {data, errors, extensions} = await client.request(shopQuery, {
...
+const shopQuery = `
+  query ShopQuery {
+    shop {
+      name
+      id
+    }
+  }
+`;
+
+// Will retry the HTTP request to the server 2 times if the requests were abandoned or the server responded with a 429 or 503 error
```
lizkenyon, line 245:
```suggestion
```

.changeset/heavy-seas-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/graphql-client': minor
+---
+
+Make fetch's keepalive configurable when making requests
```
lizkenyon, line 5: Let's add the example from the readme here as well!

lizkenyon (commented):
TY! ⭐

.changeset/heavy-seas-shave.md
mdentremont: Done!

packages/api-clients/graphql-client/src/graphql-client/types.ts
mdentremont: Good catch! Done.

PR #2007, opened 2025-01-23
Make library compatible with Remix v3_singleFetch

lizkenyon:
### WHY are these changes introduced?

Fixes #0000 

Remix has introduced the `v3_singleFetch` future flag.

_Single Fetch is a new data loading strategy and streaming format. When you enable Single Fetch, Remix will make a single HTTP call to your server on client-side transitions, instead of multiple HTTP calls in parallel (one per loader)._

_The [headers](https://remix.run/docs/en/main/route/headers) function is now used on both document and data requests when Single Fetch is enabled. You should use that function to merge any headers returned from loaders executed in parallel, or to return any given actionHeaders._

[Remix](https://remix.run/docs/en/main/route/headers#nested-routes) runs the headers function from starting with the parent route and then the sub routes

_Remix passes in the parentHeaders to your headers function. So users.tsx headers get passed to users.$userId.tsx, and then users.$userId.tsx's headers are passed to users.$userId.profile.tsx's headers._
_We don't want surprise headers in your responses, so it's your job to merge them if you'd like_.

#### Why is this a problem for the Remix package?
With this flag enabled, if there are cases where the package modifies the headers of a response, then a headers function export will need to exist on the route to properly merge the headers.

In the package headers are added/modified in the following scenarios
1) Redirecting to [accept billing](https://github.com/Shopify/shopify-app-js/blob/fcd67acc48d47e8b9486c48c8a812f88a08a6e31/packages/apps/shopify-app-remix/src/server/authenticate/admin/billing/helpers.ts#L6)
2) Retrying a [request when an ID token is invalid](https://github.com/Shopify/shopify-app-js/blob/fcd67acc48d47e8b9486c48c8a812f88a08a6e31/packages/apps/shopify-app-remix/src/server/authenticate/helpers/respond-to-invalid-session-token.ts#L11)

3) When using the [cors()](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-app-remix/src/server/authenticate/helpers/ensure-cors-headers.ts) helper function
4) When using the [liquid() helper](https://github.com/Shopify/shopify-app-js/blob/fcd67acc48d47e8b9486c48c8a812f88a08a6e31/packages/apps/shopify-app-remix/src/server/authenticate/public/appProxy/authenticate.ts#L67) function

[...]

PR #2123, opened 2025-02-19
[Patch] Don't set retry headers on extension requests

lizkenyon:

Fixes #0000 

Because of the functionality changes in [Make library compatible with Remix v3_singleFetch](https://github.com/Shopify/shopify-app-js/pull/2007), we need be able to set whether or not a request should be retried from `validateSessionToken`

Extension requests shouldn't be set to be retried, as there is no app bridge to intercept the request.

* Adds an optional param set if request should be retried.
* Defaults to true to keep current functionality
* Sets to to false for extensions

## Type of change

## Checklist


sle-c (approved):
confirmed no change in functionality, simply an optional param

PR #2157, opened 2025-02-26
Add buffer to session expiry

lizkenyon:

Fixes [Add buffer to online access token expiry](https://github.com/Shopify/first-party-library-planning/issues/706)

* We received several reports when a session has been marked as active, but then when it was actually used it was inactive, causing errors
* This pull requests adds a default param to the `isActive` method on the session with a default of 500 ms buffer for expiry
* If a session is within 500 ms of expiry it will now return false.

## Type of change

## Checklist


rezaansyed (approved):
LGTM!

sle-c (approved):

PR #2182, opened 2025-03-05
Fix Shopify local app development

alexanderMontague:

- currently local app development is broken for Shopify internal developers, as we can not render embedded apps running locally within the shopify admin
- With the transition to World, this adds the new local domains the local admin is using to the various places that need it within the app libraries
- Do we also need to support the ruby libraries for local dev? 

- add the necessary domains like we did for spin

## Type of change

## Checklist

### DIFF
--- a/packages/apps/shopify-api/lib/utils/shop-admin-url-helper.ts
+++ b/packages/apps/shopify-api/lib/utils/shop-admin-url-helper.ts
@@ -14,9 +14,12 @@ export function shopAdminUrlToLegacyUrl(shopAdminUrl: string): string | null {
   if (matches && matches.length === 2) {
     const shopName = matches[1];
     const isSpinUrl = shopUrl.includes('spin.dev/store/');
+    const isLocalUrl = shopUrl.includes('shop.dev/store/');
 
     if (isSpinUrl) {
       return spinAdminUrlToLegacyUrl(shopUrl);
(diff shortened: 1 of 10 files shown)


packages/apps/shopify-api/lib/utils/shop-admin-url-helper.ts
alexanderMontague, line 1: I wasn't sure if the compatibility helpers were needed at all but saw the spin ones and they were easy enough, so added for now

packages/apps/shopify-api/lib/utils/shop-validator.ts
```diff
@@ -7,7 +7,12 @@ import {shopAdminUrlToLegacyUrl} from './shop-admin-url-helper';
 export function sanitizeShop(config: ConfigInterface) {
   return (shop: string, throwOnInvalid = false): string | null => {
     let shopUrl = shop;
-    const domainsRegex = ['myshopify\\.com', 'shopify\\.com', 'myshopify\\.io'];
+    const domainsRegex = [
+      'myshopify\\.com',
+      'shopify\\.com',
+      'myshopify\\.io',
+      'shop\\.dev',
```
lizkenyon, line 14: Do we need to this addition here?
Or are folks also running the shop local?

Issue #1981, opened 2025-01-17
Codegen doesn't work out of the box with direct access graphql client


seaders:
I've set my graphql config up as standard,

```
import { ApiType, shopifyApiProject } from "@shopify/api-codegen-preset"

export default {
  // For syntax highlighting / auto-complete when writing operations
  schema: "https://shopify.dev/admin-graphql-direct-proxy/2024-10",
  documents: ["./src/app/**/*.{js,ts,jsx,tsx}"],
  projects: {
    // To produce variable / return types for Admin API operations
    default: shopifyApiProject({
      apiType: ApiType.Admin,
      apiVersion: "2024-10",
      documents: ["./src/app/**/*.{js,ts,jsx,tsx}"],
      outputDir: "./src/__generated__"
    })
  }
}
```

And, it generates everything perfectly correctly.

My issue is, from the app's frontend, I'm creating just about the simplest graphQL client for direct access,

```
import { createGraphQLClient } from "@shopify/graphql-client"
import { apiVersion } from "./consts"

export const graphQLClient = createGraphQLClient({
  url: `shopify:admin/api/${apiVersion}/graphql.json`,
  headers: {}
})
```

This works, 100% of the time for actually querying against, and getting the correct data back, and I also get correct typehinting when writing graphQL,

<img width="1293" alt="Image" src="https://github.com/user-attachments/assets/bfe5c099-f870-47cf-a330-c8dcdd5b8a92" />

(I'm deliberately including the incorrect image here to show that that too is working right)

Problem is?  No response is types, and everything is any,

<img width="334" alt="Image" src="https://github.com/user-attachments/assets/d81faaa4-86de-45ff-9a4f-f78f2fbccd24" />

Thing is, I know exactly the cause now, due to the generated code in `admin.generated.d.ts`.  At the bottom they declare a module as an "addon" to the client which will be making the calls,

```
...
declare module '@shopify/admin-api-client' {
  type InputMaybe<T> = AdminTypes.InputMaybe<T>;
  interface AdminQueries extends GeneratedQueryTypes {}
  interface AdminMutations extends GeneratedMutationTypes {}
}
```

Problem is, again, this is direct access, from the frontend, via the generic shopify graphQL client, NOT a backend-ish `@shopify/admin-api-client`.

I've tried setting `module` in the settings to "@shopify/graphql-client" and while it did generate the output for that,

```
declare module '@shopify/graphql-client' {
[...]

PR #2183, opened 2025-03-05
[Docs] [Codegen] Document using types with direct API access

lizkenyon:

[Related to Codegen doesn't work out of the box with direct access graphql client](https://github.com/Shopify/shopify-app-js/issues/1981)
* Because direct API access uses the Fetch API directly, the response types cannot be automatically inferred, like when using the `admin.graphql`
* Developers could also consider creating a [custom wrapper client](https://github.com/Shopify/graphql-codegen?tab=readme-ov-file#making-graphql-clients) around fetch

## Type of change

## Checklist

### DIFF
--- a/packages/api-clients/api-codegen-preset/README.md
+++ b/packages/api-clients/api-codegen-preset/README.md
@@ -333,3 +333,38 @@ To make your development flow faster, you can pass in the `--watch` flag to upda
 ```sh
 npm run graphql-codegen -- --watch
 ```
+
+### Using Generated Types with Direct API Access
+
+With Shopify App Bridge, you can use [Direct API Access](https://shopify.dev/docs/api/app-bridge-library#direct-api-access) to make API requests from your apps frontend directly to the Admin API.
+
+Because you are using the Fetch API directly, the responses cannot be automatically typed.
+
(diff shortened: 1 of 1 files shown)


packages/api-clients/api-codegen-preset/README.md
```diff
@@ -333,3 +333,38 @@ To make your development flow faster, you can pass in the `--watch` flag to upda
 ```sh
 npm run graphql-codegen -- --watch
 ```
+
+### Using Generated Types with Direct API Access
+
+With Shopify App Bridge, you can use [Direct API Access](https://shopify.dev/docs/api/app-bridge-library#direct-api-access) to make API requests from your apps frontend directly to the Admin API.
+
+Because you are using the Fetch API directly, the responses cannot be automatically typed.
```
byrichardpowell, line 341: Reading this seems to suggest that you are on your own.  Is it worth elaborating? E.g:

```
Because you are using the Fetch API directly, the responses cannot be automatically typed.  Here is how to manually type them using the automatically generated types
```

byrichardpowell (approved):
Approving so you can ship without a second review

PR #2321, opened 2025-04-09
Use the same ts-jest version in resolutions and overrides

byrichardpowell:
Before these changes, the following command would fail:

```
npx @changesets/cli status --since="origin/main"
```

With this error:

```
npm ERR! code EOVERRIDE
npm ERR! Override for ts-jest@^29.3.1 conflicts with direct dependency
```

This is the cause of CI failing for all the dependabot upgrade PR's.

Updates resolutions and overrides for `ts-jest`.  Previously the version did not match the version is `devDependencies`, which was the root of the problem. Now the versions always match.

## Type of change

## Checklist

- ~~[ ] I have added/updated tests for this change~~
- ~~[ ] I have documented new APIs/updated the documentation for modified APIs (for public APIs)~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,12 +60,12 @@
     "jest": "^29.7.0",
     "jest-circus": "^29.7.0",
     "node-fetch": "^2.6.7",
-    "ts-jest": "^29.3.0"
+    "ts-jest": "^29.3.1"
   },
   "overrides": {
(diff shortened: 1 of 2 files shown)


sle-c (approved):
before they were both 29.3.0, I wonder why they don't match 🤔

Issue #2081, opened 2025-02-08
shopify app remix: how to hook into "install" event


redochka:
app created using shopify cli 3.74

I have `shopify.server.ts` where we define the `authPathPrefix`:

```
const shopify = shopifyApp({
  apiKey      : process.env.SHOPIFY_API_KEY,
  apiSecretKey: process.env.SHOPIFY_API_SECRET || "",
  apiVersion  : apiVersion,
  scopes      : process.env.SCOPES?.split(","),
  appUrl      : process.env.SHOPIFY_APP_URL || "",
  //It does not seem to be used...
  authPathPrefix: "/auth",
  sessionStorage: new RedisSessionStorage(process.env.UPSTASH_REDIS_URL),
  distribution  : AppDistribution.AppStore,
  future        : {
    unstable_newEmbeddedAuthStrategy: true,
    removeRest                      : true,
  },
  ...(process.env.SHOP_CUSTOM_DOMAIN
    ? {customShopDomains: [process.env.SHOP_CUSTOM_DOMAIN]}
    :{}),
});
```

And i have this `app/routes/auth.$.tsx`

```
import type {LoaderFunctionArgs} from "@remix-run/node";
import {authenticate} from "../shopify.server";

//it does not seem to be called at all
export const loader = async ({request}: LoaderFunctionArgs) => {
  console.log("inside auth.$.tsx");
  await authenticate.admin(request);
  return null;
};
```

I expected `console.log("inside auth.$.tsx");` to be displayed at install time, but nothing.

Where is shopify auth being done? Sorry .. too much hidden magic.

lizkenyon:
Hi there 👋 

> I expected console.log("inside auth.$.tsx"); to be displayed at install time, but nothing.
In the shopifyApp config object `unstable_newEmbeddedAuthStrategy` this enables the [Token Exchange OAuth strategy](https://shopify.dev/docs/apps/build/authentication-authorization/access-tokens/#token-exchange). So in your app the `/auth` path won't be used, it will instead load the app in the `/app.tsx` route, see that it is missing the access token and make a request to receive a new one and store it.

> how to hook into "install" event

After the package has stored the new access token it can trigger the `afterAuth` [hook](https://shopify.dev/docs/api/shopify-app-remix/v3/entrypoints/shopifyapp#shopifyapp-propertydetail-appconfig).

[...]

PR #2351, opened 2025-04-16
feat(shopify-app-remix): store and return cached promise from idempotent promise handler

zirkelc:

Relates to #1057 and #2081

When using Shopify managed installation (`unstable_newEmbeddedAuthStrategy`), the authentication effectively happens __multiple__ times by each Remix `loader()` that calls `await authenticate.admin(request)`. For example, if the `/app` routes matches an layout route and a normal route, it will run two loaders in parallel. However, the `afterAuth` hook should only be called once during authentication, because this hook might create other resources, e.g. creates a user in the database. Since every loader does authentication, the `afterAuth` has to be de-duplicated with an [idempotent promise handler](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-app-remix/src/server/authenticate/helpers/idempotent-promise-handler.ts):
https://github.com/Shopify/shopify-app-js/blob/b8d25d087ec99e2a3aff0c4034e1d9290a541f3d/packages/apps/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts#L171-L185

This function uses the session token to avoid executing the same `afterAuth` multiple times by each loader. Only the first one runs the hook, every ones thereafter resolve with an empty promise.

However, this creates a condition where the first loader waits for the `afterAuth` hook to complete, while the other loaders continue with their execution after `await authenticate.admin(request)`. In other words, you cannot be sure if the hook did already complete after `await authenticate.admin(request)`. For example, if you create a user in a database and you need to access this user in the loaders, this user will not be available in all loaders.

This PR changes the [idempotent promise handler](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-app-remix/src/server/authenticate/helpers/idempotent-promise-handler.ts) by storing the promise returned from the `afterAuth` hook in a map. Every loader that uses `await authenticate.admin(request)` will wait for the completion of the hook. The hook itself will still be only executed once, because we still maintain only one promise, but this promise might be awaited multiple times.

## Type of change

## Checklist

### DIFF
--- a/packages/apps/shopify-app-remix/src/server/authenticate/helpers/__tests__/idempotent-promise-handler.test.ts
+++ b/packages/apps/shopify-app-remix/src/server/authenticate/helpers/__tests__/idempotent-promise-handler.test.ts
@@ -17,12 +17,12 @@ describe('IdempotentPromiseHandler', () => {
     const promiseHandler = new IdempotentPromiseHandler();
 
     // WHEN
-    promiseHandler.handlePromise({
+    await promiseHandler.handlePromise({
       promiseFunction,
       identifier: 'first-promise',
     });
 
-    promiseHandler.handlePromise({
+    await promiseHandler.handlePromise({
       promiseFunction,
       identifier: 'first-promise',
     });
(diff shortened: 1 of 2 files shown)


byrichardpowell:
Hey @zirkelc 👋 

Thanks so much for you thoughtful contribution.

Whilst I totally appreciate the issue you are addressing here, we think React Routers Middleware are a better solution to the problem.  These are sill marked as unstable, but once they are stable we'd like to release middleware support.  As such I'm going to close this PR.

I appreciate this might be disappointing but I do thank you for your contribution.

PR #2375, opened 2025-04-24
test

alex-page:


byrichardpowell:
Closing as we added some code to the App Provider to support Polaris Web Components, so we shouldn't remove it.  If we do navigation could break for `<s-link>`

PR #2606, opened 2025-06-23
Tweak authenticate pos

byrichardpowell:

`authenticate.public.pos()` was not the right place for pos, since requests from a POS terminal do not come from a public URL.

1. Move `autenticate.public.pos()` to `authenticate.pos()`
2. Export the POSContext type so it can be easilyl consumed by apps.

## Type of change

N/A: Package is not released yet

## Checklist

- ~~[ ] I have used `pnpm changeset` to create a draft changelog entry (do NOT update the `CHANGELOG.md` files manually)~~ - NA: Package is not released yet:


packages/apps/shopify-app-react-router/src/server/types-contexts.ts
```diff
@@ -56,6 +57,8 @@ export type FlowContext<App = DefaultApp> = IFlowContext<
   ConfigComponents<ShopifyConfig<App>>['resources']
 >;
 
+export type POSContext = IPOSContext;
```
lizkenyon: Nit: I believe we could just re-export without the intermediate

lizkenyon (approved):

PR #2683, opened 2025-07-16
[GraphQL Client] Improves error handling of malformed JSON responses

ecbrodie:

Fixes an issue in the `graphql-client` library for handling JSON responses that have a body that's malformed JSON and thus causes an error when `Response.json()` is invoked.

Currently, the raw error from calling `json()` will bubble up to the library consumer when `client.request()` is invoked. Instead, we should have a proper [`ResponseErrors`](https://github.com/Shopify/shopify-app-js/blob/main/packages/api-clients/graphql-client/README.md#responseerrors) object sent back to the consumer.

Specifically, the following two changes were implemented:

- Added a missing `await` to a return value in the `generateRequest` method. This was a classic case of forgetting to `return await` in a try block, which led to the `catch` clause in the method being completely skipped.
- Enhanced the errors object in the case of a malformed JSON response to include more fields from the `ResponseErrors` type. In particular, the raw `response` object is now being sent back so it can be handled accordingly by the consumer.

## Type of change

## Checklist

### DIFF
--- a/.changeset/curly-masks-feel.md
+++ b/.changeset/curly-masks-feel.md
@@ -1,0 +1,8 @@
+---
+'@shopify/shopify-app-react-router': patch
+'@shopify/graphql-client': patch
+'@shopify/shopify-app-remix': patch
+'@shopify/shopify-api': patch
+---
+
+Resolve bug with signal option on requests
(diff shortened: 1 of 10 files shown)


wshanks01 (approved):

saltamay (approved):
🙌

lizkenyon (approved):
Thank you! ⭐

PR #2711, opened 2025-09-02
Remove UUID jsonwebtoken require node 20

byrichardpowell:

We want to reduce maintenance.  We can do that by reducing dependencies. 

1. Unify on `jose`, rather than using both `jose` and `jsonwebtoken`
2. Remove `uuid` in favor of `globalThis.crypto.randomUUID()`
3. Require Node >=20 for Remix, RR, Express & API packages.

`jose` v5 requires Node >= 18, but Node 18 is close to LTS expiry and the CLI requries Node 20. So we decided to require Node 20 rathe than incur more major changes in future.

## Type of change

## Checklist

- ~~[ ] I have added/updated tests for this change~~ - N/A - No functional changes
- ~~[ ] I have documented new APIs/updated the documentation for modified APIs (for public APIs)~~ - N/A - No API changes

### DIFF
--- a/.changeset/calm-queens-pay.md
+++ b/.changeset/calm-queens-pay.md
@@ -1,0 +1,8 @@
+---
+'@shopify/shopify-app-react-router': major
+'@shopify/shopify-app-express': major
+'@shopify/shopify-app-remix': major
+'@shopify/shopify-api': major
+---
+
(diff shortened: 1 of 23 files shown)


byrichardpowell:
Closing.  Going to reopen with seperate PR's for each dependency to make it easier to review

PR #2740, opened 2025-09-03
Node 20 (remove crypto, node-fetch, jsonwebtoken)

byrichardpowell:

We want to reduce maintenance by removing dependencies.

1. Remove uuid in favor of globalThis.crypto.randomUUID()
2. Remove the crypto adapter setup, since we now require Node 20, which is the first Node version to support `globalThis.crypto.randomUUID()`.  Since all other runtimes support this, the adapter is no longer needed.
3. Require Node >=20 for API package.  Since Remix, RR & Express depends on the API package, this will be a breaking change for all 3.
4. Update CI steps to use Node 20 and above

## Type of change

## Checklist

- [x] I have added/updated tests for this chang
- ~~[ ] I have documented new APIs/updated the documentation for modified APIs (for public APIs)~~ - N/A, no changes to document

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,11 +9,37 @@ name: CI
 env:
   DISABLE_V8_COMPILE_CACHE: 1
 
+# Cancel outdated CI runs when new commits are pushed
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  CI:
-    name: CI_Node_${{ matrix.version }}
+  # Only needs to run on one version of node
+  # Doesn't need to build anything 
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+      - uses: actions/setup-node@3235b876344d2a9aa001b8d1453c930bba69e610 # v3.9.1
+        with:
+          cache: 'pnpm'
+          node-version: 20
+
+      - name: Install
+        run: pnpm install --frozen-lockfile
+
+      - name: Run linting
+        run: pnpm lint
+
+  # Test API Client packages
+  test-api-clients:
+    name: Test_ApiClients_Node_${{ matrix.version }}
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         version: [20.10.0, 22, 24]
     steps:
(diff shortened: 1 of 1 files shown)


.changeset/cyan-suits-film.md
```diff
@@ -0,0 +1,11 @@
+---
+'@shopify/shopify-api': major
+'@shopify/shopify-app-remix': major
+'@shopify/shopify-app-react-router': major
```
lizkenyon: I don't think we want a major for rr.

lizkenyon (commented):
I am thinking we should also be setting the `engines` field in the package.json.

Presently we are only setting it in the template, not these packages. But this has caused issues for devs using our packages without the template, when they are using an old node version.

.changeset/cyan-suits-film.md
byrichardpowell: Updated 👍

packages/apps/shopify-api/lib/auth/oauth/__tests__/create-session.test.ts
```diff
@@ -2,18 +2,10 @@ import {createSession} from '../create-session';
 import {Session, shopifyApi} from '../../..';
 import {testConfig} from '../../../__tests__/test-config';
 
-let shop: string;
-const STATIC_UUID = 'test-uuid';
+const SHOP = 'someshop.myshopify.io';
```
lizkenyon: nit: We do have exported conts that we use in other tests for stuff like this
https://github.com/Shopify/shopify-app-js/blob/0c085620f2460b3f2ce10dcadea747bb52c0211b/packages/apps/shopify-api/test-helpers/__tests__/setup-valid-request.test.ts#L11

packages/apps/shopify-api/package.json
```diff
@@ -5,6 +5,9 @@
   "types": "./dist/ts/lib/index.d.ts",
   "module": "./dist/esm/lib/index.mjs",
   "main": "./dist/cjs/lib/index.js",
+  "engines": {
+    "node": ">=20.0.0"
```
lizkenyon, line 9: I think we actually need to set it to >=20.10 because we need support for import attributes. 

https://github.com/Shopify/shopify-app-template-remix/blob/59f467dd6ca0462d47c26e89c6707cec695ec245/package.json#L23

Monosnap Import attributes - JavaScript | MDN 2025-09-10 14-08-16

lizkenyon: After a second look. Only shopify-app-remix requires 20.10. Because it is the only package that uses import attributes. 
https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx#L8

packages/apps/shopify-api/lib/auth/oauth/__tests__/create-session.test.ts
byrichardpowell: Fixed 👍

packages/apps/shopify-api/package.json
byrichardpowell: Good catch, fixed 👍

PR #2741, opened 2025-09-03
Node 20 remove jsonwebtoken

byrichardpowell:

We want to reduce dependencies to reduce maintenance

Swap `jsonwebtoken` for `jose`, so we only use one of the dependencies, rather than both.

## Type of change

## Checklist


packages/apps/shopify-app-react-router/src/server/authenticate/admin/__tests__/session-token-header-path.test.ts
```diff
@@ -1,5 +1,3 @@
-import {SESSION_COOKIE_NAME, Session} from '@shopify/shopify-api';
```
lizkenyon, line 1: 👍

PR #2745, opened 2025-09-05
Remove regenerator runtime

byrichardpowell:

We want to reduce maintenance by reducing dependencies

Remove the regenerator-runtime dependency since:

- This polyfills async await support, which has been supported since node 7
- This is only uses in our babel setup, which is only used in our test setup, so its not a breaking change.

When reviewing this, please confirm this only affects our test setup. Otherwise we should consider a breaking change

## Type of change

As long as this only affects out test setup, it is not a user facing change.  When reviewing this please confirm this only affects our test setup. Otherwise we should consider a breaking change

- ~~[ ] Patch: Bug (non-breaking change which fixes an issue)~~
- ~~[ ] Minor: New feature (non-breaking change which adds functionality)~~
- ~~[ ] Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)~~

## Checklist

N/A since changes only affect test infra

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
     "changeset": "changeset",
     "version-packages": "changeset version"
   },
+  "dependencies": {},
   "devDependencies": {
-    "@babel/plugin-transform-runtime": "^7.28.0",
     "@babel/preset-env": "^7.28.0",
     "@changesets/cli": "^2.29.5",
     "@cloudflare/workers-types": "^4.20250711.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -55,7 +54,6 @@
     "turbo": "^2.5.4",
     "typescript": "5.8.3"
   },
-  "dependencies": {},
```
lizkenyon, line 58: I _think_ we have ran into build issues when we removed this previously

lizkenyon: Something a part of our build process expects it to be there

lizkenyon: But seems to build fine so I guess we should be good. 🤷

byrichardpowell: Added this back as I've seen this issue as well.

PR #2766, opened 2025-09-08
[Spike] Migrate to rolldown

byrichardpowell:

Rolldown has fewer dependencies & faster builds.  Siking these changes primarilly for fewer deps, but the faster builds is also appealing.

Swap rollup for Rolldown.  This:

- Reduces build time from 160seconds to 50 seconds on my machine
- Allows us to drop 10 packages (reducing dependabot maintenance)

I don't think we should go ahead with this migration just yet because Rolldown is still in beta, meaning:

- Breaking changes seem likely, increasing maintenance
- There might be bugs affecting production builds
- In beta there are frequent releases, increasing maintnenace.
- minification support is not yet baked in (I achieved this via a plugin, but once this package leaves beta this plugin will likely not be needed)
- I YOLO changed the variable replacement we use for package version in API clients.

But, once rolldown exist beta this seems like a worthwhile migration.

## Type of change

## Checklist


byrichardpowell:
Closing for now.  Will revisit once rolldown exits beta.

PR #2781, opened 2025-09-12
[Major][shopify-api-js] Return Shopify API IDs as Strings

lizkenyon:

Fixes precision loss for Shopify REST API IDs approaching JavaScript's MAX_SAFE_INTEGER (2^53-1).

Shopify uses 64-bit integers for resource IDs, but JavaScript can only safely represent integers up to 2^53-1 (9,007,199,254,740,991). IDs beyond this limit lose precision when stored as numbers,
  potentially causing data corruption for merchants with large IDs. This change ensures all IDs are handled as strings to preserve their exact values.

  This PR migrates all REST API ID properties from number to string types to prevent precision loss. The implementation:

  - Updates the REST client's JSON parsing to preserve numeric precision for ID fields using the lossless-json library
  - Implements parseJsonWithLosslessNumbers method in the REST client (/packages/apps/shopify-api/lib/clients/admin/rest/client.ts) that:
    - Intercepts JSON responses before standard parsing would lose precision
    - Uses lossless-json to parse responses while maintaining exact numeric values
    - Automatically converts all ID fields (fields named 'id', ending with '_id', or '_ids') to strings
    - Converts other numeric fields back to regular JavaScript numbers for normal usage
  - Updates all REST resource type definitions across all API versions (2022-10 through 2024-10) to reflect string ID types
  - Maintains backward compatibility by accepting both numeric and string IDs as input parameters (auto-converted to strings internally)
  - Adds comprehensive tests for ID string conversion and lossless number parsing
  - Provides migration guide and documentation for the breaking change

  New dependency rationale: The lossless-json package (<4kB minified+gzipped, zero dependencies, MIT licensed) is used specifically in the REST client to parse JSON responses while preserving numeric
  precision. This battle-tested library handles edge cases in the JSON spec that would be complex to implement correctly in-house, ensuring reliable ID preservation without maintenance burden.

## Type of change

## Checklist

### DIFF
--- a/.changeset/dependabot-2747.md
+++ b/.changeset/dependabot-2747.md
@@ -1,0 +1,4 @@
+---
+---
+
+Updated `changesets/action` dependencies
--- a/.github/workflows/main-release.yml
+++ b/.github/workflows/main-release.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Create release Pull Request or publish to NPM
         id: changesets
-        uses: changesets/action@06245a4e0a36c064a573d4150030f5ec548e4fcc # v1.4.10
+        uses: changesets/action@e0145edc7d9d8679003495b11f87bd8ef63c0cba # v1.5.3
         with:
           title: Packages for release
           publish: pnpm release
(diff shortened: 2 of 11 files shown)


byrichardpowell (approved):
🚢  Great migration changeset

PR #2793, opened 2025-09-16
Default to Node 20 and pnpm 10.16.1 when inside this codebase

byrichardpowell:

Every time I `cd` into this directory it puts me on node 18, but the minimum version supported is Node 20.10.0 for some packages.

Update dev.yml config to use Node 20.10.0.  Whilst we are there, we may as well use a latest version of pnpm

## Type of change

N/A - won't affect packages

## Checklist

N/A - won't affect packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,5 +53,15 @@
     "jest": "^29.7.0",
     "jest-circus": "^29.7.0",
     "node-fetch": "^2.6.7"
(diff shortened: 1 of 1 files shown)


annfatah-bit:
git clone https://github.com/Shopify/shopify-app-examples.git

PR #2794, opened 2025-09-16
[AppProvider] - automatically handle the "shopify:navigate" event for Remix apps

jas7457:
This pull request adds support for automatically handling the "shopify:navigate" event in Remix apps using the `AppProvider` component. When this event is triggered (for example, by clicking a [s-link](https://shopify.dev/docs/api/app-home/polaris-web-components/actions/link)), the app will programmatically navigate to the specified URL using Remix's navigation system. Unit tests have also been added to verify this new behavior.

This aligns with the [code for the react-router template](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-app-react-router/src/react/components/AppProvider/AppProvider.tsx#L111-L126) and is needed as a complement to `s-link` to handle navigation, as pointed out in [our docs](https://shopify.dev/docs/api/app-home?accordionItem=getting-started-existing-remix-application).

Fixes https://github.com/shop/issues-polaris/issues/1183

## Type of change

## Checklist

### DIFF
--- a/.changeset/seven-monkeys-help.md
+++ b/.changeset/seven-monkeys-help.md
@@ -2,4 +2,4 @@
 '@shopify/shopify-app-remix': patch
 ---
 
-[AppProvider] - automatically handle the "shopify:navigate" event for Remix apps using Polaris Web Components
+[AppProvider] - automatically handle the "shopify:navigate" event for Remix apps using Polaris Web Components. If using Polars React, this has no effect.


.changeset/seven-monkeys-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/shopify-app-remix': patch
+---
+
+[AppProvider] - automatically handle the "shopify:navigate" event for Remix apps
```
byrichardpowell: Does this only affect Polaris Web Components? If so, I think it would be good to clarify that.

packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
```diff
@@ -111,6 +113,23 @@ export function AppProvider(props: AppProviderProps) {
...
+        navigate(href);
+      }
+    };
+
+    document.addEventListener('shopify:navigate', handleNavigate);
+
+    return () => {
+      document.removeEventListener('shopify:navigate', handleNavigate);
+    };
+  }, [navigate]);
```
byrichardpowell: I take it if you are not using Polaris Web components, then this event never gets fired, so this code never causes any harm?

byrichardpowell (commented):
Just two questions

.changeset/seven-monkeys-help.md
jas7457: As far as I know, yes. It gets triggered [here](https://github.com/shop/world/blob/ea604bcf300ed1363158c1d5ec330b0c7ed44ede/libraries/javascript/polaris/admin-ui-components/lib/componentBundleGenerationPlugin.js#L68-L73) and I don't see anywhere else triggering it.

I've updated the message to be more specific, good idea.

packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
jas7457: It shouldn't as it doesn't look like anything else triggers it. The [other instances](https://grokt.shopify.io/results?q=shopify%3Anavigate) just look like handlers.

packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
```diff
@@ -111,6 +113,23 @@ export function AppProvider(props: AppProviderProps) {
...
+
+  useEffect(() => {
+    const handleNavigate = (event: Event) => {
+      const href = (event.target as HTMLElement)?.getAttribute('href');
+      if (href) {
+        navigate(href);
+      }
+    };
+
+    document.addEventListener('shopify:navigate', handleNavigate);
```
developit:
```suggestion
    addEventListener('shopify:navigate', handleNavigate);
```

packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
```diff
@@ -111,6 +113,23 @@ export function AppProvider(props: AppProviderProps) {
...
+
+    return () => {
+      document.removeEventListener('shopify:navigate', handleNavigate);
```
developit:
```suggestion
      removeEventListener('shopify:navigate', handleNavigate);
```

packages/apps/shopify-app-remix/src/react/components/AppProvider/AppProvider.tsx
developit: correct

developit (approved):

.changeset/seven-monkeys-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/shopify-app-remix': patch
+---
+
+[AppProvider] - automatically handle the "shopify:navigate" event for Remix apps using Polaris Web Components
```
byrichardpowell: Can we make it clear that if you are using Polaris React still, this won't have any effect and is none breaking?

byrichardpowell (approved):
Approving so you don't need a second review, but please see my comment on the changeset

PR #2797, opened 2025-09-18
[WIP] Preload polaris webcomponents

byrichardpowell:
WIP whilst I confirm this helps performance.

We want apps to be fast

Preload the Polaris Web components URL

## Type of change

## Checklist

- ~~[ ] I have documented new APIs/updated the documentation for modified APIs (for public APIs)~~

### DIFF
--- a/packages/apps/shopify-app-react-router/src/server/__test-helpers/expect-document-request-headers.ts
+++ b/packages/apps/shopify-app-react-router/src/server/__test-helpers/expect-document-request-headers.ts
@@ -1,4 +1,4 @@
-import {APP_BRIDGE_URL} from '../authenticate/const';
+import {APP_BRIDGE_URL, POLARIS_URL} from '../authenticate/const';
 
 import {TEST_SHOP} from './const';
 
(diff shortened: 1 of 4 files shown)


packages/apps/shopify-app-react-router/src/server/__test-helpers/expect-document-request-headers.ts
```diff
@@ -15,7 +15,7 @@ export function expectDocumentRequestHeaders(
       )} https://admin.shopify.com https://*.spin.dev https://admin.myshopify.io https://admin.shop.dev;`,
     );
     expect(headers.get('Link')).toEqual(
```
JoviDeCroock, line 17: We probably also want to add 

```
    <link
      rel="preconnect"
      href="https://cdn.shopify.com/"
    />
```

byrichardpowell: Yeah, good idea.  Added 👍

byrichardpowell:
> The best and easiest way to measure the difference is to goto the performance tab and look at the assets starting to load in line with the HTML parse step. With a small HTML document the impact of this is ofcourse smaller (for instance an SPA) as opposed to an SSR or streamed application.

Thanks 👍

lizkenyon (approved):

byrichardpowell:
Added the pre-connect directive.

I didn't see any real performance gain, but it did no harm.  It's likely an app would only see a performance gain if:

- They have a large head element with lots of resources to download
- They are doing a stream response, with Polaris being inside a delayed chunk
- The user has a very slow connection

PR #2929, opened 2025-11-13
[shopify-app-react-router][patch] Fix authentication loop for apps with React Router basename

lizkenyon:
Apps using React Router basename (e.g., /base-path) experienced infinite authentication loops because pathname comparisons used exact equality (===) instead of suffix matching, causing routes like /base-path/auth/session-token to fail matching /auth/session-token

This was a design oversight from the initial implementation that assumed apps would be deployed at root path - affects any Shopify app using React Router with basename configuration

###Before
image (6)

### After
https://github.com/user-attachments/assets/68b1ccd3-8ee0-4335-89aa-c3feaf221da0
https://github.com/user-attachments/assets/520457d4-fb0a-45ec-806d-2fd45cb6ca58

Fixes #0000 

## Type of change

## Checklist

### DIFF
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/admin/__tests__/patch-session-token-path.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/admin/__tests__/patch-session-token-path.test.ts
@@ -75,7 +75,6 @@ describe('authorize.admin path session token path', () => {
 
     // THEN
     // Should render the bounce page (AppBridge script) even though pathname doesn't exactly match config.auth.patchSessionTokenPath
-    // Currently FAILS because url.pathname="/base-path/auth/session-token" doesn't match config.auth.patchSessionTokenPath="/auth/session-token"
     expect(response.status).toBe(200);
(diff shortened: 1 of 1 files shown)


packages/apps/shopify-app-react-router/src/server/authenticate/admin/helpers/validate-shop-and-host-params.ts
```diff
@@ -33,7 +33,7 @@ function redirectToLoginPath(request: Request, params: BasicParams): never {
   const {config, logger} = params;
 
   const {pathname} = new URL(request.url);
-  if (pathname === config.auth.loginPath) {
+  if (pathname.endsWith(config.auth.loginPath)) {
```
lizkenyon, line 36: This is probably the simplest fix. 

We could add a configuration for basename in ShopifyApp.

We can't rely on the configuration for auth because
```
  - The basename is added by React Router (client-side)
  - But patchSessionTokenPath is configured server-side
  - The comparison logic used === (exact match) instead of .endsWith() (suffix match)
```

PR #2938, opened 2025-11-17
feat: Add support for expiring offline access tokens

fwaadahmad1:
Resolves https://github.com/shop/issues-app-access/issues/836

## 🎯 Store Refresh Tokens in Session Model

This PR adds support for storing OAuth refresh tokens and their expiration dates as part of the session object, enabling token refresh for long-lived sessions.

#### Session Model Extensions

The `Session` class now includes two new optional fields:

- `refreshToken`: The OAuth refresh token returned from Shopify
- `refreshTokenExpires`: The expiration date of the refresh token

These fields are populated during OAuth flows (both authorization code and token exchange) when Shopify returns refresh token information.

#### Example Session Object

```ts
{
  id: 'offline_session_id',
  shop: 'my-shop.myshopify.com',
  state: 'session-state',
  isOnline: false,
  scope: 'read_products,write_orders',
  accessToken: 'shpat_abc123',
  expires: 2025-11-17T18:00:00.000Z,
  refreshToken: 'shprt_def456',           // ✨ New field
  refreshTokenExpires: 2025-12-17T18:00:00.000Z,  // ✨ New field
}
```

### Key Benefits

1. **Automatic Token Refresh**: Apps can now refresh expired access tokens without requiring user re-authentication
2. **Backward Compatible**: Existing sessions continue to work without modification (fields are optional)

#### Core Libraries (`@shopify/shopify-api`)

- Extended `Session` class with `refreshToken` and `refreshTokenExpires` properties
- Updated OAuth flows to capture and store refresh token information
- Added comprehensive test coverage for refresh token handling

#### React Router (`@shopify/shopify-app-react-router`)

- Updated all authentication strategies to preserve refresh token data
- Modified session creation/loading logic across admin, flow, fulfillment-service, webhooks, and app proxy authenticators
- Removed deprecated helper functions and consolidated session handling

#### Session Storage (`@shopify/shopify-app-session-storage-*`)

- Updated test utilities to support refresh token validation
- Schema changes are handled at the storage adapter level (Prisma, MySQL, PostgreSQL, etc.)

### Usage Example

```ts
const session = await sessionStorage.loadSession(sessionId);

// Access refresh token information
console.log('Refresh token:', session.refreshToken);
console.log('Expires:', session.refreshTokenExpires);

// Check if refresh token is available
if (session.refreshToken) {
  // Use refresh token to get a new access token
  // (automatic refresh logic can be implemented)
}
```

### Migration Notes

For apps using session storage adapters like Prisma:

[...]

PR #2940, opened 2025-11-19
feat(api) Add support for expiring offline access tokens

fwaadahmad1:

Resolves https://github.com/shop/issues-app-access/issues/839 

This PR adds full support for OAuth refresh tokens in the `@shopify/shopify-api` package, enabling apps to use expiring access tokens and rotate them securely.

- **Session Model Updates**: Added `refreshToken` and `refreshTokenExpires` properties to the `Session` class with proper serialization support
- **New Auth Methods**:
  - `shopify.auth.refreshToken`: Exchange a refresh token for a new access token
  - `shopify.auth.migrateToExpiringToken`: Migrate existing non-expiring offline tokens to expiring tokens
- **Token Exchange Updates**: Added `expiring` parameter to request expiring access tokens

## Type of change

## Checklist

### DIFF
--- a/.changeset/red-mails-cheat.md
+++ b/.changeset/red-mails-cheat.md
@@ -1,0 +1,34 @@
+---
+'@shopify/shopify-api': minor
+---
+
+This change introduces full support for OAuth refresh tokens in the `@shopify/shopify-api` package, enabling apps to use expiring access tokens and rotate them securely.
+
+- **Session Model Updates**: The `Session` class now includes properties to store refresh token information:
+  - `refreshToken`: The refresh token string.
+  - `refreshTokenExpires`: The date when the refresh token expires.
+  - Serialization logic has been updated to persist these fields.
+
+- **New Auth Methods**:
+  - `shopify.auth.refreshToken`: A new method to exchange a refresh token for a new access token.
+  - `shopify.auth.migrateToExpiringToken`: A helper method to migrate existing non-expiring offline tokens to expiring tokens.
+
+- **Token Exchange Updates**:
+  - The `tokenExchange` method now accepts an `expiring` parameter to request expiring access tokens.
+
+```ts
+// Migrating a non-expiring token
+const {session} = await shopify.auth.migrateToExpiringToken({
+  shop: 'my-shop.myshopify.com',
+  nonExpiringOfflineAccessToken: 'shpo_...',
+});
+
+// Refreshing an expired token
+const {session: newSession} = await shopify.auth.refreshToken({
+  shop: session.shop,
+  refreshToken: session.refreshToken,
+});
+
+console.log(newSession.accessToken); // New access token
+console.log(newSession.refreshToken); // New refresh token
+```
--- a/packages/apps/shopify-api/lib/auth/oauth/__tests__/migrate-to-expiring-tokens.test.ts
+++ b/packages/apps/shopify-api/lib/auth/oauth/__tests__/migrate-to-expiring-tokens.test.ts
@@ -1,0 +1,124 @@
+import * as ShopifyErrors from '../../../error';
+import {DataType, shopifyApi} from '../../..';
+import {testConfig} from '../../../__tests__/test-config';
+import {queueMockResponse} from '../../../__tests__/test-helper';
+import {RequestedTokenType} from '../token-exchange';
+
+let shop: string;
+let nonExpiringOfflineAccessToken: string;
+
+beforeEach(() => {
+  shop = 'someshop.myshopify.io';
+  nonExpiringOfflineAccessToken = 'some-access-token';
+});
+
+describe('migrateToExpiringToken', () => {
+  describe('with valid parameters', () => {
+    test('returns a session with expiring access token', async () => {
+      const shopify = shopifyApi(testConfig());
+      const scopes = shopify.config.scopes
+        ? shopify.config.scopes.toString()
+        : '';
+
+      const successResponse = {
+        access_token: 'new_access_token',
+        scope: scopes,
+        expires_in: 525600,
+        refresh_token: 'some_refresh_token',
+        refresh_token_expires_in: 2592000,
+      };
+
+      const expectedExpiration = new Date(
+        Date.now() + successResponse.expires_in * 1000,
+      ).getTime();
+      const expectedRefreshTokenExpiration = new Date(
+        Date.now() + successResponse.refresh_token_expires_in * 1000,
+      ).getTime();
+
+      queueMockResponse(JSON.stringify(successResponse));
+
+      const response = await shopify.auth.migrateToExpiringToken({
+        shop,
+        nonExpiringOfflineAccessToken,
+      });
+
+      expect({
+        method: 'POST',
+        domain: shop,
+        path: '/admin/oauth/access_token',
+        headers: {
+          Accept: DataType.JSON,
+        },
+        data: {
+          client_id: shopify.config.apiKey,
+          client_secret: shopify.config.apiSecretKey,
+          grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
+          subject_token: nonExpiringOfflineAccessToken,
+          subject_token_type: RequestedTokenType.OfflineAccessToken,
+          requested_token_type: RequestedTokenType.OfflineAccessToken,
+          expiring: '1',
+        },
+      }).toMatchMadeHttpRequest();
+
+      expect(response.session).toEqual(
+        expect.objectContaining({
+          accessToken: successResponse.access_token,
+          scope: successResponse.scope,
+          refreshToken: successResponse.refresh_token,
+        }),
+      );
+
+      expect(response.session?.expires?.getTime()).toBeWithinSecondsOf(
+        expectedExpiration,
+        1,
+      );
+
+      expect(
+        response.session?.refreshTokenExpires?.getTime(),
+      ).toBeWithinSecondsOf(expectedRefreshTokenExpiration, 1);
+    });
+
+    test('throws error when response is bad request', async () => {
+      const shopify = shopifyApi(testConfig());
+
+      const errorResponse = {
+        error: 'invalid_request',
+        error_description: 'Invalid subject token',
+      };
+
+      queueMockResponse(JSON.stringify(errorResponse), {
+        statusCode: 400,
+        statusText: 'Bad request',
+      });
+
+      const exchangePromise = shopify.auth.migrateToExpiringToken({
+        shop,
+        nonExpiringOfflineAccessToken,
+      });
+
+      exchangePromise.catch((error) => {
+        expect(error).toHaveProperty('response.body', errorResponse);
+      });
+      await expect(exchangePromise).rejects.toThrow(
+        ShopifyErrors.HttpResponseError,
+      );
+    });
+
+    describe('with invalid shop', () => {
+      beforeEach(() => {
+        shop = 'someshop.myspopify.io';
+      });
+
+      test('throws InvalidShopError', async () => {
+        const shopify = shopifyApi(testConfig());
+
+        await expect(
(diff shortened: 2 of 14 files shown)


fwaadahmad1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2940?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2940** <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2940?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2940?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2938** <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2938?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3 other dependent PRs ([#2941](https://github.com/Shopify/shopify-app-js/pull/2941) <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2941?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2942](https://github.com/Shopify/shopify-app-js/pull/2942) <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2942?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2946](https://github.com/Shopify/shopify-app-js/pull/2946) <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2946?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

.changeset/red-mails-cheat.md
```diff
@@ -0,0 +1,34 @@
...
+
+- **Token Exchange Updates**:
+  - The `tokenExchange` method now accepts an `expiring` parameter to request expiring access tokens.
+
+```ts
+// Migrating a non-expiring token
+const {session} = await shopify.auth.migrateToExpiringToken({
+  shop: 'my-shop.myshopify.com',
+  nonExpiringOfflineAccessToken: 'shpo_...',
+});
```
byrichardpowell, line 24: A few questions here:

1. How do we expect users to perform this migration? A background Job? 
2. However we expect users to perform this migration, are we providing a comprehensive example of how to migrate all sessions?
3. What other options did we consider?  One of the tradeoffs with adding this API to tis package is we have to manage long term maintenance and eventually a breaking change where we remove it.  Did we consider simply providing a GIST? Did we consider providing a seperate package?
4. If we _do_ want to provide this functionality in the package, is it actually a property of the Session object instead?  E.g: Load each "Session", then call `session.migrate...`

packages/apps/shopify-api/lib/auth/oauth/__tests__/migrate-to-expiring-tokens.test.ts
```diff
@@ -0,0 +1,124 @@
...
+let shop: string;
+let nonExpiringOfflineAccessToken: string;
+
+beforeEach(() => {
+  shop = 'someshop.myshopify.io';
+  nonExpiringOfflineAccessToken = 'some-access-token';
+});
+
+describe('migrateToExpiringToken', () => {
+  describe('with valid parameters', () => {
```
byrichardpowell, line 16: Do we need t consider what happens when:

1. Params are invalid?
2. The user attempts to migrate a token that has already been migrated?

packages/apps/shopify-api/lib/auth/oauth/__tests__/refresh-token.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      const expectedRefreshTokenExpiration = new Date(
+        Date.now() + successResponse.refresh_token_expires_in * 1000,
+      ).getTime();
+
+      queueMockResponse(JSON.stringify(successResponse));
+
+      const result = await shopify.auth.refreshToken({
+        shop,
+        refreshToken: refreshTokenValue,
+      });
```
byrichardpowell, line 40: What happens if you pass invalid params?

This makes me wonder if its beneficial to add this functionality to the `Session` class since we can use the logic there that already prevents you from loading an invalid session.

packages/apps/shopify-api/lib/auth/oauth/create-session.ts
```diff
@@ -47,10 +47,14 @@ export function createSession({
...
   const getOfflineSessionProperties = (responseBody: OfflineAccessResponse) => {
-    const {expires_in} = responseBody;
+    const {expires_in, refresh_token, refresh_token_expires_in} = responseBody;
     return {
       id: getOfflineId(config)(shop),
       ...(expires_in && {expires: getSessionExpiration(expires_in)}),
+      ...(refresh_token && {refreshToken: refresh_token}),
+      ...(refresh_token_expires_in && {
+        refreshTokenExpires: getSessionExpiration(refresh_token_expires_in),
+      }),
```
byrichardpowell: I assume it would be an error state to have a `refresh_token` without a `refresh_token_expires_in` ? If so, can this be combined?

```
...(refresh_token && refresh_token_expires_in  && {refreshToken: refresh_token, refreshTokenExpires: getSessionExpiration(refresh_token_expires_in)}),
```

packages/apps/shopify-api/lib/auth/oauth/token-exchange.ts
```diff
@@ -43,6 +45,7 @@ export function tokenExchange(config: ConfigInterface): TokenExchange {
       subject_token: sessionToken,
       subject_token_type: IdTokenType,
       requested_token_type: requestedTokenType,
+      expiring: expiring ? '1' : '0',
```
byrichardpowell, line 48: Why convert truthy to 1 or 0 ?

packages/apps/shopify-api/lib/session/types.ts
```diff
@@ -30,6 +30,14 @@ export interface SessionParams {
...
    */
   accessToken?: string;
+  /**
+   * The refresh token for the session.
+   */
+  refreshToken?: string;
+  /**
+   * The date the refresh token expires.
+   */
+  refreshTokenExpires?: Date;
```
byrichardpowell, line 40: Are these optional to support migration from without refresh to with refresh? Make sense.  

If so, please do we have a long term plan to make these mandatory with a breaking change in future?  I'd like to avoid a situation where these properties remain optional even though we enforce them at some future point

packages/apps/shopify-api/lib/auth/oauth/__tests__/migrate-to-expiring-tokens.test.ts
fwaadahmad1: \>> Params are invalid?  
In the scenario where either the shop, or the nonExpiringOfflineAccessToken is invalid, the API would get an appropriate response from the server  
invalid shop_url:  


invalid access_token:



  
\>> The user attempts to migrate a token that has already been migrated?  
In the scenario too, the API would get an appropriate response from the server

packages/apps/shopify-api/lib/auth/oauth/__tests__/refresh-token.test.ts
fwaadahmad1: Similar to the previous comment when you pass invalid params an approprite response is sent from the backend

fwaadahmad1: From what I can tell, the **`Session`** class in the **`shopify-api`** package is essentially a class-based representation of a session record in your database, enhanced with some helper methods that encapsulate common logic and add more semantic meaning to that record. While adding a method that makes an API call to keep the session alive does make sense conceptually, I think it would break the current architectural design by giving the **`Session`** class multiple responsibilities.

packages/apps/shopify-api/lib/auth/oauth/create-session.ts
fwaadahmad1: I agree with you, it would be an error state to not have both at the same time, I will combine this

packages/apps/shopify-api/lib/auth/oauth/token-exchange.ts
fwaadahmad1: We need to convert this to either '1' or '0' because no other values are accepted by core.  
Reference core [PR](https://app.graphite.com/github/pr/shop/world/209967/Can-mint-expiring-offline-tokens-with-refresh-tokens-during-token-exchange)

packages/apps/shopify-api/lib/session/types.ts
fwaadahmad1: These are optional because the session object obtained from the `client-credentials`​flow does not have a refresh token. Since there exists a scenario (client-credentials) where refresh_tokens are not present; The type for these will remain as optional

.changeset/red-mails-cheat.md
fwaadahmad1: \>> How do we expect users to perform this migration? A background Job?  
Yes, we Expect users to create a background job for this  
  
\>> However we expect users to perform this migration, are we providing a comprehensive example of how to migrate all sessions?  
Yes, there is a [PR](https://app.graphite.com/github/pr/Shopify/shopify-app-js/2942) up-stack for docs that includes docs for all the packages in which we have added a step by step guide for the users to follow along with an example background job, which should work out of the box if users are following shopify's database schema or else they can modify the example to their needs.  
  
\>> What other options did we consider?  One of the tradeoffs with adding this API to tis package is we have to manage long term maintenance and eventually a breaking change where we remove it.  Did we consider simply providing a GIST? Did we consider providing a seperate package?  
We considered an providing a example script which the developers can use/modify for migration; but in the end to improve the Dx, we chose to go this route accepting the trade-off for a breaking change where we remove it.  
  
\>> If we _do_ want to provide this functionality in the package, is it actually a property of the Session object instead?  E.g: Load each "Session", then call `session.migrate...`  
We can do this but we wanted this functionality to be more generic, so apps with a database schema different than what we provide can also easily use this functionality to migrate their tokens

byrichardpowell: Makes sense, thanks for the detailed context.

packages/apps/shopify-api/lib/auth/oauth/create-session.ts
byrichardpowell: Thanks 👍

packages/apps/shopify-api/lib/auth/oauth/token-exchange.ts
byrichardpowell: Thanks for the context 👍

packages/apps/shopify-api/lib/session/types.ts
byrichardpowell: Great context, thank you 👍

packages/apps/shopify-api/lib/auth/oauth/__tests__/refresh-token.test.ts
byrichardpowell: Ok, thanks for the context. 

I'm happy with this direction 👍   One addition that might be nice to add is to validate the `refreshToken` string is the format we expect.  I can imagine users passing the wrong string to this function and  I think an explicit wrror message for that scenario might be helpful.

Optional though, doesn't need to hold up this PR.

PR #2941, opened 2025-11-19
feat(react-router) Add support for expiring offline access tokens

fwaadahmad1:

Fixes https://github.com/shop/issues-app-access/issues/792

This PR adds support for expiring offline access tokens, allowing apps to refresh expired tokens automatically.

- Adds a new `expiringOfflineAccessTokens` configuration option to enable token refresh functionality
- Implements token refresh logic for expired offline sessions across all authentication methods
- Adds a threshold check to refresh tokens that are about to expire within 1000ms
- Ensures tokens are only refreshed when the feature flag is enabled and the app is not using ShopifyAdmin distribution
- Adds comprehensive test coverage for the new functionality

## Type of change

## Checklist

### DIFF
--- a/.changeset/metal-towns-flow.md
+++ b/.changeset/metal-towns-flow.md
@@ -1,0 +1,48 @@
+---
+'@shopify/shopify-app-react-router': minor
+---
+
+We are introducing support for expiring offline access tokens. This feature improves security by limiting the lifespan of offline access tokens and automatically refreshing them using refresh tokens.
+
+- **New Future Flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`. When enabled, the library will start using expiring offline tokens and automatically check if it is expired or nearing expiration. If expired/expiring, it attempts to refresh the access token using the stored refresh token. Defaults to `false` for backward compatibility.
+
+- **Automatic Token Refresh**: Integrated token refresh logic into authentication flows (`flow`, `fulfillmentService`, `appProxy`, `webhooks`) and unauthenticated contexts (`admin`, `storefront`). When a session is loaded and found to be expired (or expiring within 5 minutes), and the feature is enabled, the library transparently refreshes the token and persists the new session data. This behavior applies to both offline and online tokens.
+
+To enable expiring offline access tokens in your app, you must ensure your session storage can persist refresh tokens. For now, this will only work if you are using the Prisma Session Storage package. We're starting with Prisma since this is what the majority of our developers use. If you're using a different session storage adapter and would like support for expiring offline tokens, we'd love to hear from you! If you are using Prisma, follow these steps:
+
+1. Update your `@shopify/shopify-api` and `@shopify/shopify-app-session-storage-prisma` packages to the latest version.
+
+2. Update your Prisma schema to include the `refreshToken` and `refreshTokenExpires` fields in the `Session` model:
+
+```prisma
+model Session {
+  // ...
+  refreshToken        String?
+  refreshTokenExpires DateTime?
+}
+```
+
+3. Run a migration to update your database:
+
+```sh
+npx prisma migrate dev
+```
+
+4. Update the generated types to include the new fields:
+
+```sh
+npx prisma generate
+```
+
+5. Enable the future flag in your app configuration:
+
+```ts
+const shopify = shopifyApp({
+  // ... other config
+  future: {
+    expiringOfflineAccessTokens: true,
+  },
+});
+```
+
+When enabled, calls to `shopify.authenticate.admin`, `shopify.authenticate.flow`, etc., will automatically handle token refreshing for offline sessions.
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/flow/__tests__/authenticate.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/flow/__tests__/authenticate.test.ts
@@ -1,14 +1,17 @@
 import {SessionStorage} from '@shopify/shopify-app-session-storage';
 import {MemorySessionStorage} from '@shopify/shopify-app-session-storage-memory';
+import {Session} from '@shopify/shopify-api';
 
 import {shopifyApp} from '../../..';
 import {
   expectAdminApiClient,
+  expectTokenRefresh,
   getHmac,
   getThrownResponse,
   setUpValidSession,
   testConfig,
 } from '../../../__test-helpers';
+import {TestOverridesArg} from '../../../test-helpers/test-config';
 
 const FLOW_URL = 'https://example.myapp.io/authenticate/flow';
 
@@ -118,10 +121,44 @@ describe('authenticating flow requests', () => {
       return {admin, expectedSession, actualSession};
     });
   });
+
+  describe('token refresh for expired offline sessions', () => {
+    expectTokenRefresh(
+      async (
+        sessionStorage: SessionStorage,
+        session: Session,
+        configOverrides: TestOverridesArg,
+      ) => {
+        const shopify = shopifyApp(
+          testConfig({
+            sessionStorage,
+            ...configOverrides,
+          }) as any,
+        );
+
+        const body = {shopify_domain: session.shop};
+        const bodyString = JSON.stringify(body);
+
+        const request = new Request(FLOW_URL, {
+          body: bodyString,
+          method: 'POST',
+          headers: {
+            'X-Shopify-Hmac-Sha256': getHmac(bodyString),
+          },
+        });
+
+        const {session: actualSession} =
+          await shopify.authenticate.flow(request);
+        return actualSession;
+      },
+    );
+  });
 });
 
 async function getValidRequest(sessionStorage: SessionStorage) {
-  const session = await setUpValidSession(sessionStorage, {isOnline: false});
+  const session = await setUpValidSession(sessionStorage, {
+    isOnline: false,
+  });
 
   const body = {shopify_domain: session.shop};
   const bodyString = JSON.stringify(body);
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/webhooks/authenticate.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/webhooks/authenticate.ts
@@ -3,7 +3,7 @@ import {WebhookValidationErrorReason} from '@shopify/shopify-api';
 import type {BasicParams} from '../../types';
 import {adminClientFactory} from '../../clients';
 import {handleClientErrorFactory} from '../admin/helpers';
-import {createOrLoadOfflineSession} from '../helpers';
+import {ensureValidOfflineSession} from '../../helpers';
 
 import type {
   AuthenticateWebhook,
@@ -49,7 +49,7 @@ export function authenticateWebhookFactory<Topics extends string>(
         throw new Response(undefined, {status: 400, statusText: 'Bad Request'});
       }
     }
-    const session = await createOrLoadOfflineSession(check.domain, params);
+    const session = await ensureValidOfflineSession(params, check.domain);
     const webhookContext: WebhookContextWithoutSession<Topics> = {
       apiVersion: check.apiVersion,
       shop: check.domain,
--- a/packages/apps/shopify-app-react-router/src/server/future/flags.ts
+++ b/packages/apps/shopify-app-react-router/src/server/future/flags.ts
@@ -4,11 +4,21 @@ import type {AppConfig} from '../config-types';
 
 // When adding new flags, you should also add them to the `TEST_FUTURE_FLAGS` object in `test-config.ts` to ensure that
 // it doesn't cause regressions.
-export interface FutureFlags {}
+export interface FutureFlags {
+  /**
+   * When enabled, the app will start using expiring offline access tokens and automatically refresh them when they are close to expiring.
+   *
+   * @default false
+   */
+  expiringOfflineAccessTokens?: boolean;
+}
 
 // When adding new flags, use this format:
 // apiFutureFlag: Future extends FutureFlags ? Future['reactRouterFutureFlag'] : false;
-export interface ApiFutureFlags<_Future extends FutureFlagOptions> {
+export interface ApiFutureFlags<Future extends FutureFlagOptions> {
+  expiringOfflineAccessTokens: Future extends FutureFlags
+    ? Future['expiringOfflineAccessTokens']
+    : false;
   unstable_managedPricingSupport: true;
 }
 
--- a/packages/apps/shopify-app-react-router/src/server/helpers/__tests__/ensure-offline-token-is-not-expired.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/helpers/__tests__/ensure-offline-token-is-not-expired.test.ts
@@ -1,0 +1,32 @@
+import {SessionStorage} from '@shopify/shopify-app-session-storage';
+
+import {testConfig} from '../../__test-helpers';
+import {expectTokenRefresh} from '../../__test-helpers/expect-token-refresh';
+import {AppConfigArg} from '../../config-types';
+import {deriveApi} from '../../shopify-app';
+import {BasicParams} from '../../types';
+import {
+  ensureOfflineTokenIsNotExpired,
+  WITHIN_MILLISECONDS_OF_EXPIRY,
+} from '../ensure-offline-token-is-not-expired';
+import {FutureFlagOptions} from '../../future/flags';
+
+describe('ensureOfflineTokenIsNotExpired', () => {
+  it('uses 5 minutes for WITHIN_MILLISECONDS_OF_EXPIRY', async () => {
+    expect(WITHIN_MILLISECONDS_OF_EXPIRY).toBe(5 * 60 * 1000);
+  });
+  expectTokenRefresh(async (sessionStorage, session, configOverrides) => {
+    const config = testConfig({
+      sessionStorage,
+      ...configOverrides,
+    }) as AppConfigArg<SessionStorage, FutureFlagOptions>;
+    const api = deriveApi(config);
+    const params = {
+      api,
+      config,
+      logger: api.logger,
+    } as unknown as BasicParams;
+
+    return ensureOfflineTokenIsNotExpired(session, params, session.shop);
+  });
+});
--- a/packages/apps/shopify-app-react-router/src/server/helpers/__tests__/refresh-token.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/helpers/__tests__/refresh-token.test.ts
@@ -1,0 +1,168 @@
+import {HttpResponseError, InvalidJwtError} from '@shopify/shopify-api';
+import {setUpValidSession as setUpValidSessionImport} from '@shopify/shopify-api/test-helpers';
+import {MemorySessionStorage} from '@shopify/shopify-app-session-storage-memory';
+
+import {testConfig, TEST_SHOP} from '../../__test-helpers';
+import {deriveApi} from '../../shopify-app';
+import {BasicParams} from '../../types';
+import refreshToken from '../refresh-token';
+
+describe('refreshToken', () => {
+  let sessionStorage: MemorySessionStorage;
+  let params: BasicParams;
+  let mockRefreshToken: jest.Mock;
+
+  beforeEach(() => {
+    sessionStorage = new MemorySessionStorage();
+    const config = testConfig({sessionStorage});
+
+    mockRefreshToken = jest.fn();
+
+    // Create api using deriveApi
+    const api = deriveApi(config);
+
+    // Create params with mocked api.auth.refreshToken
+    params = {
+      api: {
+        ...api,
+        auth: {
+          ...api.auth,
+          refreshToken: mockRefreshToken,
+        },
+      },
+      config,
+      logger: api.logger,
+    } as unknown as BasicParams;
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('refreshToken', () => {
+    it('successfully refreshes a token and returns a session', async () => {
+      // GIVEN
+      const refreshTokenValue = 'test-refresh-token';
+      const newSession = setUpValidSessionImport({
+        shop: TEST_SHOP,
+        expires: new Date(Date.now() + 86400000),
+        refreshToken: 'new-refresh-token',
+        isOnline: false,
+        accessToken: 'new-access-token',
+      });
+
+      mockRefreshToken.mockResolvedValue({session: newSession});
+
+      // WHEN
+      const result = await refreshToken(params, TEST_SHOP, refreshTokenValue);
+
+      // THEN
+      expect(result).toBeDefined();
+      expect(result.accessToken).toBe('new-access-token');
+      expect(result.refreshToken).toBe('new-refresh-token');
+      expect(mockRefreshToken).toHaveBeenCalledWith({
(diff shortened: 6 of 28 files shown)


fwaadahmad1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2941?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2941** <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2941?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2941?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2938** <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2938?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3 other dependent PRs ([#2940](https://github.com/Shopify/shopify-app-js/pull/2940) <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2940?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2942](https://github.com/Shopify/shopify-app-js/pull/2942) <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2942?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2946](https://github.com/Shopify/shopify-app-js/pull/2946) <a href="https://app.graphite.com/github/pr/Shopify/shopify-app-js/2946?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,34 @@
...
+
+We are introducing support for expiring offline access tokens. This feature improves security by limiting the lifespan of offline access tokens and automatically refreshing them using refresh tokens.
+
+- **New Future Flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`.
+  - When enabled, the library will start using expiring offline tokens and automatically check if it is expired or nearing expiration.
+  - If expired/expiring, it attempts to refresh the access token using the stored refresh token.
+  - Defaults to `false` for backward compatibility.
+
+- **Automatic Token Refresh**:
+  - Integrated token refresh logic into authentication flows (`flow`, `fulfillmentService`, `appProxy`, `webhooks`) and unauthenticated contexts (`admin`, `storefront`).
```
byrichardpowell: Do we also need to integrate it into checkout and customer account?

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,34 @@
...
+  - When a session is loaded and found to be expired (or expiring within 1 second), and the feature is enabled, the library transparently refreshes the token and persists the new session data.
+
+- **Internal Helpers**:
+  - Introduced `getOfflineSession`: A unified helper to load an offline session and ensure it's valid (refreshing if necessary).
+  - Added `ensureOfflineTokenIsNotExpired`: Logic to check expiration and trigger refresh.
+  - Added `refreshToken` helper: Wraps the API client's refresh token functionality.
```
byrichardpowell: If these are internal helpers do they need to be in the changelog? End users don't care?

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,34 @@
+---
+'@shopify/shopify-app-react-router': minor
+---
+
+We are introducing support for expiring offline access tokens. This feature improves security by limiting the lifespan of offline access tokens and automatically refreshing them using refresh tokens.
+
+- **New Future Flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`.
+  - When enabled, the library will start using expiring offline tokens and automatically check if it is expired or nearing expiration.
+  - If expired/expiring, it attempts to refresh the access token using the stored refresh token.
+  - Defaults to `false` for backward compatibility.
```
byrichardpowell: Not sure the Nested bullet points is the correct approach in terms of grammer?  Use headings with paragraphs? Or just paragraphs?

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,34 @@
+---
+'@shopify/shopify-app-react-router': minor
+---
+
+We are introducing support for expiring offline access tokens. This feature improves security by limiting the lifespan of offline access tokens and automatically refreshing them using refresh tokens.
+
+- **New Future Flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`.
+  - When enabled, the library will start using expiring offline tokens and automatically check if it is expired or nearing expiration.
```
byrichardpowell: But, if I understand correctly, this will only work if you are using Prisma.  If you are using any other Session Storage adapter this feature is not supported?  I think we should either fix that limitation or call it out explicitly.

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,34 @@
...
+To enable expiring offline access tokens in your app:
+
+```ts
+const shopify = shopifyApp({
+  // ... other config
+  future: {
+    expiringOfflineAccessTokens: true,
+  },
+});
+```
```
byrichardpowell: I don't think this is _all_ the user needs to do? I think they also need to update their prisma session storage version, update the schema, then run a migration.

If I'm right, please can we include all the steps?

packages/apps/shopify-app-react-router/src/server/authenticate/flow/__tests__/authenticate.test.ts
```diff
@@ -118,10 +126,212 @@ describe('authenticating flow requests', () => {
       return {admin, expectedSession, actualSession};
     });
   });
+
+  describe('token refresh for expired offline sessions', () => {
```
byrichardpowell, line 125: It looks like there is a lot of repetition with the tests for this behaviour, since every authenticate function with this behaviour is tested with separate tests.  I worry that this invites drift where different surfaces implement refresh in different ways.

We have a pattern elsewhere where we provide an `expectSomething` test helper, which can be reused by each surface that uses it.  

Here is an `expectAdminApiClient` example:

```
expectAdminApiClient(async () => {
    const {
      admin,
      expectedSession,
      session: actualSession,
    } = await setUpEmbeddedFlow();

    return {admin, expectedSession, actualSession};
  });
```

packages/apps/shopify-app-react-router/src/server/authenticate/webhooks/authenticate.ts
```diff
@@ -49,7 +49,7 @@ export function authenticateWebhookFactory<Topics extends string>(
         throw new Response(undefined, {status: 400, statusText: 'Bad Request'});
       }
     }
-    const session = await createOrLoadOfflineSession(check.domain, params);
+    const session = await getOfflineSession(params, check.domain);
```
byrichardpowell: `createOrLoadOfflineSession` and `getOfflineSession` have their argument orders reversed.  Please can we be consistant?

packages/apps/shopify-app-react-router/src/server/future/flags.ts
```diff
@@ -27,15 +37,16 @@ export type FeatureEnabled<
...
-  // }
+  const logFlag = (flag: string, message: string) =>
+    logger.info(`Future flag ${flag} is disabled.\n\n  ${message}\n`);
+
+  if (!config.future.expiringOfflineAccessTokens) {
+    logFlag(
+      'expiringOfflineAccessTokens',
+      'Enable this to start using expiring offline access tokens and automatically refresh them when they are close to expiring.',
+    );
+  }
```
byrichardpowell: I'm not sure this is the correct use of this function.  `logDisabledFutureFlags` is for logging when a user is still using a future flag that we've since removed.  Here we use it to log that should adopt a new future flag, which is not the name of this function.  

I also _think_ it's not something we've done in the past. I can see the argument that we might want these logs to nudge migration, but it's not something we've done in the past? Are we sure want to add this to the users logs? If we do, I think it would be good to provide a link to a comprehensive guide that gives them everything they need to migrate.

packages/apps/shopify-app-react-router/src/server/helpers/__tests__/ensure-offline-token-is-not-expired.test.ts
```diff
@@ -0,0 +1,380 @@
+import {setUpValidSession as setUpValidSessionImport} from '@shopify/shopify-api/test-helpers';
```
byrichardpowell: In this package we focus on a more integrated style of tests.  For example, most of our tests involve passing a request to an authenticate function, having that execute E2E then checking either:

- The Response that was thrown
- The context object that was returned

I _think_ the tests in this file are more unit style.  

Please can we align on the approach that is most idiomatic for this codebase? I think that would be an `expectTokenRefresh` test helper which we use for every authenticate function that abstracts that behaviour.  You can look at `packages/apps/shopify-app-react-router/src/server/__test-helpers/expect-admin-api-client.ts` for an example of how that might work.

I think this change would save us lines of code and make refactoring easier since function API internals can change and as long as the external API behaviour has not changed, tests will still pass.

packages/apps/shopify-app-react-router/src/server/helpers/__tests__/get-offline-session.test.ts
```diff
@@ -0,0 +1,268 @@
+import {setUpValidSession as setUpValidSessionImport} from '@shopify/shopify-api/test-helpers';
```
byrichardpowell: Similar question to [this comment](https://github.com/Shopify/shopify-app-js/pull/2941/files#r2578025156) here

packages/apps/shopify-app-react-router/src/server/helpers/__tests__/refresh-token.test.ts
```diff
@@ -0,0 +1,168 @@
+import {HttpResponseError, InvalidJwtError} from '@shopify/shopify-api';
```
byrichardpowell, line 1: Similar question to this [comment here](https://github.com/Shopify/shopify-app-js/pull/2941/files#r2578025156)

packages/apps/shopify-app-react-router/src/server/helpers/ensure-offline-token-is-not-expired.ts
```diff
@@ -0,0 +1,33 @@
+import {Session} from '@shopify/shopify-api';
+
+import {AppDistribution, BasicParams} from '../types';
+
+import refreshToken from './refresh-token';
+
+// se if this needs to be moved
```
byrichardpowell: Comment needs to be adressed?

packages/apps/shopify-app-react-router/src/server/helpers/ensure-offline-token-is-not-expired.ts
```diff
@@ -0,0 +1,33 @@
...
+const WITHIN_MILLISECONDS_OF_EXPIRY = 1000;
+
+// test if accessible to public use
```
byrichardpowell: Comment needs to be adressed?

packages/apps/shopify-app-react-router/src/server/helpers/get-offline-session.ts
```diff
@@ -0,0 +1,12 @@
+import {BasicParams} from '../types';
+
+import {createOrLoadOfflineSession} from './create-or-load-offline-session';
+import {ensureOfflineTokenIsNotExpired} from './ensure-offline-token-is-not-expired';
+
+export async function getOfflineSession(params: BasicParams, shop: string) {
```
byrichardpowell: This function is doing a lot more than just getting an Offline Session.  I wouldn't have known that from the function name.  

I think we have 2 options:

1. Rename. 
2. Refactor. 

If we rename, I expect we'd have a long name, which actually could be a smell that this should be split into multiple functions and we accept a little repetition.  Explicit code > DRY Unexplicit code.

.changeset/metal-towns-flow.md
fwaadahmad1: You are right, But the react-router template ships with Prisma, so we updated only prisma for now. I can start working on all the other packages so we don't have this limitation anymore

.changeset/metal-towns-flow.md
fwaadahmad1: I don't believe we need to, because the Authentication factory for these return a `sessionToken`​ instead of a `session`​  
Reference: [Authentication Factory for checkout](https://github.com/Shopify/shopify-app-js/blob/fa273328065af81a58b2ff1a7b38cc9983a0ee62/packages/apps/shopify-app-react-router/src/server/authenticate/public/checkout/authenticate.ts#L9C3-L12C29) and [Authentication Factory for Customer account](https://github.com/Shopify/shopify-app-js/blob/fa273328065af81a58b2ff1a7b38cc9983a0ee62/packages/apps/shopify-app-react-router/src/server/authenticate/public/customer-account/authenticate.ts#L9C3-L12C36) both use [authenticateExtensionFactory](https://github.com/Shopify/shopify-app-js/blob/fa273328065af81a58b2ff1a7b38cc9983a0ee62/packages/apps/shopify-app-remix/src/server/authenticate/public/extension/authenticate.ts#L44C5-L52C7) which returns a `sessionToken` instead of a `session`

.changeset/metal-towns-flow.md
fwaadahmad1: Right

.changeset/metal-towns-flow.md
fwaadahmad1: I had add these steps in the docs but I will add them here too

packages/apps/shopify-app-react-router/src/server/authenticate/webhooks/authenticate.ts
fwaadahmad1: I will update both to have params as the first parameter

packages/apps/shopify-app-react-router/src/server/future/flags.ts
fwaadahmad1: I was not familiar with the usage of this function, I assumed it's functionality from the description provided for the function which said "When future flags are added to the FutureFlags interface, add logging here".   
Should I update the description as well along with reverting this?

fwaadahmad1: The docs for `shopify-api`​ package includes a guide that gives everything the user need to migrate

packages/apps/shopify-app-react-router/src/server/helpers/ensure-offline-token-is-not-expired.ts
fwaadahmad1: Forgot to remove

packages/apps/shopify-app-react-router/src/server/helpers/ensure-offline-token-is-not-expired.ts
fwaadahmad1: Forgot to remove

packages/apps/shopify-app-react-router/src/server/helpers/get-offline-session.ts
fwaadahmad1: How about these names:

- getValidOfflineSession
- **retrieveOfflineSession**
- **acquireOfflineSession**
- **loadValidOfflineSession**

.changeset/metal-towns-flow.md
byrichardpowell: Ah, yeah, makes sense.  Thanks!

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,50 @@
+---
+'@shopify/shopify-app-react-router': minor
+---
+
+We are introducing support for expiring offline access tokens. This feature improves security by limiting the lifespan of offline access tokens and automatically refreshing them using refresh tokens.
+
+- **New Future Flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`. When enabled, the library will start using expiring offline tokens and automatically check if it is expired or nearing expiration. If expired/expiring, it attempts to refresh the access token using the stored refresh token. Defaults to `false` for backward compatibility.
+
+- **Automatic Token Refresh**: Integrated token refresh logic into authentication flows (`flow`, `fulfillmentService`, `appProxy`, `webhooks`) and unauthenticated contexts (`admin`, `storefront`). When a session is loaded and found to be expired (or expiring within 1 second), and the feature is enabled, the library transparently refreshes the token and persists the new session data.
```
byrichardpowell: 1 second seems like a very small tolerence.  A network request could easily take a second.  Worth increasing this to 5 minutes?

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,50 @@
...
+  // ... other config
+  future: {
+    expiringOfflineAccessTokens: true,
+  },
+});
+```
+
+When enabled, calls to `shopify.authenticate.admin`, `shopify.authenticate.flow`, etc., will automatically handle token refreshing for offline sessions.
+
+- None. The feature is opt-in via the `expiringOfflineAccessTokens` future flag.
```
byrichardpowell: Should this bullet point be removed?

.changeset/metal-towns-flow.md
byrichardpowell: Thanks for updating 👍

.changeset/metal-towns-flow.md
byrichardpowell: Thanks for updating @fwaadahmad1 👍

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,50 @@
...
+'@shopify/shopify-app-react-router': minor
+---
+
+We are introducing support for expiring offline access tokens. This feature improves security by limiting the lifespan of offline access tokens and automatically refreshing them using refresh tokens.
+
+- **New Future Flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`. When enabled, the library will start using expiring offline tokens and automatically check if it is expired or nearing expiration. If expired/expiring, it attempts to refresh the access token using the stored refresh token. Defaults to `false` for backward compatibility.
+
+- **Automatic Token Refresh**: Integrated token refresh logic into authentication flows (`flow`, `fulfillmentService`, `appProxy`, `webhooks`) and unauthenticated contexts (`admin`, `storefront`). When a session is loaded and found to be expired (or expiring within 1 second), and the feature is enabled, the library transparently refreshes the token and persists the new session data.
+
+To enable expiring offline access tokens in your app, you must ensure your session storage can persist refresh tokens. If you are using Prisma, follow these steps:
```
byrichardpowell: I think we need to make it really obvious and really explicit that this will only work for Prisma.  E.g:

For now this will only work if you are using the Prisma Session Storage package. We will be updating the other session storage packages in the coming [TIMEFRAME]

.changeset/metal-towns-flow.md
byrichardpowell: Yeah, makes sense. Will resolve this comment, but please see: https://github.com/Shopify/shopify-app-js/pull/2941/files#r2585535839

.changeset/metal-towns-flow.md
byrichardpowell: Thanks!

packages/apps/shopify-app-react-router/src/server/authenticate/webhooks/authenticate.ts
byrichardpowell: Thanks 👍

packages/apps/shopify-app-react-router/src/server/future/flags.ts
byrichardpowell: Yeah, on reflection this is confusing and I'm not actually convinced I understand how we've used this function.  

Stepping back from this individual function. I think given we won't support all session storage packages at release a log bugging the developer to upgrade will be annoying.  So, I think we do this, in this order: 

1. Add support to all other session storage packages
2. Add back the log to upgrade if the future flag is not enabled

Not precious on what function we use to achieve this.

packages/apps/shopify-app-react-router/src/server/helpers/get-offline-session.ts
byrichardpowell: `ensureValidOfflineSession` or `retrieveOfflineSession` works for me.

packages/apps/shopify-app-react-router/src/server/authenticate/flow/__tests__/authenticate.test.ts
byrichardpowell: Thanks for making these changes.  Much nicer 👍

packages/apps/shopify-app-react-router/src/server/helpers/__tests__/ensure-offline-token-is-not-expired.test.ts
byrichardpowell: Thanks for making these  @fwaadahmad1

.changeset/metal-towns-flow.md
fwaadahmad1: I agree, The reason for this being 1 second was because we saw a similar implementation for online tokens; where we see if the online token is within `500 ms`​ of expiry then refresh it.  
  
Do we also want to revisit this for online tokens?

.changeset/metal-towns-flow.md
fwaadahmad1: Yes

.changeset/metal-towns-flow.md
byrichardpowell: Yeah, that would be good, thanks!

byrichardpowell: Thanks for updating 👍

.changeset/metal-towns-flow.md
```diff
@@ -0,0 +1,48 @@
...
+
+- **New Future Flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`. When enabled, the library will start using expiring offline tokens and automatically check if it is expired or nearing expiration. If expired/expiring, it attempts to refresh the access token using the stored refresh token. Defaults to `false` for backward compatibility.
+
+- **Automatic Token Refresh**: Integrated token refresh logic into authentication flows (`flow`, `fulfillmentService`, `appProxy`, `webhooks`) and unauthenticated contexts (`admin`, `storefront`). When a session is loaded and found to be expired (or expiring within 5 minutes), and the feature is enabled, the library transparently refreshes the token and persists the new session data. This behavior applies to both offline and online tokens.
+
+To enable expiring offline access tokens in your app, you must ensure your session storage can persist refresh tokens. For now, this will only work if you are using the Prisma Session Storage package. We will be updating the other session storage packages in the future. If you are using Prisma, follow these steps:
```
byrichardpowell: > For now, this will only work if you are using the Prisma Session Storage package. We will be updating the other session storage packages in the future.

This text should be earlier in the document.  Would be annoying to get this far then realize "Oh, this won't apply to me"

PR #2955, opened 2025-12-10
Implement domain transformations support for split-domain architectures

weslleyaraujo:
## Problem

Currently, multiple Shopify apps patch `@shopify/shopify-api` to enable local core development with split domain architecture. This creates maintenance burden and requires coordination across teams when the library changes.

[**Slack discussion**](https://shopify.slack.com/archives/C015JSW9X0X/p1765375432211379)

**Affected apps:**

- [Bundles](https://github.com/Shopify/bundles-app/blob/main/patches/@shopify+shopify-api+11.14.1.patch)
- [Subscriptions](https://github.com/Shopify/subscriptions-app-remix/)
- [Combined Listings App](https://github.com/Shopify/combined-listings-app/)

This PR introduces a proposal to solve this problem by adding a new `domainTransformations` configuration option that allows apps to define domain mappings without patching the library.

```ts
interface DomainTransformation {
  // Pattern to match source domain
  match: RegExp | string;

  // Transformation: function or template string
  transform: ((matches: RegExpMatchArray) => string | null) | string;

  // Whether to include in host validation (default: true)
  includeHost?: boolean;
}
```

**Template-based transformation (simple):**

```typescript
import { shopifyApi } from "@shopify/shopify-api";

const shopify = shopifyApi({
  // ... other config ...
  domainTransformations: [
    {
      match: /^([a-zA-Z0-9][a-zA-Z0-9-_]*)\.my\.shop\.dev$/,
      transform: "$1.dev-api.shop.dev",
    },
  ],
});
```

**Function-based transformation:**

```typescript
const shopify = shopifyApi({
  // ... other config ...
  domainTransformations: [
    {
      match: /^([a-zA-Z0-9-_]+)\.admin\.example\.com$/,
      transform: (matches) => {
        const shopName = matches[1];
        return shopName.startsWith("test-")
          ? `${shopName}.api-staging.example.com`
          : `${shopName}.api.example.com`;
      },
    },
  ],
});
```

For the affected apps, it would look like this:

```typescript
// app/shopify.server.ts
import { shopifyApp } from "@shopify/shopify-app-remix/server";

const shopify = shopifyApp({
  apiKey: process.env.SHOPIFY_API_KEY!,
  apiSecretKey: process.env.SHOPIFY_API_SECRET!,
  domainTransformations: isDevelopment
    ? [
        {
          match: /^([a-zA-Z0-9][a-zA-Z0-9-_]*)\.my\.shop\.dev$/,
          transform: "$1.dev-api.shop.dev",
        },
      ]
    : undefined,
});
```

### Demo

Here is an example of the bundles app running locally with a custom build version of this PR:
Screenshot 2025-12-10 at 14 33 03

PS: shopify-app-remix is also linked here, but is unrelated. Bundles has an outdated version, which is not compatible with the current shopify-app library.

### Resources

- https://github.com/shop/issues-bundles-and-combined-listings/issues/1250


packages/apps/shopify-api/lib/utils/__tests__/domain-transformer.test.ts
```diff
@@ -0,0 +1,249 @@
...
+        },
+      ],
+    } as ConfigInterface;
+
+    expect(
+      applyDomainTransformations('shop1.ui.example.com', config),
+    ).toBeNull();
+  });
+
+  it('handles complex transformation logic', () => {
```
lizkenyon: Is this test neccessary or is covered by `applies function-based transformations`

packages/apps/shopify-api/lib/utils/domain-transformer.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  for (const transformation of transformations) {
+    const regex =
+      typeof transformation.match === 'string'
+        ? new RegExp(transformation.match)
+        : transformation.match;
+
+    const domainPattern = regex.source.match(/\\\.([\\.\w-]+)\$?$/);
+
+    if (domainPattern) {
```
lizkenyon, line 108: We should document what regex patterns this support? 
Claude gave me some examples where this wouldn't always work `(\.staging)?\.example\.com$` or `\.api[0-9]\.example\.com$`

We should at least have some debug level logging, for those scenarios, so developer can debug a bit easier.

packages/apps/shopify-api/docs/reference/shopifyApi.md
lizkenyon: ✂️

lizkenyon: we can remove

packages/apps/shopify-api/lib/utils/__tests__/domain-transformer.test.ts
weslleyaraujo: yea that makes sense 🙂 I went ahead and removed it: https://github.com/Shopify/shopify-app-js/pull/2955/changes/6fed3845b58167c82d94697acd59af0edbb4ddad

packages/apps/shopify-api/lib/utils/domain-transformer.ts
weslleyaraujo: thats a good call! I've added more info into the JSDoc and now I am logging if the match doesn't work out: https://github.com/Shopify/shopify-app-js/pull/2955/changes/6fed3845b58167c82d94697acd59af0edbb4ddad

packages/apps/shopify-api/docs/reference/shopifyApi.md
weslleyaraujo: updated https://github.com/Shopify/shopify-app-js/pull/2955/changes/6fed3845b58167c82d94697acd59af0edbb4ddad 🙏

weslleyaraujo: removed here: https://github.com/Shopify/shopify-app-js/pull/2955/changes/6fed3845b58167c82d94697acd59af0edbb4ddad thank you!

packages/apps/shopify-api/lib/utils/domain-transformer.ts
```diff
@@ -0,0 +1,125 @@
...
+ * @param transformations - Array of domain transformations
+ * @returns Array of domain regex patterns
+ *
+ * @example
+ * const transformations = [{
+ *   match: /^([a-zA-Z0-9-_]+)\.my\.shop\.dev$/,
+ *   transform: '$1.dev-api.shop.dev'
+ * }];
+ * getTransformationDomains(transformations);
+ * // Returns: ['my\\.shop\\.dev', 'dev-api\\.shop\\.dev']
```
lizkenyon, line 88: Everywhere else in the utils the full `ConfigInterface` gets passed through — e.g. [`fetch-request.ts:11`](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-api/lib/utils/fetch-request.ts#L11), [`hmac-validator.ts:169`](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-api/lib/utils/hmac-validator.ts#L169).

Probably easiest to just pass `config` into `getTransformationDomains` so the debug message goes through the app's real logger.

weslleyaraujo: oh awesome, I had a claude version with this but I wasn't sure if this was really a pattern to follow, makes sense now! 😅 updated here: a58b6028f504d0b8eaf5241ff217d6bc65ae3380

weslleyaraujo:
**Update**

@lizkenyon 👋 pushed this: a58b6028f504d0b8eaf5241ff217d6bc65ae3380, updating the logger to follow the config pattern and setting packages to minor updates as per the comment 🙂

PR #2997, opened 2026-01-19
Update drizzle session mappings for user and refresh token fields

sanghogu:
drizzle session is too old. There is no refresh-related data.

Update drizzle's session storage data.

## Type of change

- [ v] Patch: Bug (non-breaking change which fixes an issue)


sanghogu:
I have signed the CLA!

PR #3028, opened 2026-02-09
Support new webhook headers

lizkenyon:
## Summary

- Add support for verifying webhooks delivered with the new lowercase `shopify-*` headers alongside the existing `X-Shopify-*` headers
- Auto-detect which header format is present and validate accordingly
- Refactor `WebhookFields` to a discriminated union on `webhookType` for type-safe field access
- Surface new webhook fields (`webhookType`, `handle`, `action`, `resourceId`, `triggeredAt`, `eventId`) in React Router and Remix webhook context
- Guard `process()` to reject non-traditional webhooks (which use TOML-based registration, not programmatic)

### Packages affected

| Package | Bump | Reason |
|---------|------|--------|
| `@shopify/shopify-api` | **major** | `WebhookFields` is now a discriminated union — consumers must narrow on `webhookType` to access type-specific fields |
| `@shopify/shopify-app-react-router` | minor | New optional fields added to webhook context |
| `@shopify/shopify-app-remix` | minor | New optional fields added to webhook context |

### How it works

Webhooks may now be delivered with new lowercase `shopify-*` headers. The library detects which format is present by checking for `shopify-hmac-sha256`, then extracts the appropriate type-specific fields.

For the framework packages (React Router / Remix), `webhookId` is set to the `eventId` value for backwards compatibility when new headers are present. This will be removed in the next major version.


.changeset/webhook-events-discriminated-union.md
```diff
@@ -0,0 +1,45 @@
+---
+'@shopify/shopify-api': major
+'@shopify/shopify-app-react-router': minor
+'@shopify/shopify-app-remix': minor
+---
+
+Add support for verifying webhooks delivered with the new `shopify-*` headers (replacing the previous `x-shopify-*` headers), and refactor webhook validation types to a discriminated union on `webhookType`.
+
+**Breaking change in `@shopify/shopify-api`:** `WebhookFields` is now a discriminated union (`WebhooksWebhookFields | EventsWebhookFields`) keyed on the required `webhookType` field. `webhookId` only exists on `WebhooksWebhookFields`; `eventId` is required on `EventsWebhookFields`. Consumers must narrow on `webhookType` to access type-specific fields. Both `WebhooksWebhookFields` and `EventsWebhookFields` are exported for use in type narrowing.
```
lizkenyon, line 9: I did this as a breaking change, as we are already planning another breaking change for shopify-api. 

For the framework packages I did not do a breaking change, and the fields are optional.

I was trying to balance the friction of breaking changes, while still providing the proper types. I do think in the next major version of the RR package we should tighten up the types, but does that have to be now?

packages/apps/shopify-api/lib/types.ts
```diff
@@ -35,12 +35,27 @@ export enum ShopifyHeader {
...
+  Name = 'X-Shopify-Name',
+  TriggeredAt = 'X-Shopify-Triggered-At',
+  EventId = 'X-Shopify-Event-Id',
   StorefrontPrivateToken = 'Shopify-Storefront-Private-Token',
   StorefrontSDKVariant = 'X-SDK-Variant',
   StorefrontSDKVersion = 'X-SDK-Version',
 }
 /* eslint-enable @shopify/typescript/prefer-pascal-case-enums */
 
+export const ShopifyEventsHeader = {
```
lizkenyon, line 47: Shopify Header enum has a mix of webhook and non webhook headers, but I thought it might be confusing to add the new headers there?

PR #3036, opened 2026-02-13
[Patch] Publish TypeScript source files for Go to Definition support

lizkenyon:

When consumers use "Go to Definition" in their IDE on imports from our packages, they land on compiled `.d.ts` declaration files instead of the real TypeScript source. Our source maps already point to the correct relative paths — we just weren't including the source files they reference in the published npm packages.

Inspired by [Preact's approach](https://unpkg.com/browse/preact@10.26.8/) — they ship `src/` to npm and their source maps in `dist/` point to `../src/`.

Adds TypeScript source directories (`src/` or `lib/`) to the `files` field in all 20 published packages. Test files are excluded via negation patterns to keep package size minimal.

No build config, tsconfig, or source map changes were needed.

### Before

https://github.com/user-attachments/assets/7fc9f200-649c-4f58-bcc2-5ce354296b70

### After

https://github.com/user-attachments/assets/e4bda743-d411-4bbb-8709-c6e8f1848d55

## Type of change

## Checklist


lizkenyon:
Did a quick package size analysis on the two largest packages in this PR.

**`@shopify/shopify-api`**

| | Files | Unpacked Size |
|---|---|---|
| Before | 6,847 | 28.1 MB |
| After | 7,999 | 32.7 MB |
| Delta | +1,152 | +4.6 MB (+16%) |

Breakdown of added source files:

| Directory | Files | Size |
|---|---|---|
| `rest/` | 1,046 | 4.2 MB |
| `lib/` | 74 | 0.2 MB |
| `runtime/` | 13 | ~20 KB |
| `adapters/` | 10 | ~30 KB |
| `test-helpers/` | 8 | ~10 KB |
| `future/` | 1 | ~4 KB |

`rest/` accounts for 91% of the size increase — it contains ~1,046 generated REST resource files across multiple API versions. These are already shipped 3x over in `dist/` (cjs, esm, ts), so the source copies add a relatively modest 4.2 MB on top of the existing 28 MB package. Could consider excluding `rest` from the `files` field since those are generated/codegen'd resources that aren't particularly useful to navigate to.

**`@shopify/shopify-app-react-router`**

| | Files | Unpacked Size |
|---|---|---|
| Before | 1,242 | 1.9 MB |
| After | 1,404 | 2.2 MB |
| Delta | +162 | +0.3 MB (+16%) |

All 162 added files are from `src/` — straightforward TypeScript source. Very clean increase here.

PR #3047, opened 2026-02-17
feat: add refresh token support to MySQL session storage

byrichardpowell:
## Summary

Adds refresh token support to the MySQL session storage package, following the pattern from #2938 and using the centralized `batteryOfTests` suite for refresh token testing.

## Changes

- **Database Migration**: Added automatic migration to add `refreshToken` (text) and `refreshTokenExpires` (timestamp) columns
- **Schema**: Updated CREATE TABLE to include refresh token fields for new installations
- **Serialization**: Convert `refreshTokenExpires` from milliseconds to seconds when storing
- **Deserialization**: Convert `refreshTokenExpires` from seconds to milliseconds when loading
- **Tests**: Enabled refresh token tests via `batteryOfTests(sessionStorage, false, true)`
- **Documentation**: Added MIGRATION_TO_EXPIRING_TOKENS.md with upgrade instructions
- **Changeset**: Minor version bump with migration guide

## Migration Required

**Yes** - See MIGRATION_TO_EXPIRING_TOKENS.md for schema update instructions.

The migration runs automatically on the first connection after upgrading. Manual migration SQL is also provided in the migration guide for users who prefer to run it manually.

## Testing

- ✅ Build passed
- ✅ Type check passed
- ✅ Lint passed
- ⚠️ Tests require Docker/Podman (will run in CI)
- ✅ Backward compatibility verified (nullable columns)

## Dependencies

Depends on #3041 (centralized refresh token tests in batteryOfTests)

## Review Focus

- Migration system integration (follows existing pattern)
- Timestamp handling (milliseconds to seconds conversion)
- Backward compatibility (nullable columns, automatic migration)

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) using the `session-storage:add-refresh-token-support` skill

Related: #2938, #3041

### DIFF
--- a/.changeset/session-storage-mysql-refresh-tokens.md
+++ b/.changeset/session-storage-mysql-refresh-tokens.md
@@ -8,7 +8,7 @@ Add support for storing refresh tokens and refresh token expiration dates. This 
 
 Two new columns have been added to the session storage table to support expiring offline access tokens:
 
-- `refreshToken` (text, nullable) - Stores the refresh token used to obtain new access tokens
+- `refreshToken` (varchar(255), nullable) - Stores the refresh token used to obtain new access tokens
 - `refreshTokenExpires` (integer, nullable) - Stores the expiration date of the refresh token
 
 ## Automatic Migration
@@ -23,7 +23,7 @@ If you prefer to run the migration manually before upgrading, execute the follow
 
 ```sql
 ALTER TABLE shopify_sessions
-  ADD COLUMN refreshToken text,
+  ADD COLUMN refreshToken varchar(255),
   ADD COLUMN refreshTokenExpires integer NULL;
 ```
 
--- a/packages/apps/session-storage/shopify-app-session-storage-mysql/MIGRATION_TO_EXPIRING_TOKENS.md
+++ b/packages/apps/session-storage/shopify-app-session-storage-mysql/MIGRATION_TO_EXPIRING_TOKENS.md
@@ -4,7 +4,7 @@
 
 Two new columns have been added to the session storage table to support expiring offline access tokens:
 
-- `refreshToken` (text, nullable) - Stores the refresh token used to obtain new access tokens
+- `refreshToken` (varchar(255), nullable) - Stores the refresh token used to obtain new access tokens
 - `refreshTokenExpires` (integer, nullable) - Stores the expiration date of the refresh token
 
 ## Automatic Migration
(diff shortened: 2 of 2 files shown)


packages/apps/session-storage/shopify-app-session-storage-mysql/src/mysql.ts
```diff
@@ -163,7 +162,9 @@ export class MySQLSessionStorage implements SessionStorage {
           scope varchar(255),
           expires integer,
           onlineAccessInfo varchar(255),
-          accessToken varchar(255)
+          accessToken varchar(255),
+          refreshToken text,
+          refreshTokenExpires integer
```
lizkenyon, line 167: This should be `varchar(255)` for consistency with the `accessToken` column on line 166, and with how the SQLite and PostgreSQL adapters define this column in their refresh token PRs (#3042 and #3044).

packages/apps/session-storage/shopify-app-session-storage-mysql/README.md
```diff
@@ -38,4 +38,18 @@ const shopify = shopifyApp({
 });
 ```
 
+## Scalability
```
lizkenyon: This section header doesn't match the content — the section is about refresh token support / expiring offline access tokens, not scalability. The PostgreSQL PR (#3044) uses `## Expiring Offline Access Tokens` for the same content, which is more accurate.

packages/apps/session-storage/shopify-app-session-storage-mysql/src/migrations.ts
```diff
@@ -7,8 +7,33 @@ export const migrationList = [
...
+  connection: MySqlConnection,
+): Promise<void> {
+  // Check if columns already exist before attempting to add them
+  const [rows] = await connection.query(
+    `SELECT COLUMN_NAME
+     FROM INFORMATION_SCHEMA.COLUMNS
+     WHERE TABLE_SCHEMA = DATABASE()
+       AND TABLE_NAME = ?
+       AND COLUMN_NAME IN ('refreshToken', 'refreshTokenExpires')`,
+    [connection.sessionStorageIdentifier],
```
lizkenyon: Nit: The PostgreSQL PR (#3044) checks each column independently and only adds the missing one(s). This approach skips the entire migration if *either* column exists, which could leave the table in a partial state if someone manually added only one column. Low risk in practice, but checking each column separately would be more robust.

byrichardpowell: Fixed 👍

packages/apps/session-storage/shopify-app-session-storage-mysql/README.md
byrichardpowell: Fixed 👍

packages/apps/session-storage/shopify-app-session-storage-mysql/src/mysql.ts
byrichardpowell: Fixed 👍

PR #3051, opened 2026-02-19
feat: store full online access user data in session storage adapters (KV, SQLite, MySQL, PostgreSQL)

byrichardpowell:
## Summary

Four session storage adapters (KV, SQLite, MySQL, PostgreSQL) were silently discarding all online access user fields beyond the numeric user ID. This PR fixes that by passing the `includeUserData` flag to `toPropertyArray` / `fromPropertyArray` in every affected adapter, and adds the required schema migrations for the SQL backends.

- **KV** (`minor`): schema-less so no migration needed. Backward-compatible with sessions stored in the old format.
- **SQLite** (`major`): `addUserInfoColumns` migration uses the table-recreation pattern (BEGIN/COMMIT), consistent with the existing `migrateScopeFieldToVarchar1024` migration. `createTable()` updated to the new schema.
- **MySQL** (`major`): `addUserInfoColumns` migration uses idempotent steps (INFORMATION_SCHEMA guards) to work around DDL auto-commit semantics. `createTable()` updated to the new schema.
- **PostgreSQL** (`major`): `addUserInfoColumns` migration runs inside `connection.transaction()` for clean rollback on failure. Fixes `migrateToCaseSensitivity` INSERT to extract `userId` from `onlineaccessinfo` instead of copying the now-removed column. `createTable()` updated to the new schema.
- **Memory / Drizzle**: were already correct. Their `batteryOfTests` calls updated from `false` → `true` to exercise the path in CI.  No changelog since it's a test only change.

Also update batteryOfTests to remove `testUserInfo` and `testRefreshTokens` arguments since all session storage packages now always pass true, or already adopted the behaviour that would have been tested if true was passed.  No point keeping flags around when they are needed.

## Breaking changes

SQL adapters (SQLite, MySQL, PostgreSQL) replace the single `onlineAccessInfo varchar(255)` column with seven typed columns. Automatic migrations run on first connection. Manual migration SQL is provided in each changeset for those who prefer to migrate before deploying.

Apps that query `shopify_sessions` directly and read `onlineAccessInfo` will need to update those queries to use the new individual columns.

The session storage test suite package hasa breaking change, but thats an internal package only.

### DIFF
--- a/.changeset/full-user-data-sqlite.md
+++ b/.changeset/full-user-data-sqlite.md
@@ -49,13 +49,16 @@ CREATE TABLE shopify_sessions (
   accountOwner integer,
   locale varchar(255),
   collaborator integer,
-  emailVerified integer
+  emailVerified integer,
+  refreshToken varchar(255),
+  refreshTokenExpires integer
 );
 
--- Copy data, preserving userId from the old onlineAccessInfo column
-INSERT INTO shopify_sessions (id, shop, state, isOnline, expires, scope, accessToken, userId)
+-- Copy data, preserving userId from the old onlineAccessInfo column and refresh token fields
+INSERT INTO shopify_sessions (id, shop, state, isOnline, expires, scope, accessToken, userId, refreshToken, refreshTokenExpires)
   SELECT id, shop, state, isOnline, expires, scope, accessToken,
-    CASE WHEN onlineAccessInfo IS NOT NULL THEN CAST(onlineAccessInfo AS INTEGER) ELSE NULL END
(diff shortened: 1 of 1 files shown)


lizkenyon:
The manual migration SQL in `.changeset/full-user-data-sqlite.md` is missing `refreshToken` and `refreshTokenExpires` columns. Since we have two PRs shipping schema changes to these tables in the same release, and SQLite uses a table-recreation pattern (rename → create → insert → drop), the manual SQL in this changeset needs to account for the columns added by the other PR — otherwise someone following it would lose their refresh token data.

The automatic migration in `sqlite/src/migrations.ts` already handles this correctly. MySQL and PostgreSQL are fine since their manual SQL uses additive `ALTER TABLE ADD COLUMN`.

PR #3053, opened 2026-02-20
feat: automated preview docs sync workflow

majd-shopify:
## Problem
Closes https://github.com/Shopify/temp-project-mover-Archetypically-20260312105649/issues/16 

Context is in the issue since this is a public repo

## Top-hatting
- https://github.com/Shopify/shopify-app-js/pull/3052
- https://github.com/Shopify/shopify-dev/pull/68420

### DIFF
--- a/.github/workflows/shopify-dev-preview-automation.yml
+++ b/.github/workflows/shopify-dev-preview-automation.yml
@@ -3,7 +3,7 @@
 #
 # shopify-dev in the monorepo will then sync these docs files for preview.
 
-name: Shopify Dev Docs Sync
+name: Shopify Dev Docs Preview
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


majd-shopify:
@lizkenyon this is now good to go!

.github/workflows/shopify-dev-preview-automation.yml
```diff
@@ -0,0 +1,183 @@
+# This workflow dispatches a docs sync event to shopify-dev in the monorepo when generated docs files change.
+# It monitors changes to generated docs files in PRs.
+#
+# shopify-dev in the monorepo will then sync these docs files for preview.
+
+name: Shopify Dev Docs Sync
```
lizkenyon: Could we name this Shopify Dev Docs Preview? 

To confirm this does not actually sync the content. That would need to happen in a separate process?

majd-shopify: Oh this does sync the content, it copies all the content to the correct places in shopify-dev, if you like the changes you can merge the PR that gets generated in shop/world! :)

majd-shopify: I can definitely rename it if you prefer! but its name is the same in all the other repos it was added to!

PR #3058, opened 2026-03-02
chore: migrate ESLint → Oxlint

byrichardpowell:
## Why

ESLint (with `@shopify/eslint-plugin`) has accumulated significant startup overhead and transitive dependencies. Oxlint is a Rust-based linter that is ~50–100x faster, with a much smaller dependency footprint.

## What

- **Replace ESLint with Oxlint**: deletes `eslint.config.js`, adds `oxlint.json` with equivalent rules (TypeScript, import, and promise plugins)
- **Update all 20 package `lint` scripts** from `eslint .` to `oxlint .`
- **Remove ESLint devDependencies** (`eslint`, `@shopify/eslint-plugin`, `eslint-plugin-jest`) and add `oxlint@^1.0.0`
- **Decouple Prettier**: ESLint previously ran Prettier as a plugin. Prettier now runs as a standalone `prettier` script (`prettier --check .`) separate from linting. The `.prettierignore` glob patterns for Remix/React Router example files are also fixed (`*.example.tsx?` → explicit `.ts`/`.tsx` patterns)
- **Fix 42 Oxlint violations** surfaced by stricter rule enforcement:
  - `no-non-null-asserted-optional-chain` — billing helpers and Express config
  - `no-unsafe-optional-chaining` — `request-mock.ts` (Remix + React Router)
  - `no-unused-vars` — catch params renamed to `_error`/`_err`, unused imports removed

## Testing

All CI checks pass (lint + tests across Node 20/22/24).


Suleimanlatrsh:
Moving this back to draft. The direction is good and oxlint did catch real issues, but it isn't safe to merge as-is and the green checks are misleading.

A few things going on:
- The suite passes partly because it checks less than before. Unused-vars is downgraded to a warning, which can't fail CI. Prettier is no longer run in CI at all. And the per-package lint scripts run bare oxlint, which doesn't load the new oxlint.json (oxlint only auto-loads .oxlintrc.json), so most packages lint with default settings.
- A couple of the billing changes swap optional chaining for non-null assertions (?.x?.y! becomes !.x!.y), which turns a possibly-missing value into a thrown error at runtime instead of undefined.

Given it's a large, months-old change that also touches the lockfile, it would be much easier to land safely split into three: the prettier decoupling, the code fixes, then the linter swap on its own. Happy to help with any of those.

byrichardpowell:
Closing as this has been addressed in another PR

byrichardpowell:
Closing as this has been addressed in another PR

Issue #427, opened 2023-09-04
Crashes On 50x API Responses


exp131:
# Issue summary

Faced a very strange issue using Node.js as the app backend with this library.
We have an auth middleware that utilizes shopify.validateAuthenticatedSession():

```
// middleware

const customAuthMiddleware = async (req, res, next) => {
  try {
    shopify.validateAuthenticatedSession()(req, res, next);
  } catch (e) {
    console.log(e);
  }
};

app.use('/api/listing', customAuthMiddleware, MainRoutes);
```

And when we get an API response from Shopify with code > 500, the try...catch block doesn't work, and the app crashes. It happens randomly and depends only on Shopify API response. If the response is successful - everything is OK, but if not - it crashes (with different codes, 500, 502, 503 or 504).
![image](https://github.com/Shopify/shopify-app-js/assets/18207837/1dcf0507-6ba8-48d1-93a1-a2b1e0a154bf)
And the weird thing is that we can't catch and process errors. It seems some async function inside the library prevents the error from being caught. So, the user sees just a white screen, and the app fails to load—very bad UX.

Does anybody know how to fix such behavior?

PR #3100, opened 2026-03-13
Fix server crash on 5xx errors in validateAuthenticatedSession

### LINKED ISSUE #1860: Shopify 503 errors in validateAuthenticatedSession crash server
The middleware function hasValidAccessToken runs a graphql query to check authentication. If this query fails due to a 503 error, which will occasionally occur at random from Shopify api, the function throws an error which isn't handled by the caller (validateAuthenticatedSession).
This causes my whole server to go down randomly because I can't catch the errors anywhere due to them being async.

Also see https://github.com/Shopify/shopify-app-js/issues/427

## Expected behavior
50x errors should cause the middleware to fail normally rather than going uncaught

hasValidAccessToken() should be wrapped in a try/catch block inside validateAuthenticatedSession and have errors handled by handleSessionError()

byrichardpowell:

Fixes #1860

When Shopify returns a non-401 HTTP error (e.g. 503 Service Unavailable) during session token validation, `hasValidAccessToken` re-throws the error. In `validateAuthenticatedSession`, this call was unguarded — the error escaped the async Express middleware as an unhandled promise rejection, crashing the server.

This is a recurring issue: it was also reported in #427 (2023, closed stale without a fix).

Wraps the `await hasValidAccessToken(api, session)` call in `validateAuthenticatedSession` with a try/catch. On any non-401 error, it logs the error and returns `false`, treating the session as invalid and redirecting to re-authentication — the same defensive pattern already used by `sessionHasValidAccessToken` in `ensureInstalledOnShop`.

**Before:** transient Shopify 5xx errors crash the server with an unhandled promise rejection.

**After:** transient errors are caught, logged, and the user is redirected to re-authenticate instead.

Also adds a test for the 503 case, which was previously uncovered.

## Type of change

## Checklist

### DIFF
--- a/.changeset/fix-validate-session-503-crash.md
+++ b/.changeset/fix-validate-session-503-crash.md
@@ -1,0 +1,7 @@
+---
+"@shopify/shopify-app-express": patch
+---
+
+Fix server crash when Shopify returns 5xx errors during session token validation.
+
+`validateAuthenticatedSession` now catches non-401 HTTP errors thrown by `hasValidAccessToken` (e.g. 503 Service Unavailable) and treats them as an invalid session, redirecting to re-authentication instead of crashing the server with an unhandled promise rejection.
--- a/packages/apps/shopify-app-express/src/middlewares/__tests__/validate-authenticated-session.test.ts
+++ b/packages/apps/shopify-app-express/src/middlewares/__tests__/validate-authenticated-session.test.ts
@@ -225,6 +225,22 @@ describe('validateAuthenticatedSession', () => {
 
       expect((response.error as any).text).toBe('Storage error');
     });
+
+    it('redirects to auth if Shopify returns a 503 during token validation', async () => {
+      mockShopifyResponse({errors: 'Service Unavailable'}, {status: 503});
+
+      const response = await request(app)
+        .get('/test/shop?shop=my-shop.myshopify.io')
+        .set({Authorization: `Bearer ${validJWT}`})
+        .expect(403);
+
+      expect(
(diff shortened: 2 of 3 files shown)


Suleimanlatrsh:
Approving. Verified locally: the middleware suite passes, and reverting the try/catch makes the new 503 test fail, confirming it reproduces the crash. Clean fix.

Optional: the title says 5xx but only 503 is covered. A 500 or 502 case would match the title fully. Not needed to merge.

Suleimanlatrsh (approved):

Suleimanlatrsh:
I reproduced the original issue end to end in shopify-app-template-node.

Setup: seeded an active offline session, gated route is the template's default `app.use("/api/*", shopify.validateAuthenticatedSession())`, and forced the token-validation GraphQL call (hasValidAccessToken) to return 503 via a network intercept (Shopify does return 5xx during outages; this simulates that upstream response).

Result:
- Shipped shopify-app-express: the unguarded throw becomes an unhandled promise rejection and the server process crashes (subsequent requests get connection refused). This reproduces #1860.
- With this PR's try/catch: the request returns 403 with X-Shopify-API-Request-Failure-Reauthorize + Reauthorize-Url headers, and the server stays up across requests.

So the original crash is reproduced on the shipped build and resolved by this PR, verified in the Node template.

Suleimanlatrsh:
Final status: approved, green. Reproduced the original crash (#1860) in shopify-app-template-node with the 5xx simulated (a real 5xx can't be forced without a Shopify outage); shipped crashes, fix returns a clean 403 and stays up. Ready to merge.

PR #3101, opened 2026-03-13
Fix app proxy HMAC validation dropping duplicate query parameters

### LINKED ISSUE #2374: App Proxy authenticate.public.appProxy Fails Signature Validation with Specific Query Parameter Names (e.g., consented, consentGiven)
Okay, here's the bug report filled out using the template and the information we gathered:

# Issue summary

-   [x] Deps
    -   `@shopify/shopify-app-remix` package and version 3.8.1
    -   Node version: 20.18.1
    -   Operating system: macOS 15.4
-   [x] Set `{ logger: { level: LogSeverity.Debug } }` in my configuration, when applicable (Logs provided below confirm this)
-   [x] Found a reliable way to reproduce the problem that indicates it's a problem with the package
-   [x] Looked for similar issues in this repository 
-   [x] Checked that this isn't an issue with a Shopify API (Manual signature verification confirms the API/signature is correct)

When using an App Proxy route with Remix (specifically an `action` handling POST requests), the `shopify.authenticate.public.appProxy(request)` function fails signature validation if certain query parameters, such as `consented` or `consentGiven`, are present in the request URL. The same function works correctly for GET/DELETE requests or if the conflicting parameters are renamed (e.g., to `cGiven`). Manual signature verification confirms the signature received from Shopify *is* valid for the request including the parameters that cause the failure. This suggests specific parameter names interfere with the internal validation logic for POST requests within `authenticate.public.appProxy`.

## Expected behavior

`shopify.authenticate.public.appProxy` should successfully validate the App Proxy request signature for POST requests regardless of the presence of additional, valid custom query parameters like `consented` or `consentGiven`, provided the signature matches the full set of parameters used by Shopify during signing.

## Actual behavior

`shopify.authenticate.public.appProxy` fails signature validation for POST requests when query parameters named `consented` or `consentGiven` are included, returning a 400 Bad Request ("invalid signature"), despite the signature being mathematically valid for the received parameters.

1.  Set up a Remix app using `@shopify/shopify-app-remix` and configure an App Proxy URL in the Shopify Partner Dashboard pointing to the app (e.g., via a tunnel like Cloudflare).
2.  Create an App Proxy route file (e.g., `app/routes/proxy.consent.record.tsx`) with an `action` function that handles POST requests.
3.  Inside the `action`, call `await shopify.authenticate.public.appProxy(request)`.
4.  From the Shopify storefront context (e.g., using `fetch` in theme JS or an SDK), make a `POST` request to the corresponding proxy path (e.g., `/apps/proxy-prefix/consent/record`). Include a custom query parameter named `consentGiven` (e.g., `/apps/proxy-prefix/consent/record?consentGiven=true`). Shopify will automatically add the standard proxy parameters.
5.  **Observe:** The Remix server logs show `shopify.authenticate.public.appProxy` throws a `Response` with status 400 ("invalid signature").
6.  **(Optional):** Add manual signature verification code (using Node `crypto.createHmac`) inside the `action` *before* the `authenticate.public.appProxy` call. Observe that the manual check returns `true`.
7.  **Modify Test:** Rename the custom query parameter in both the frontend `fetch` call and the backend `action`'s logic to something neutral, like `cGiven=true`.
8.  Repeat the POST request from step 4 using the renamed parameter `cGiven=true`.
9.  **Observe:** The `shopify.authenticate.public.appProxy` call now succeeds.

## Debug logs

```
// Log showing POST request failure with parameter "consentGiven=true"
21:33:05 │ remix │ Received Request URL: https://players-exam-cancel-inspiration.trycloudflare.com/proxy/consent/record?consentGiven=true&shop=test-234...&logged_in_customer_id=232...&path_prefix=%2Fapps%2Floyalty-motion&timestamp=1745436785&signature=...
21:33:05 │ remix │ Received Request Headers: Headers { ... 'host': 'players-exam-cancel-inspiration.trycloudflare.com', 'x-forwarded-host': 'test-234***.myshopify.com', ... }
21:33:05 │ remix │ [shopify-app/INFO] Authenticating app proxy request | {shop: test-234***.myshopify.com}
21:33:05 │ remix │ [shopify-app/INFO] App proxy request has invalid signature | {shop: test-234***.myshopify.com}
21:33:05 │ remix │ Error in proxy.consent.record.tsx authenticate: Response { status: 400, statusText: 'Bad Request', ... }
21:33:05 │ remix │ App Proxy authentication resulted in a Response (likely redirect or auth error):

// Log showing successful manual signature verification for the same request:
// (This log would be generated if manual check code was added before the library call)
// MANUAL CHECK - Received Signature: ...
// MANUAL CHECK - Calculated Signature: ... (matches received)
// MANUAL CHECK - Query String for Hashing: consentGiven=true&logged_in_customer_id=...&path_prefix=...&shop=...&timestamp=...
// MANUAL CHECK - Signatures Match: true

// Log showing POST request success after renaming "consentGiven" to "cGiven"
// (Assuming success after renaming - replace with actual log if available)
// Received Request URL: https://players-exam-cancel-inspiration.trycloudflare.com/proxy/consent/record?cGiven=true&shop=test-234***&logged_in_customer_id=232...&path_prefix=%2Fapps%2Floyalty-motion&timestamp=...&signature=...
// Received Request Headers: Headers { ... 'host': 'players-exam-cancel-inspiration.trycloudflare.com', 'x-forwarded-host': 'test-234***.myshopify.com', ... }
// [shopify-app/INFO] Authenticating app proxy request | {shop: test-234***.myshopify.com}
// Consent recorded successfully for customer 232... in shop test-234***.
```

byrichardpowell:

Fixes #2374

`authenticate.public.appProxy()` throws a spurious 400 for requests where a query parameter appears more than once. `Object.fromEntries(searchParams.entries())` silently drops duplicate keys before HMAC validation, producing a different string than Shopify signed.

`stringifyQueryForAppProxy` in `hmac-validator.ts` already handles arrays by joining with commas, but never received arrays because `Object.fromEntries` collapsed them first.

Adds a `searchParamsToQuery()` helper that iterates `URLSearchParams` entries preserving duplicates as arrays, and replaces all three `Object.fromEntries(searchParams.entries())` calls inside `validateAppProxyHmac` in both `shopify-app-remix` and `shopify-app-react-router`.

Also fixes the `createAppProxyHmac` test helper which had the same bug (masking the issue in tests), and adds a regression test for the duplicate-parameter case.

## Type of change

## Checklist

### DIFF
--- a/.changeset/fix-app-proxy-hmac-duplicate-params-react-router.md
+++ b/.changeset/fix-app-proxy-hmac-duplicate-params-react-router.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-react-router': patch
+---
+
+Fix app proxy HMAC validation failing when a URL query parameter appears more than once. `Object.fromEntries` was silently dropping duplicate keys before HMAC validation, producing a different string than Shopify signed. The fix preserves all values for repeated parameters and joins them with commas, matching Shopify's signing behaviour.
--- a/.changeset/fix-app-proxy-hmac-duplicate-params-remix.md
+++ b/.changeset/fix-app-proxy-hmac-duplicate-params-remix.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-remix': patch
+---
+
+Fix app proxy HMAC validation failing when a URL query parameter appears more than once. `Object.fromEntries` was silently dropping duplicate keys before HMAC validation, producing a different string than Shopify signed. The fix preserves all values for repeated parameters and joins them with commas, matching Shopify's signing behaviour.
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/public/appProxy/__tests__/authenticate.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/public/appProxy/__tests__/authenticate.test.ts
@@ -405,21 +405,19 @@ async function getValidRequest(): Promise<Request> {
 }
 
 async function createAppProxyHmac(url: URL): Promise<string> {
-  const params: Record<string, string | string[]> = {};
+  const params: Record<string, string> = {};
   for (const [key, value] of url.searchParams.entries()) {
     const existing = params[key];
     if (existing === undefined) {
       params[key] = value;
     } else {
-      params[key] = Array.isArray(existing)
-        ? [...existing, value]
-        : [existing, value];
+      params[key] = `${existing},${value}`;
     }
   }
   const string = Object.entries(params)
     .sort(([val1], [val2]) => val1.localeCompare(val2))
     .reduce((acc, [key, value]) => {
-      return `${acc}${key}=${Array.isArray(value) ? value.join(',') : value}`;
+      return `${acc}${key}=${value}`;
     }, '');
 
   return createSHA256HMAC(API_SECRET_KEY, string, HashFormat.Hex);
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/public/appProxy/__tests__/authenticate.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/public/appProxy/__tests__/authenticate.test.ts
@@ -408,12 +408,12 @@ async function createAppProxyHmac(url: URL): Promise<string> {
   const params: Record<string, string | string[]> = {};
   for (const [key, value] of url.searchParams.entries()) {
     const existing = params[key];
-    if (existing !== undefined) {
+    if (existing === undefined) {
+      params[key] = value;
+    } else {
       params[key] = Array.isArray(existing)
         ? [...existing, value]
         : [existing, value];
-    } else {
-      params[key] = value;
     }
   }
   const string = Object.entries(params)
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/public/appProxy/__tests__/authenticate.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/public/appProxy/__tests__/authenticate.test.ts
@@ -96,6 +96,26 @@ describe('authenticating app proxy requests', () => {
     expect(response.statusText).toBe('Bad Request');
   });
 
+  it('Succeeds when a query parameter appears more than once', async () => {
+    // GIVEN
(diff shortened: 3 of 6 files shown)


byrichardpowell:
Marking as WIP as I need to test this

Suleimanlatrsh:
Approving once the WIP tag comes off and the remaining task is confirmed. Verified locally: both remix and react-router appProxy suites pass, and reverting the fix fails the duplicate-param test in each.

Minor, non-blocking: the new test only asserts `liquid` is defined. It does catch the regression (auth throws without the fix), but rendering the liquid and asserting the 200 response would match the pattern used by the other valid-request tests in the file.

Suleimanlatrsh (approved):

Suleimanlatrsh:
Update: verified end to end with real Shopify traffic. Built this branch's fixed build, ran it inside the React Router template on a live dev store, set up an App Proxy, and hit the storefront so Shopify signed and forwarded a real request with duplicate params. The fixed build returned 200 (authenticated) where the shipped build returned 400. This is the exact in-template real-request test, and it passes. Still just needs the [WIP] tag dropped.

Suleimanlatrsh:
Final status: approved, green, verified against real Shopify traffic in the React Router template (real signed App Proxy duplicate-param request accepted). Ready to merge once the [WIP] tag is dropped.

Suleimanlatrsh:
Closing this as obsolete.

The duplicate-param hardening already landed on main via normalizeQuery in shopify-api (packages/apps/shopify-api/lib/utils/hmac-validator.ts). It comma-joins repeated params the way Shopify signs them and rejects repeated hmac/shop/signature/timestamp.

This branch predates that, so rebasing it as-is would be a regression rather than a fix: searchParamsToQuery returns a plain object, which normalizeQuery passes through untouched, so the repeated-security-param rejection would stop firing. The green CI here is running against a stale base, which is why it looks safe.

Worth keeping in a small follow-up: the positive test that a repeated application param (the consentGiven case from #2374) still authenticates. main only covers the rejection side. Also worth a separate look: setup-valid-request.ts still uses Object.fromEntries(url.searchParams.entries()), so the published test helper cannot sign a duplicate-param request.

PR #3102, opened 2026-03-13
Fix HMAC verification failure when query params contain spaces

### LINKED ISSUE #2701: HMAC mismatch when query contains spaces — URLSearchParams encodes spaces differently than Shopify expects
# Issue summary

  - Node version:
  - Operating system:
- [x] Set `{ logger: { level: LogSeverity.Debug } }` in my configuration, when applicable
- [x] Checked that this isn't an issue with a Shopify API

When a query parameter value contains spaces, the query string generated by JavaScript's new URLSearchParams(...) in the Shopify SDK uses an encoding that doesn't match what Shopify's signature/verification expects, causing the HMAC verification to fail.

Image

## Expected behavior

validated by `api.utils.validateHmac(query)` work.

## Actual behavior

not work.  I'm a bit confused — how exactly should the HMAC check be performed here?  After reading the code of `api.utils.validateHmac`, I tried a version that doesn't encode spaces, and the HMAC verification passed.

1. Open [Shopify App Store],  find any App.
2. add some params with escapes at query.
3. click `Install`
4. copy the searchParams, try validate it.

byrichardpowell:

Fixes #2701

`URLSearchParams.toString()` encodes spaces as `+` (application/x-www-form-urlencoded format), but Shopify generates its HMAC signature using percent-encoding (`%20`). This mismatch caused `api.utils.validateHmac()` to return `false` for any valid OAuth request where a query parameter value contained a space.

- In `ProcessedQuery.stringify()`, replaces `+` with `%20` after calling `URLSearchParams.toString()` so the canonical query string matches Shopify's signing format.
- Updates existing test expectations from `+` to `%20` in hardcoded expected query strings.
- Adds a regression test: `'spaces in param values are percent-encoded (%20), not plus-encoded'`.

## Type of change

## Checklist


byrichardpowell:
Marking as WIP as I need to test this

Suleimanlatrsh:
Approving once the WIP tag comes off. Verified locally: hmac, processed-query, and oauth suites pass, and reverting the `%20` change fails 6 tests.

I checked the two callers of `ProcessedQuery.stringify()`. Besides the HMAC path, the only other caller is the OAuth authorize redirect URL, and the oauth suite stays green, so the shared change is safe there too.

Suleimanlatrsh (approved):

Suleimanlatrsh:
Update: no real-traffic test is possible here. The fix only changes behavior when a signed param value contains a literal space, and Shopify never sends space-containing params in signed hmac callbacks (code, hmac, host, shop, state, timestamp). It's a defensive fix, covered by unit tests plus a revert check (space case fails without the fix, passes with it). Just needs the [WIP] tag dropped.

Suleimanlatrsh:
Final status: approved, green. No real-traffic test is possible (Shopify never sends space-containing params in signed hmac requests); covered by unit tests + a revert check. Ready to merge once the [WIP] tag is dropped.

PR #3106, opened 2026-03-13
Fix: Add cookiePath config option for multi-shop non-embedded apps

### LINKED ISSUE #451: Session cookie race condition with multiple shops
# Issue summary

In our standalone Remix app (not embedded), we use `await authenticate.admin(request)` to create a session (offline). This sets a cookie and the user's session has the auth token for the shop they accessed our app from.

However, if the same user accesses our app from another shop, the cookie retrieves a session for the previous shop -- thereby referencing the wrong session information and access token. The cookie also still states `cookies: 'shopify_app_session=offline_previous-shop.myshopify.com;`.

This same thing happens with online sessions. The `authenticate.admin` function doesn't ensure that the `shop` querystring parameter is used for session storage and retrieval, to handle the same user accessing the same app from multiple shops.

Is there a way to prevent a cookie from being set, to make these requests stateless?

- @shopify/app@3.49.3
- @shopify/cli@3.49.3
- @shopify/shopify-app-remix@1.2.1
- @shopify/shopify-app-session-storage-dynamodb@1.0.7
- Node version: v18.18.0
- Operating system: macOS 13.4.1 (Ventura)

```
15:29:09 │ remix    │ LOADER (ROOT) http://app.test/?hmac=4b1e0c2f88131f91000f87f592a2f22475000cd073ed3d8712d4478b880722b6&host=YWRtaW4oo2hvcGlmeS5jb20oo3RvcmUvb2R4LWxvY2FsLWVhdHM&shop=second-store.myshopify.com&timestamp=123456789 {
15:29:09 │ remix    │   cookies: 'shopify_app_session=offline_previous-store.myshopify.com; shopify_app_session.sig=jFJzqzEB9g/WxxVarL///iFQaOpbtdj6n2pMS8t1iFk='
15:29:09 │ remix    │ }
15:29:09 │ remix    │ LOADER (_index) http://app.test/?hmac=4b1e0c2f88131f91000f87f592a2f22475000cd073ed3d8712d4478b880722b6&host=YWRtaW4oo2hvcGlmeS5jb20oo3RvcmUvb2R4LWxvY2FsLWVhdHM&shop=second-store.myshopify.com&timestamp=123456789
15:29:09 │ remix    │ [shopify-app/INFO] Authenticating admin request
15:29:09 │ remix    │ SESSION Session {
15:29:09 │ remix    │   isOnline: false,
15:29:09 │ remix    │   shop: 'previous-store.myshopify.com',
15:29:09 │ remix    │   scope: 'read_all_orders,read_inventory,read_price_rules,read_product_listings,write_customers,write_draft_orders,write_files,write_fulfillments,write_merchant_managed_fulfillment_orders,write_order_edits,write_orders,write_products,write_script_tags,write_shipping,write_themes,write_third_party_fulfillment_orders',
15:29:09 │ remix    │   id: 'offline_previous-store.myshopify.com',
15:29:09 │ remix    │   state: 'XXXXXXXXXX',
15:29:09 │ remix    │   accessToken: 'shpca_XXXXXXXXXXXX',
15:29:09 │ remix    │   expires: undefined
15:29:09 │ remix    │ }
```

## Expected behavior

1. Users should be able to access the same app from different shops
2. `authenticate.admin` should retrieve session data based on the shop the request originated from (e.g. respect the `shop` querystring parameter)

## Actual behavior

1. Users accessing the app from multiple shops resumes the initial session and ignores that links have `hmac` and `shop` parameters indicating they are accessing from a different shop

1. Create Remix app from template
2. Call `const { session } = await authenticate.admin(request)` to initiate a session when accessing the app homepage
3. Log or output the `session` object to see what shop it holds information for
4. Install the app on Shop 1 and see that `session` has Shop 1 data
5. Install the app on Shop 2 and see that `session` still has Shop 1 data

byrichardpowell:

Fixes https://github.com/Shopify/shopify-app-js/issues/451
Related: https://github.com/Shopify/shopify-app-template-react-router/issues/156

Non-embedded apps that support multiple shops simultaneously in separate browser tabs hit a cookie collision: all shops share a single `shopify_app_session` cookie scoped to `path=/`, so each OAuth callback silently overwrites the previous shop's session. The user in Tab A then unknowingly operates on Tab B's shop after Tab B completes auth.

This is a known limitation acknowledged by `paulomarg` in [issue #451](https://github.com/Shopify/shopify-app-js/issues/451) (Sept 2023) and has since resurfaced in the React Router template repo.

Adds an optional `cookiePath` config option to `ConfigParams` (in `@shopify/shopify-api`) that controls the `path` attribute written on the `shopify_app_session` cookie during OAuth callback.

**Before:** hardcoded `path: '/'` — domain-wide, one cookie shared across all shops.

**After:** configurable via a static string or a per-session factory function:

```ts
// Default — unchanged behaviour
shopifyApi({ cookiePath: '/' })

// Multi-shop apps: scope cookie to a shop-specific URL prefix
shopifyApi({
  cookiePath: (session) => `/shops/${session.shop}/`,
})
```

When each shop is served under a distinct URL prefix, the browser maintains one cookie per shop and delivers only the matching cookie per request — no collision.

**Changes:**
- `lib/base-types.ts` — adds `cookiePath?: string | ((session: Session) => string)` to `ConfigParams` with full JSDoc
- `lib/auth/oauth/oauth.ts` — resolves the configured path at callback time and passes it to `setAndSign`
- Tests — three new test cases: default path `/`, static string path, factory function path
- Changeset — `minor` bump for `@shopify/shopify-api`

**Note:** this fix requires apps to structure their routing with shop-scoped URL prefixes. The library cannot derive the path automatically (see investigation notes on the chicken-and-egg read-time problem). The JSDoc explains this requirement.

## Type of change

## Checklist

### DIFF
--- a/.changeset/cookie-path-config.md
+++ b/.changeset/cookie-path-config.md
@@ -1,0 +1,26 @@
+---
+'@shopify/shopify-api': minor
+---
+
+Add `cookiePath` config option for multi-shop non-embedded apps
+
+Non-embedded apps that need to support multiple shops simultaneously in
+separate browser tabs were affected by a cookie collision: all shops shared
+a single `shopify_app_session` cookie at `path=/`, so authenticating a new
+shop would silently overwrite the previous shop's session.
+
+The new optional `cookiePath` config option lets you scope the session
+cookie to a shop-specific URL prefix, so each shop's cookie coexists
+independently in the browser.
+
+```ts
+// Static path (default behaviour, unchanged)
+cookiePath: '/'
+
+// Factory function — recommended for multi-shop apps
+cookiePath: (session) => `/shops/${session.shop}/`
+```
+
+**Requirement:** the configured path must match your app's URL structure.
+Each shop must be served under a distinct URL prefix for the browser to
+deliver the correct cookie per request.
--- a/packages/apps/shopify-api/lib/auth/oauth/__tests__/oauth.test.ts
+++ b/packages/apps/shopify-api/lib/auth/oauth/__tests__/oauth.test.ts
@@ -721,6 +721,126 @@ describe('callback', () => {
       responseCookies.shopify_app_session.expires?.getTime(),
     ).toBeUndefined();
   });
+
+  test('uses default path "/" for session cookie when cookiePath is not configured', async () => {
+    const shopify = shopifyApi(testConfig({isEmbeddedApp: false}));
+
+    const beginResponse: NormalizedResponse = await shopify.auth.begin({
+      shop,
+      isOnline: false,
+      callbackPath: '/some-callback',
+      rawRequest: request,
+    });
+    setCallbackCookieFromResponse(
+      request,
+      beginResponse,
+      shopify.config.apiSecretKey,
+    );
+
+    const testCallbackQuery: QueryMock = {
+      shop,
+      state: VALID_NONCE,
+      timestamp: getCurrentTimeInSec().toString(),
+      code: 'some random auth code',
+    };
+    const expectedHmac = await generateLocalHmac(shopify.config)(
+      testCallbackQuery,
+    );
+    testCallbackQuery.hmac = expectedHmac;
+    request.url += `?${new URLSearchParams(testCallbackQuery).toString()}`;
+
+    queueMockResponse(JSON.stringify({access_token: 'token', scope: ''}));
+
+    const callbackResponse = await shopify.auth.callback({rawRequest: request});
+    const responseCookies = Cookies.parseCookies(
+      callbackResponse.headers['Set-Cookie'],
+    );
+
+    expect(responseCookies.shopify_app_session.path).toEqual('/');
+  });
+
+  test('uses static cookiePath string for session cookie', async () => {
+    const shopify = shopifyApi(
+      testConfig({isEmbeddedApp: false, cookiePath: '/my-app/'}),
+    );
+
+    const beginResponse: NormalizedResponse = await shopify.auth.begin({
+      shop,
+      isOnline: false,
+      callbackPath: '/some-callback',
+      rawRequest: request,
+    });
+    setCallbackCookieFromResponse(
+      request,
+      beginResponse,
+      shopify.config.apiSecretKey,
+    );
+
+    const testCallbackQuery: QueryMock = {
+      shop,
+      state: VALID_NONCE,
+      timestamp: getCurrentTimeInSec().toString(),
+      code: 'some random auth code',
+    };
+    const expectedHmac = await generateLocalHmac(shopify.config)(
+      testCallbackQuery,
+    );
+    testCallbackQuery.hmac = expectedHmac;
+    request.url += `?${new URLSearchParams(testCallbackQuery).toString()}`;
+
+    queueMockResponse(JSON.stringify({access_token: 'token', scope: ''}));
+
+    const callbackResponse = await shopify.auth.callback({rawRequest: request});
+    const responseCookies = Cookies.parseCookies(
+      callbackResponse.headers['Set-Cookie'],
+    );
+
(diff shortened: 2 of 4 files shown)


Suleimanlatrsh:
Verified locally: the oauth suite passes, reverting the change fails the static and factory cookiePath tests, and eslint runs clean on the changed files. The code is ready.

The only thing blocking is stale CI. This branch is behind main, and the failing lint plus the 6-hour cancelled session-storage job are from March. Please rebase on main and re-run CI. Once it's green I'll approve.

Suleimanlatrsh:
Updated the branch onto main and pushed a one-line fix: the Lint failure was a real prettier error in `oauth.test.ts` (the `toEqual(`/shops/${shop}/`)` call was wrapped across lines). Not stale after all, my earlier local check only linted the source files and missed the test file. Ran prettier's autofix, package lint is now clean and the oauth suite still passes 60/60. Waiting on the fresh CI, will approve once green.

Suleimanlatrsh (approved):
Approving. Branch is now up to date with main, the lint fix landed, and all required checks are green. Verified locally too: oauth suite passes 60/60 and reverting the fix fails the static and factory cookiePath tests, so the coverage is regression-proof.

Suleimanlatrsh:
Update: not run against real Shopify traffic. Doing so needs a non-embedded Express harness plus a real OAuth round trip, which is heavier than warranted for a deterministic config option. Coverage is unit tests plus a revert check: the static and factory cookiePath cases fail without the fix and pass with it. Code and tests verified locally; just needs a rebase/Update branch before merge.

Suleimanlatrsh:
Real end-to-end test in shopify-app-template-node (non-embedded). Configured the template with isEmbeddedApp:false and cookiePath:(session)=>`/shops/${session.shop}/`, then drove a real OAuth begin->callback (token exchange stubbed, HMAC signed as Shopify would).

Result, inspecting the shopify_app_session Set-Cookie on the callback response:
- Shipped: Path=/  (all shops share one cookie -> the collision in #451)
- With this PR: Path=/shops/test-shop.myshopify.com/  (shop-scoped, so multiple shops coexist)

So the original multi-shop cookie collision is reproduced on the shipped build and resolved by this PR. Same caveat as #3100: I applied the PR's exact change to the shipped compiled shopify-api in the template (branch build tooling was broken / major-version mismatch), so it's the real fix logic at the template's version rather than a clean published build.

Suleimanlatrsh:
Update: verified against REAL Shopify traffic in shopify-app-template-node. Ran the template non-embedded with cookiePath:(session)=>`/shops/${session.shop}/`, did an actual OAuth install on a dev store (real authorize + real token exchange, app shows as Installed).

The session cookie set by the real callback:
  shopify_app_session=...; sameSite=lax; secure=true; path=/shops/suli-store-scd2lkdu.myshopify.com/

So the fix scopes the cookie to the shop-specific path under real OAuth (shipped uses path=/). Confirmed end to end, not simulated.

Note: after install, Shopify showed oauth_error=same_site_cookies / 'app couldn't load'. That's expected and actually confirms the scoping: the cookie is now at /shops/<shop>/, but the template still serves its UI from /, so the browser doesn't send it back for the app routes. This matches the PR's documented requirement that the app must be served under a per-shop URL prefix for cookiePath to be useful. Not a defect in the fix.

(Same disclosure as before: I applied the PR's exact change to the template's installed shopify-api build by hand due to a broken branch build + major-version mismatch, so it's the real fix logic at the template version.)

Suleimanlatrsh:
Final status: approved, green, verified against real Shopify traffic (real OAuth install; session cookie set to path=/shops/<shop>/ vs shipped path=/). Ready to merge after a rebase/Update branch.

PR #3157, opened 2026-04-14
fix: add allowed_tools to investigate workflow to unblock CI execution

byrichardpowell:
## Summary

The `devtools-investigate-for-gardener` workflow was missing an `allowed_tools` configuration on the `claude-code-action` step. In CI (non-interactive), Claude Code's permission system blocks all network-accessing and file-modifying commands unless explicitly allowed. This caused the investigation to fail completely — the agent couldn't even fetch the issue it was supposed to investigate.

**Evidence:** [Action run #24408641456](https://github.com/Shopify/shopify-app-js/actions/runs/24408641456) completed with 28 permission denials and produced [this report](https://github.com/Shopify/shopify-app-js/actions/runs/24408641456#:~:text=Issue%20Overview) stating _"This investigation could not be completed because all network-accessing tools (gh, curl, python3 urllib, node https) were blocked by Claude Code's permission system."_

### Changes

- **Workflow** (`.github/workflows/devtools-investigate-for-gardener.yml`):
  - Add `allowed_tools` with permissions for `gh` CLI (issues, PRs, releases), git operations, and file tools (Read, Glob, Grep, Edit, Write)
  - Invoke the skill via `/investigating-github-issues` slash command instead of telling Claude to read the SKILL.md file directly — this enables the skill's frontmatter `allowed-tools` to take effect through the skill system
  - Simplify the prompt to pass the issue URL as the skill argument, removing redundant instructions already in the skill
  - Add fallback instruction to read SKILL.md directly if the slash command doesn't load the skill (unclear if `claude-code-action` fully supports slash command invocation in prompts)

- **Skill** (`.claude/skills/investigating-github-issues/SKILL.md`):
  - Add `allowed-tools` to the YAML frontmatter as machine-readable configuration for the skill system
  - Add note explaining why `pnpm` and `npx` are excluded

### Security considerations

- `pnpm` and `npx` are **intentionally excluded** to prevent arbitrary code execution via prompt injection from untrusted issue content. Changesets can be created by writing markdown files directly to `.changeset/` instead.
- `git push` is allowed but repo-level branch protections prevent direct pushes to `main` — the agent can only push to new branches and open PRs for review.

## Test plan

- [ ] Re-run the workflow on an issue with the `devtools-investigate-for-gardener` label and verify the agent can fetch issue details and complete the investigation
- [ ] Verify the agent is not prompted for permission on `gh issue view`, `gh issue list`, etc.

### DIFF
--- a/.claude/skills/investigating-github-issues/SKILL.md
+++ b/.claude/skills/investigating-github-issues/SKILL.md
@@ -1,23 +1,36 @@
 ---
 name: investigating-github-issues
 description: Investigates and analyzes GitHub issues for Shopify/shopify-app-js. Fetches issue details via gh CLI, searches for duplicates, examines the codebase for relevant context, applies version-based maintenance policy classification, and produces a structured investigation report. Use when a GitHub issue URL is provided, when asked to analyze or triage an issue, or when understanding issue context before starting work.
+allowed-tools:
+  - Bash(gh issue view *)
+  - Bash(gh issue list *)
+  - Bash(gh pr list *)
+  - Bash(gh pr view *)
+  - Bash(gh pr create *)
+  - Bash(gh pr checks *)
+  - Bash(gh pr diff *)
+  - Bash(gh release list *)
+  - Bash(git log *)
+  - Bash(git tag *)
+  - Bash(git diff *)
+  - Bash(git show *)
+  - Bash(git branch *)
+  - Bash(git checkout *)
+  - Bash(git push *)
+  - Bash(git commit *)
+  - Bash(git add *)
+  - Read
+  - Glob
+  - Grep
+  - Edit
+  - Write
 ---
 
 # Investigating GitHub Issues
 
 Use the GitHub CLI (`gh`) for all GitHub interactions — fetching issues, searching, listing PRs, etc. Direct URL fetching may not work reliably.
 
-## Required Tools (CI / Automated Contexts)
-
-When running this skill in CI (e.g., `claude-code-action`), the following `allowed_tools` must be configured. Without them, all network-accessing and file-modifying commands will be blocked by the permission system.
-
-```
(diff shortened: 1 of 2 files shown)


.claude/skills/investigating-github-issues/SKILL.md
```diff
@@ -1,12 +1,37 @@
...
+  - Bash(gh pr checks *)
+  - Bash(gh pr diff *)
+  - Bash(gh release list *)
+  - Bash(git log *)
+  - Bash(git tag *)
+  - Bash(git diff *)
+  - Bash(git show *)
+  - Bash(git branch *)
+  - Bash(git checkout *)
+  - Bash(git push *)
```
lizkenyon: Could we tighten this to `Bash(git push -u origin *)` ? The `*` wildcard here would allow `git push --force`, which could rewrite history on unprotected branches. Scoping to `-u origin` covers the intended use case (pushing a new fix branch) without opening that door.

Same change would need to go in the workflow `allowed_tools` string to stay in sync.

.claude/skills/investigating-github-issues/SKILL.md
```diff
@@ -1,12 +1,37 @@
...
+  - Bash(gh pr create *)
+  - Bash(gh pr checks *)
+  - Bash(gh pr diff *)
+  - Bash(gh release list *)
+  - Bash(git log *)
+  - Bash(git tag *)
+  - Bash(git diff *)
+  - Bash(git show *)
+  - Bash(git branch *)
+  - Bash(git checkout *)
```
lizkenyon: Could we scope this to `Bash(git checkout -b *)` ? The broad pattern lets the agent checkout any existing branch — combined with `Edit`/`Write` + `git push`, it could modify someone else's in-progress branch. Restricting to `-b` limits it to creating new branches, which is all the fix-PR flow needs.

Same change in the workflow file too.

lizkenyon (approved):
Let's some comments. But approving so no re-review necessary

.claude/skills/investigating-github-issues/SKILL.md
byrichardpowell: Yeah, good catch.

.claude/skills/investigating-github-issues/SKILL.md
byrichardpowell: Another good catch.  Updated 👍

PR #3158, opened 2026-04-14
Document webhook reconciliation behavior in shopify-app-express

joshwheeler:
The README does not document that when `auth.callback()` completes an offline OAuth session, it automatically calls `register()`, which deletes any webhook subscriptions not declared in `webhookHandlers`.

Added a Webhook registration section explaining this behavior and a migration warning for apps moving from custom implementations.


byrichardpowell:
Thanks @joshwheeler 

I've pushed some updates:

1. Made it clear [app specific webhooks](https://shopify.dev/docs/apps/build/webhooks/subscribe#app-specific-vs-shop-specific-subscriptions) are preferable.
2. Made it clear which API's apply to shop vs app specific webhooks
3. Added proper changelog entry & version bump so the README gets picked up on npm.

If yu are comfortable with these changes please merge 👍

byrichardpowell (approved):

PR #3182, opened 2026-04-22
Fix cross-origin parameter propagation in admin redirect helper

lizkenyon:
## Summary

- The admin `redirect` helper decided whether to copy request parameters onto the destination using `url.startsWith('/')`. That lexical check disagrees with how `new URL(url, base)` actually resolves URLs — protocol-relative (`//attacker.com/x`) and backslash-prefixed (`/\attacker.com/x`) inputs start with `/` but resolve cross-origin, so request parameters (including `id_token`, `shop`, `host`, and `embedded`) were being copied to non-same-origin destinations.
- Swapped the lexical check for an origin compare: `parsedUrl.origin === new URL(config.appUrl).origin`. The origin compare alone is sufficient to close the cross-origin parameter leak — any URL that resolves cross-origin fails this check regardless of its lexical shape.
- The `new URL(config.appUrl).origin` normalization on the right side is a separate, *functional* fix rather than a second security layer. Without it, same-origin detection silently breaks for apps configured with a trailing slash or path on `appUrl` (e.g. `https://my-app.example/`) — which is what the old `url.startsWith('/')` branch was quietly compensating for.
- Tightened the JSDoc on `redirect` to call out that parameters are only copied to same-origin destinations and that callers shouldn't pass unvalidated user input.

Applies to both `@shopify/shopify-app-react-router` and `@shopify/shopify-app-remix` — same fix, identical diff in each.

## Test plan

- [ ] Redirect tests pass in both packages: `pnpm jest --testPathPatterns='redirect.test.ts'`
- [ ] Full `authenticate/admin` suite passes in both packages (116 + 404 tests)
- [ ] Three new tests per package cover `//attacker`, `/\attacker`, and the trailing-slash-`appUrl` regression
- [ ] Pre-existing typecheck errors are unchanged (verified against `main`)


byrichardpowell (approved):
🎉

PR #3216, opened 2026-05-15
Render App Bridge instead of login redirect when shop/host params are missing in embedded apps

byrichardpowell:
## Problem

During local development of embedded Shopify apps, there's an annoying workflow where the login screen appears inappropriately:

1. Navigate to a new URL within the app (the `id_token` / `shop` / `host` query params are lost from the URL)
2. Make a code change
3. Vite triggers a full page reload (this happens intermittently — not all changes trigger HMR, some cause a full reload)
4. The server receives a document request with no `shop` or `host` params and no `Authorization` header
5. `validateShopAndHostParams` redirects to the login page

This is incorrect because the app is still running inside the Shopify admin iframe — it just lost the URL search params during SPA navigation + reload.

## Solution

In `validateShopAndHostParams`, when `shop` or `host` params are missing/invalid, we now render a minimal App Bridge page instead of redirecting to login.

When App Bridge loads inside the Shopify admin iframe, it:
1. Detects it's embedded in the admin
2. Retrieves the session token from the parent frame
3. Re-authenticates the app seamlessly

This is the same mechanism used by the existing "bounce page" flow (`/auth/session-token`), just applied earlier in the auth pipeline.

The previous code had a `throw redirect(config.auth.loginPath)` fallback for non-embedded apps, but `isEmbeddedApp` is hardcoded to `true` in `shopify-app.ts` for all apps using this library (the `ShopifyAdmin` distribution is already excluded earlier in the function). This made the login redirect dead code. This PR removes that dead code — no public APIs are changed.

Note: this is a production behavior change, not just a dev improvement. Any request to an embedded app without `shop`/`host` params (e.g., bookmarked deep links, direct URL access) will now get the App Bridge page instead of a login redirect. This is correct behavior — App Bridge will handle re-embedding in the admin.

### What changed

- **`validate-shop-and-host-params.ts`**: Replaced the `redirectToLoginPath` helper with `renderAppBridgeOrError`. Instead of redirecting to the login page, renders the App Bridge page so it can recover the session. Removed the unused `redirect` import from `react-router` and the dead login redirect code path. The login path guard (500 error when `authenticate.admin()` is called from the login route) is preserved.
- **`doc-request-path.test.ts`**: Updated the 4 test cases for missing/invalid shop/host params to expect an App Bridge page (200 with App Bridge script tag) instead of a login redirect (302).
- **`reject-bot-request.test.ts`**: Updated the "allowed Shopify agents" assertions to expect 200 (App Bridge page) instead of 302 (login redirect), confirming the request passes the bot check and proceeds to auth validation.
- **`expect-login-redirect.ts`**: Removed — no longer used by any test.
- **`__test-helpers/index.ts`**: Removed the `expect-login-redirect` export.

### DIFF
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/admin/helpers/validate-shop-and-host-params.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/admin/helpers/validate-shop-and-host-params.ts
@@ -1,6 +1,6 @@
-import {redirect} from 'react-router';
-
 import {BasicParams, AppDistribution} from '../../../types';
+
+import {renderAppBridge} from './render-app-bridge';
 
 export function validateShopAndHostParams(
   params: BasicParams,
@@ -12,24 +12,24 @@ export function validateShopAndHostParams(
     const url = new URL(request.url);
     const shop = api.utils.sanitizeShop(url.searchParams.get('shop')!);
     if (!shop) {
-      logger.debug('Missing or invalid shop, redirecting to login path', {
+      logger.debug('Missing or invalid shop, rendering App Bridge', {
         shop,
       });
-      throw redirectToLoginPath(request, params);
+      throw renderAppBridgeOrError(request, params);
     }
 
     const host = api.utils.sanitizeHost(url.searchParams.get('host')!);
     if (!host) {
-      logger.debug('Invalid host, redirecting to login path', {
+      logger.debug('Invalid host, rendering App Bridge', {
         shop,
         host: url.searchParams.get('host'),
       });
-      throw redirectToLoginPath(request, params);
+      throw renderAppBridgeOrError(request, params);
     }
   }
 }
 
-function redirectToLoginPath(request: Request, params: BasicParams): never {
+function renderAppBridgeOrError(request: Request, params: BasicParams): never {
   const {config, logger} = params;
 
   const {pathname} = new URL(request.url);
@@ -42,5 +42,8 @@ function redirectToLoginPath(request: Request, params: BasicParams): never {
     throw new Response(message, {status: 500});
   }
 
-  throw redirect(config.auth.loginPath);
+  logger.debug(
+    'Missing shop or host params, rendering App Bridge to retrieve session',
+  );
+  throw renderAppBridge(params, request);
 }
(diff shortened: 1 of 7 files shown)


packages/apps/shopify-app-react-router/src/server/authenticate/admin/helpers/validate-shop-and-host-params.ts
```diff
@@ -12,24 +12,24 @@ export function validateShopAndHostParams(
     const url = new URL(request.url);
     const shop = api.utils.sanitizeShop(url.searchParams.get('shop')!);
     if (!shop) {
-      logger.debug('Missing or invalid shop, redirecting to login path', {
+      logger.debug('Missing or invalid shop, rendering App Bridge', {
         shop,
       });
-      throw redirectToLoginPath(request, params);
+      throw renderAppBridgeOrError(request, params);
```
lizkenyon, line 18: Should we distinguish shop missing from invalid here?

For example I believe further down the call stack with this flow we will set the CSP headers with that invalid shop. 
> renderAppBridge() re-reads the raw query param and passes it into header construction

lizkenyon (approved):

packages/apps/shopify-app-react-router/src/server/authenticate/admin/helpers/validate-shop-and-host-params.ts
byrichardpowell: Good catch — this was a real (small) hole. Fixed in 8c1776558.

**Fix:** sanitized the shop param inside `renderAppBridge` itself before passing it to `addDocumentResponseHeaders`, rather than branching missing-vs-invalid in the validator. This way every caller of `renderAppBridge` benefits, and it mirrors the existing pattern in `addDocumentResponseHeadersFactory`:

```ts
const shop = api.utils.sanitizeShop(
  new URL(request.url).searchParams.get('shop')!,
);
addDocumentResponseHeaders(responseHeaders, isEmbeddedApp, shop);
```

**Behavior matrix:**

| Shop input | Before | After |
|---|---|---|
| Valid (`*.myshopify.com`, etc.) | header set with shop | header set with shop *(unchanged)* |
| Invalid (e.g. `evil.com`) | **leaked into CSP `frame-ancestors` + Link preconnect** | no header set |
| Missing / null | no header set | no header set *(unchanged)* |

**Regression test** added in `doc-request-path.test.ts` that hits the endpoint with `?shop=evil.com` and asserts:
- CSP, if present, does not contain the attacker value.
- `Link` preconnect header is absent.

Fails against the old code, passes against the new code. Full package test suite (349 tests) green, `tsc` clean.

PR #3237, opened 2026-05-29
Fix admin extension CORS preflight in `validateAuthenticatedSession`

### LINKED ISSUE #1420: Admin extensions preflight authentication
# Issue summary
According to the [docs](https://shopify.dev/docs/api/admin-extensions#app-authentication), I should be able to make authenticated api calls the extension (which runs in the frontend) to my app's backend (which runs on the same domain as app_url).
The app uses the Express.js template

I have an admin action extension that makes the following API call:

Image

The route is guarded by the validateAuthenticatedSession middleware and allowed the cors configurations as the [docs](https://shopify.dev/docs/api/admin-extensions#security) requires.

Image

  - `@shopify/*` package and version: @shopify/shopify-app-express version 5.0.3
  - Node version: 20.17.0
  - Operating system: MacOS

## Expected behavior

The fetch request should pass the authentication.

## Actual behavior

The request's preflight redirects to /api/auth (meaning it didn't authenticated currectly) which fails because of cors (but would fail regardless)

This is the preflight

Image

This is the "real" request which fails because of cors (due to the preflight failing). As you can see, the header is there, but it can't pass the validateAuthenticatedSession middleware

Image

ryandiginomad:
### What

Admin UI extension `fetch` calls to an app's backend (Express template) fail because the CORS preflight `OPTIONS` request is run through `validateAuthenticatedSession`. The preflight carries no `Authorization` header, so the middleware redirects/403s it, and the browser then blocks the real (authenticated) request on CORS.

A maintainer confirmed the fix direction in #1420:

> we're not handling `OPTIONS` requests properly in that package - we should be just setting the CORS headers and responding in that case, instead of trying to authenticate it.

### How

`validateAuthenticatedSession` now short-circuits `OPTIONS` requests before the auth logic runs: it responds `204` with the CORS headers for `https://extensions.shopifycdn.com` — the origin that Shopify's [admin-extensions security docs](https://shopify.dev/docs/api/admin-extensions#security) require apps to allow. The short-circuit is config-agnostic, so it applies to both embedded and non-embedded apps.

### Before / after

Before: `OPTIONS /api/...` → `302` redirect to `/api/auth` (or `403`) → browser blocks the real request on CORS.
After: `OPTIONS /api/...` → `204` with `Access-Control-Allow-Origin: https://extensions.shopifycdn.com` (+ Allow-Methods / Allow-Headers) → the real request proceeds and is authenticated exactly as before.

### Tests

Added 3 tests to `validate-authenticated-session.test.ts`: preflight returns `204` without calling `getCurrentId` (no auth attempt); allows the `Authorization` header; allows the request methods. Full `shopify-app-express` suite: 88 passing, no regressions. `tsc` + lint clean. Changeset added (patch).

### Open to feedback

- **Allowed origin** — hardcoded to `https://extensions.shopifycdn.com` per the docs. Happy to make it configurable or echo a validated `Origin` if you'd prefer.
- **Allow-Methods / Allow-Headers** — currently `GET, POST, PUT, PATCH, DELETE, OPTIONS` and `Authorization, Content-Type`. Can narrow/widen to match what extensions actually send.
- **Allow-Credentials** — not set (admin extension auth uses the `Authorization` header, not cookies). Let me know if a case needs it.

### AI disclosure

This change was developed with AI assistance (Claude). The approach, code, and tests were reviewed and verified locally by me.

Fixes #1420

### DIFF
--- a/.changeset/fix-admin-extension-cors-preflight.md
+++ b/.changeset/fix-admin-extension-cors-preflight.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-app-express": patch
+---
+
+Respond to CORS preflight (`OPTIONS`) requests in `validateAuthenticatedSession` instead of trying to authenticate them. Admin UI extension `fetch` calls to an app's backend send a preflight request that carries no `Authorization` header, so the middleware would redirect/403 and the browser would block the real request on CORS. The middleware now short-circuits `OPTIONS` requests, responding `204` with the CORS headers for `https://extensions.shopifycdn.com`.
--- a/packages/apps/shopify-app-express/src/middlewares/__tests__/validate-authenticated-session.test.ts
+++ b/packages/apps/shopify-app-express/src/middlewares/__tests__/validate-authenticated-session.test.ts
@@ -225,6 +225,43 @@ describe('validateAuthenticatedSession', () => {
 
       expect((response.error as any).text).toBe('Storage error');
     });
+
+    it('responds to a CORS preflight OPTIONS request without authenticating', async () => {
+      const getCurrentIdSpy = jest.spyOn(shopify.api.session, 'getCurrentId');
+
+      const response = await request(app)
+        .options('/test/shop')
+        .set('Origin', 'https://extensions.shopifycdn.com')
+        .expect(204);
+
+      expect(response.headers['access-control-allow-origin']).toBe(
+        'https://extensions.shopifycdn.com',
+      );
+      expect(getCurrentIdSpy).not.toHaveBeenCalled();
+    });
+
+    it('allows the Authorization header on the preflight response', async () => {
+      const response = await request(app)
+        .options('/test/shop')
+        .set('Origin', 'https://extensions.shopifycdn.com')
+        .expect(204);
+
+      expect(response.headers['access-control-allow-headers']).toContain(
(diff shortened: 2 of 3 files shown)


ryandiginomad:
Quick check-in — this PR has been ready for review for a week. CLA signed, CI green (capture + cla). Ready whenever the team has a window, and happy to address anything that comes up.

ryandiginomad:
Quick check-in — this PR has been quiet on the review queue for a couple weeks. CLA signed, CI green, and it's a small bounded fix for #1420. Ready whenever the team has a window, happy to address anything that comes up.

ryandiginomad:
Following up on this one — it's been open since late May without a review. If this isn't something the team wants to take on, I'm happy to close it, no hard feelings. If it is, I'll rebase it onto the current base and address whatever comes up.

ryandiginomad:
Closing this. It's been open since late May without a review, and the branch has drifted 214 commits behind main, so I'd rather close it cleanly than keep rebasing something that isn't in anyone's queue.

The bug is still there: `validateAuthenticatedSession` runs preflight `OPTIONS` requests through the auth path, so #1420 still reproduces on main. The issue is open and unassigned, and the branch stays on my fork if it's useful to whoever picks it up.

PR #3238, opened 2026-05-29
Fix: respond to CORS preflight OPTIONS requests without authenticating (shopify-app-express)

### LINKED ISSUE #1420: Admin extensions preflight authentication

morgan-coded:

Fixes #1420

The `validateAuthenticatedSession` middleware in `@shopify/shopify-app-express`
runs every request — including CORS preflight `OPTIONS` requests — through the
full authentication flow. A preflight request never carries a session cookie or
`Authorization: Bearer` header, so it always fails authentication and is
redirected to `/auth` (HTTP 302). The browser's CORS preflight then fails on the
redirect, so the real authenticated request from an admin extension to the app's
own backend never fires.

This is the behavior the maintainer confirmed in #1420: for `OPTIONS` we should
just set the CORS headers and respond, instead of trying to authenticate.

- Adds a small `respondToOptionsRequest` helper to the Express package that, for
  `OPTIONS` requests, sets the CORS headers (`Access-Control-Allow-Origin`,
  `Access-Control-Allow-Headers`, `Access-Control-Expose-Headers`,
  `Access-Control-Max-Age`) and responds `204` without authenticating. This
  adapts the preflight handling already used by `shopify-app-remix` and
  `shopify-app-react-router` to the Express middleware path.
- Calls it at the top of `validateAuthenticatedSession`, before any session
  lookup, returning early for preflight requests.
- Adds unit tests covering the `OPTIONS` path: the preflight now returns `204`
  with CORS headers, never invokes session lookup, never redirects, and still
  handles preflights without an `Origin` header.

No behavior change for any non-`OPTIONS` request.

#### How was this verified?
- Targeted `jest` for `validate-authenticated-session.test.ts`: 18/18 tests
  pass, including the hardened preflight coverage.
- Red→green check: with the fix removed, the new tests fail with
  `expected 204, got 302` — reproducing the reported redirect — and pass once
  restored.
- `eslint .` and `tsc --noEmit` for the package: both clean.

## Type of change

## Checklist

- [ ] I have documented new APIs/updated the documentation for modified APIs (for public APIs) — n/a, no public API change

### DIFF
--- a/.changeset/express-options-preflight-cors.md
+++ b/.changeset/express-options-preflight-cors.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-express': patch
+---
+
+Respond to CORS preflight `OPTIONS` requests in `validateAuthenticatedSession` by setting CORS headers and returning `204`, instead of running them through the authentication flow. Preflight requests don't carry a session token, so authenticating them always failed and redirected to auth, breaking authenticated fetches from admin extensions.
--- a/packages/apps/shopify-app-express/src/middlewares/__tests__/validate-authenticated-session.test.ts
+++ b/packages/apps/shopify-app-express/src/middlewares/__tests__/validate-authenticated-session.test.ts
@@ -360,5 +360,32 @@ describe('validateAuthenticatedSession', () => {
         'X-Shopify-Api-Request-Failure-Reauthorize-Url',
       );
     });
+
+    it('still responds 204 with CORS headers when no Origin header is sent', async () => {
+      // Admin extension preflights do not always carry an Origin header; the
+      // handler is unconditional, so it must not depend on Origin being present.
+      const getCurrentIdSpy = jest.spyOn(shopify.api.session, 'getCurrentId');
+
+      const response = await request(app)
+        .options('/test/shop?shop=my-shop.myshopify.io')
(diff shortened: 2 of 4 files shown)


morgan-coded:
I have signed the CLA!

PR #3239, opened 2026-05-29
Fix: default cookies to HttpOnly, matching the CookieData JSDoc

### LINKED ISSUE #3207: OAuth state cookie set without HttpOnly (CookieData JSDoc claims default true)
### Issue summary

The OAuth state cookie set during `auth.begin()` is emitted without the `HttpOnly` attribute, despite the `CookieData.httpOnly` JSDoc in `runtime/http/cookies.ts` documenting that the default is `true`. The same flaw applies to the session cookie set on the OAuth callback for non-embedded apps, and to any other cookie written through `Cookies.set` / `Cookies.setAndSign` without explicitly passing `httpOnly: true`.

Found while triaging an OWASP ZAP baseline scan against an embedded Shopify app, which flagged `Cookie No HttpOnly Flag` on the OAuth-begin redirect.

### Versions affected

- `@shopify/shopify-api`: `13.0.0` (latest at time of writing).
- Reproduces on `main` at SHA `f39d1bc1b77230f66bdd493ac7fd69641413553b`.
- Transitively affects every consumer including `@shopify/shopify-app-remix` (verified at `4.2.0`), `@shopify/shopify-app-express`, and `@shopify/shopify-app-react-router`.

### Reproduction

Any OAuth-begin trigger emits a `Set-Cookie` for `shopify_app_state` (and its `.sig` companion) that lacks `HttpOnly`. For example with a stock `shopify-app-remix` install:

```
$ curl -sI "https://<your-app>/auth?shop=<your-shop>.myshopify.com" | grep -i set-cookie
Set-Cookie: shopify_app_state=<nonce>;sameSite=lax; secure=true; path=/auth/callback;expires=<date>
Set-Cookie: shopify_app_state.sig=<sig>;sameSite=lax; secure=true; path=/auth/callback;expires=<date>
```

Notice: no `HttpOnly` attribute on either cookie.

Source trace:

1. [`packages/apps/shopify-api/lib/auth/oauth/oauth.ts#L95-L100`](https://github.com/Shopify/shopify-app-js/blob/f39d1bc1b77230f66bdd493ac7fd69641413553b/packages/apps/shopify-api/lib/auth/oauth/oauth.ts#L95-L100) calls `cookies.setAndSign(STATE_COOKIE_NAME, state, { expires, sameSite, secure, path })`. No `httpOnly` is passed.
2. [`packages/apps/shopify-api/runtime/http/cookies.ts#L154-L176`](https://github.com/Shopify/shopify-app-js/blob/f39d1bc1b77230f66bdd493ac7fd69641413553b/packages/apps/shopify-api/runtime/http/cookies.ts#L154-L176) (`set` and `setAndSign`) spread `opts` into the cookie jar without applying any default for `httpOnly`.
3. [`packages/apps/shopify-api/runtime/http/cookies.ts#L83-L95`](https://github.com/Shopify/shopify-app-js/blob/f39d1bc1b77230f66bdd493ac7fd69641413553b/packages/apps/shopify-api/runtime/http/cookies.ts#L83-L95) (`encodeCookie`) only emits attributes that are own properties of the cookie data, so the absent `httpOnly` produces a header with no flag.

The same flaw is present on the non-embedded session-cookie write at [`lib/auth/oauth/oauth.ts#L219-L226`](https://github.com/Shopify/shopify-app-js/blob/f39d1bc1b77230f66bdd493ac7fd69641413553b/packages/apps/shopify-api/lib/auth/oauth/oauth.ts#L219-L226).

### Expected behaviour

Either:

1. The state and session cookies are emitted with `HttpOnly`, OR
2. `Cookies.set` / `Cookies.setAndSign` apply `httpOnly: true` by default, matching the existing JSDoc on [`CookieData.httpOnly`](https://github.com/Shopify/shopify-app-js/blob/f39d1bc1b77230f66bdd493ac7fd69641413553b/packages/apps/shopify-api/runtime/http/cookies.ts#L34-L38) which already states "true by default".

### Actual behaviour

Neither location applies the documented default. Cookies are written without `HttpOnly` unless the caller explicitly passes it. No caller in `lib/auth/oauth/oauth.ts` does.

### Proposed fix (smallest possible)

The simplest, surgical fix is one line in `lib/auth/oauth/oauth.ts`:

```ts
await cookies.setAndSign(STATE_COOKIE_NAME, state, {
  expires: new Date(Date.now() + 60000),
  sameSite: 'lax',
  secure: true,
  httpOnly: true,
  path: callbackPath,
});
```

A more thorough fix that also makes the JSDoc accurate and protects every other call site (including the non-embedded session cookie at line 220) is to apply the default in `Cookies.set`:

```ts
set(name: string, value: string, opts: Partial<CookieData> = {}): void {
  this.outgoingCookieJar[name] = {
    httpOnly: true, // matches CookieData.httpOnly JSDoc
    ...opts,
    name,
    value,
  };
  this.updateHeader();
}
```

Either fix is non-breaking: callers can still pass `httpOnly: false` explicitly if they need JS access (none in `lib/auth` do).

### Risk / severity

Low. Defense-in-depth, not a primary vulnerability. The state cookie is short-lived (60s), single-use, path-scoped to the callback, and HMAC-signed. An attacker who can read it via XSS already has bigger primitives on the same origin. But the cookie has zero legitimate need for JS access, and the JSDoc already promises the default is `true`. Fixing this aligns the implementation with the documented contract and with [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#cookies) and [RFC 6265 § 4.1.2.6](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.6) guidance.

Happy to send a PR if the team agrees on the broader (`Cookies.set` default) fix vs. the surgical one.

### Checklist

- [x] Upgraded to the latest version of the relevant packages: `@shopify/shopify-app-remix@4.2.0` (latest), transitively `@shopify/shopify-api@13.0.0` (latest).
- [x] Searched existing issues. Closest are #195 and #2509, neither relevant. No prior issue addresses missing `HttpOnly` on the OAuth state cookie.
- [x] Confirmed not a Shopify API issue. The cookie is set by this library, not by Shopify.

morgan-coded:

Fixes #3207

`CookieData.httpOnly` is documented as "true by default", but `Cookies.set`
never applies that default — it only spreads the caller's options. No caller in
`lib/auth/oauth/oauth.ts` passes `httpOnly`, so the OAuth `shopify_app_state`
cookie (and the non-embedded session cookie) are emitted **without** the
`HttpOnly` flag. An OWASP ZAP baseline scan flags this as "Cookie No HttpOnly
Flag". These cookies have no legitimate need for client-side JavaScript access.

- `Cookies.set` (and therefore `Cookies.setAndSign`) now defaults `httpOnly` to
  `true`, matching the existing `CookieData.httpOnly` JSDoc. The default is
  applied before spreading caller options, so callers can still opt out by
  explicitly passing `httpOnly: false`. This fixes every call site at once,
  including the non-embedded session cookie.
- Extends the existing OAuth begin/callback tests to assert that the state
  cookie, its `.sig` companion, and the non-embedded session cookie all carry
  `HttpOnly`.
- Adds focused coverage for the `Cookies.set` default, explicit opt-out, and
  preservation of other caller options.

No behavior change other than the added flag; existing callers that relied on
the absence of `HttpOnly` can pass `httpOnly: false`.

#### How was this verified?
- Targeted `jest --selectProjects library --testPathPatterns "oauth\\.test"` in
  `packages/apps/shopify-api`: 23/23 pass.
- Red→green: with the change reverted, the new assertions fail
  (`httpOnly` is `undefined`); with it applied they pass.
- `eslint .` and `tsc --noEmit`: both clean.

## Type of change

- [ ] Minor: New feature
- [ ] Major: Breaking change

## Checklist

- [x] I have used `pnpm changeset` to create a draft changelog entry
- [ ] I have documented new APIs/updated the documentation for modified APIs — n/a (no public API surface change; the JSDoc already documented this default)

### DIFF
--- a/.changeset/fix-cookie-httponly-default.md
+++ b/.changeset/fix-cookie-httponly-default.md
@@ -1,0 +1,12 @@
+---
+'@shopify/shopify-api': patch
+---
+
+Set `httpOnly` by default when writing cookies
+
+`Cookies.set` (and therefore `Cookies.setAndSign`) now applies `httpOnly: true`
+by default, matching the long-standing `CookieData.httpOnly` JSDoc which already
+documented `true` as the default. Previously the default was never applied, so
+the OAuth `shopify_app_state` cookie (and the non-embedded session cookie) were
+emitted without the `HttpOnly` flag. Callers that need client-side JavaScript
+access can still opt out by explicitly passing `httpOnly: false`.
--- a/packages/apps/shopify-api/lib/auth/oauth/__tests__/oauth.test.ts
+++ b/packages/apps/shopify-api/lib/auth/oauth/__tests__/oauth.test.ts
@@ -65,6 +65,13 @@ describe('beginAuth', () => {
       expect(cookies.outgoingCookieJar.shopify_app_state.value).toEqual(
         VALID_NONCE,
       );
+      // The state cookie (and its signature) must be HttpOnly so it is not
+      // exposed to client-side JavaScript. See issue #3207. The value is a
+      // string here because it is parsed back out of the Set-Cookie header.
+      expect(cookies.outgoingCookieJar.shopify_app_state.httpOnly).toBe('true');
+      expect(cookies.outgoingCookieJar['shopify_app_state.sig'].httpOnly).toBe(
+        'true',
+      );
     });
   });
 
@@ -323,6 +330,8 @@ describe('callback', () => {
       callbackResponse.headers['Set-Cookie'],
     );
     expect(responseCookies.shopify_app_session.value).toEqual(expectedId);
+    // The non-embedded session cookie must also be HttpOnly. See issue #3207.
+    expect(responseCookies.shopify_app_session.httpOnly).toBe('true');
     expect(callbackResponse.session.accessToken).toBe(
       successResponse.access_token,
     );
--- a/packages/apps/shopify-api/lib/auth/oauth/__tests__/oauth.test.ts
+++ b/packages/apps/shopify-api/lib/auth/oauth/__tests__/oauth.test.ts
@@ -747,3 +747,36 @@ function setCallbackCookieFromResponse(
     `shopify_app_state.sig=${responseCookies.outgoingCookieJar['shopify_app_state.sig'].value}`,
   ].join(';');
 }
+
+describe('Cookies.set httpOnly default (issue #3207)', () => {
+  // These assert the in-memory CookieData directly (a real boolean), which is a
+  // stronger check than the header-parsed string assertions above, and pin the
+  // opt-out path the fix promises in its changeset and PR body.
+  const response = {} as NormalizedResponse;
+
+  it('defaults httpOnly to true when the caller does not pass it', () => {
+    const cookies = new Cookies({} as NormalizedRequest, response, {keys: []});
+
+    cookies.set('test_cookie', 'value');
+
+    expect(cookies.outgoingCookieJar.test_cookie.httpOnly).toBe(true);
+  });
+
+  it('lets a caller opt out with httpOnly: false', () => {
+    const cookies = new Cookies({} as NormalizedRequest, response, {keys: []});
+
+    cookies.set('test_cookie', 'value', {httpOnly: false});
+
+    expect(cookies.outgoingCookieJar.test_cookie.httpOnly).toBe(false);
+  });
+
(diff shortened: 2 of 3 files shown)


morgan-coded:
I have signed the CLA!

PR #3240, opened 2026-05-29
Add apiSecretKeyFallback to validate inbound HMACs during client-secret rotation

### LINKED ISSUE #3183: Webhooks dropped for ~30 min post-revocation: @shopify/shopify-api has no way to accept a second (rotating) secret
## Summary

Shopify's [client-secret rotation docs](https://shopify.dev/docs/apps/build/authentication-authorization/client-secrets/rotate-revoke-client-credentials#step-2-configure-webhooks) describe webhooks being signed with "the oldest non-revoked secret" during rotation, with revocation as the final step.

**In practice, we observed webhooks continuing to be signed with a revoked secret for up to ~30 minutes after revocation in Partners.** `@shopify/shopify-api` accepts a single `apiSecretKey` with no way to trust a second secret, so every such webhook returns 401 from `authenticate.webhook()`. On high-volume topics (`orders/create`, `orders/updated`) the drop volume is not negligible.

## Reproduction

1. App running with `apiSecretKey = S1`.
2. In Partners: create `S2`, then revoke `S1` per the documented procedure.
3. Deploy with `apiSecretKey = S2`.
4. For up to ~30 minutes, inbound webhooks signed with `S1` all return 401.

Delaying revocation is **not** a mitigation: Shopify signs with whichever secret is oldest-non-revoked at send time, so as long as `S1` remains non-revoked, `S1`-signed webhooks continue indefinitely. Revocation is what triggers the ~30-min drop zone.

## Proposed API

Additive, backward compatible, inbound-only:

```ts
shopifyApp({
	apiSecretKey: process.env.SHOPIFY_API_SECRET,
	apiSecretKeyFallback: process.env.SHOPIFY_API_SECRET_FALLBACK,
	// ...
});
```

- `apiSecretKey` remains the single secret used for **outbound** signing (`generateLocalHmac`). Unchanged.
- `apiSecretKeyFallback`, when set, is consulted only for **inbound** HMAC validation. If both secrets fail the `safeCompare`, the request is rejected as before.
- **When validation succeeds via the fallback secret**, the SDK emits a `warn`-level log (via the existing `logger(config)` used by `hmac-validator.ts`) with `{ topic, shop, webhookId }`. This is the completion signal operators watch to know when to unset the fallback.

Implementation is localized to `packages/apps/shopify-api/lib/utils/hmac-validator.ts` plus a type addition in `base-types.ts` and corresponding test coverage.

## Target rotation procedure (with proposed API)

1. In Shopify Partners, create a new client secret. The current secret (`S_old`) remains valid.
2. Update env vars: set `SHOPIFY_API_SECRET = S_new` and `SHOPIFY_API_SECRET_FALLBACK = S_old`.
3. Deploy your app. It now accepts webhooks signed with either secret. Each webhook validated via the fallback secret emits a warn log, e.g. `warn: webhook HMAC validated via apiSecretKeyFallback { topic, shop, webhookId }`.
4. In Shopify Partners, revoke `S_old`.
5. Watch the warn logs. As long as they keep appearing, Shopify is still signing webhooks with `S_old`. **Do not unset the fallback yet.**
6. Once the warn logs stop (typically ~30 min post-revocation and with no new occurrences for a safety margin), unset `SHOPIFY_API_SECRET_FALLBACK` and redeploy.

No webhooks dropped. Step 6 is observable (log-driven) rather than guessed.

## Current user-space workaround

Validate HMAC in user space against both secrets, re-sign with the primary, delegate to the SDK so session lookup / topic parsing / admin client stay the SDK's concern:

```ts
export const authenticateWebhookWithFallback = async (request: Request) => {
	const primary = process.env.SHOPIFY_API_SECRET ?? '';
	const fallback = process.env.SHOPIFY_API_SECRET_FALLBACK;

	if (!fallback) return authenticate.webhook(request);

	const rawBody = await request.text();
	const received = request.headers.get('x-shopify-hmac-sha256') ?? '';

	const rebuild = (hmacOverride?: string) => {
		const headers = new Headers(request.headers);
		if (hmacOverride) headers.set('x-shopify-hmac-sha256', hmacOverride);
		return new Request(request.url, { method: request.method, headers, body: rawBody });
	};

	const primaryHmac = computeHmac(rawBody, primary);
	if (safeEqual(received, primaryHmac)) return authenticate.webhook(rebuild());

	if (safeEqual(received, computeHmac(rawBody, fallback))) {
		return authenticate.webhook(rebuild(primaryHmac)); // re-signed with primary
	}

	return authenticate.webhook(rebuild());
};
```

Works, but every Shopify app ends up maintaining ~40 lines of security-adjacent code independently, and the re-sign-then-delegate pattern is a workaround rather than a principled mechanism.

## Willingness to contribute

Happy to open a PR with tests and a changeset. Open to adjusting the API shape (naming, webhook-only vs admin-HMAC scope, etc.) based on maintainer preference.

## References

- [Shopify client-secret rotation](https://shopify.dev/docs/apps/build/authentication-authorization/client-secrets/rotate-revoke-client-credentials)
- [Shopify webhook HMAC verification](https://shopify.dev/docs/apps/build/webhooks/subscribe#verify-the-webhook)

morgan-coded:

Fixes #3183.

Shopify signs outbound requests with the *oldest non-revoked* client secret, and
(as reported in #3183) continues to do so for a window observed up to ~30 minutes
after the old secret is revoked. `@shopify/shopify-api` only ever validates inbound
HMACs against the single `apiSecretKey`, so the moment an app rotates `apiSecretKey`
to the new secret, every still-in-flight request signed with the old secret fails
HMAC validation and is rejected with a 401. On high-volume topics (`orders/create`,
`orders/updated`) the drop volume during a rotation is not negligible.

Adds an optional `apiSecretKeyFallback` config option, consulted **only** for
**inbound** HMAC validation (webhooks, Flow, and fulfillment-service — all of which
share `validateHmacString`):

- The primary `apiSecretKey` is checked first; the fallback is tried only if the
  primary fails. If both fail, the request is rejected exactly as before.
- **Outbound** signing (`generateLocalHmac`) is unchanged and always uses
  `apiSecretKey`, so the app never emits requests signed with the old secret.
- When a request validates via the fallback, a `Warning`-level log is emitted (via
  the existing `logger(config)`). Operators watch these logs to know when Shopify has
  stopped using the old secret, at which point the fallback can be safely unset.

Rotation procedure with this option:
1. Create the new secret in Partners (old secret stays valid).
2. Set `apiSecretKey = S_new`, `apiSecretKeyFallback = S_old`; deploy.
3. Revoke `S_old` in Partners.
4. Watch the warn logs; once they stop, unset `apiSecretKeyFallback` and redeploy.

This is additive and backward compatible: when `apiSecretKeyFallback` is unset,
behavior is identical to today.

### How to test your changes?

Targeted `jest --selectProjects library --testPathPatterns "hmac-validator"`
(25/25 passing after the hardening patch). The new coverage includes primary
match, fallback match, the warn log, no-warn-on-primary, both-fail rejection,
no-fallback rejection, and a Flow fallback request.

### Checklist

- [x] I have added/updated tests for my changes
- [x] I have added a changeset (`.changeset/webhook-hmac-secret-fallback.md`)
- [x] I have updated relevant documentation

### DIFF
--- a/.changeset/webhook-hmac-secret-fallback.md
+++ b/.changeset/webhook-hmac-secret-fallback.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-api': minor
+---
+
+Added an optional `apiSecretKeyFallback` config option. When set, it is used as a secondary secret when validating **inbound** HMAC signatures (webhooks, Flow, and fulfillment-service requests), so requests are not dropped during a client-secret rotation while Shopify is still signing with the previous secret. Outbound signing continues to use `apiSecretKey` only. When a request validates against the fallback secret, a `Warning`-level log is emitted so you can tell when the old secret is no longer in use and the fallback can be removed.
--- a/packages/apps/shopify-api/lib/base-types.ts
+++ b/packages/apps/shopify-api/lib/base-types.ts
@@ -27,6 +27,25 @@ export interface ConfigParams<
    */
   apiSecretKey: string;
   /**
+   * An additional, secondary API secret key used only to validate **inbound**
+   * HMAC signatures (webhooks, Flow, and fulfillment-service requests).
+   *
+   * This exists to avoid dropping inbound requests during a client-secret
+   * rotation. Shopify signs outbound requests with the oldest non-revoked
+   * secret, so after you rotate `apiSecretKey` to the new secret there can be a
+   * window (observed up to ~30 minutes after revoking the old secret) where
+   * requests are still signed with the previous secret. Setting the previous
+   * secret here lets those requests continue to validate.
+   *
+   * When a request validates against this fallback (and not `apiSecretKey`), a
+   * `Warning`-level log is emitted so you can observe when Shopify has stopped
+   * using the old secret and it is safe to unset this value.
+   *
+   * Outbound signing (`generateLocalHmac`) always uses `apiSecretKey` and is
+   * unaffected by this option.
+   */
+  apiSecretKeyFallback?: string;
+  /**
    * The scopes your app needs to access the API. Not required if using Shopify managed installation.
    */
   scopes?: string[] | AuthScopes;
--- a/packages/apps/shopify-api/lib/utils/__tests__/hmac-validator.test.ts
+++ b/packages/apps/shopify-api/lib/utils/__tests__/hmac-validator.test.ts
@@ -3,7 +3,14 @@ import crypto from 'crypto';
 import {testConfig} from '../../__tests__/test-config';
 import {AuthQuery} from '../../auth/oauth/types';
 import * as ShopifyErrors from '../../error';
-import {HMACSignator, getCurrentTimeInSec} from '../hmac-validator';
+import {
+  HMACSignator,
+  getCurrentTimeInSec,
+  validateHmacFromRequestFactory,
+} from '../hmac-validator';
+import {HmacValidationType} from '../types';
+import {ShopifyHeader, LogSeverity} from '../../types';
+import {type NormalizedRequest} from '../../../runtime';
 import {shopifyApi} from '../..';
 
 describe('validateHmac', () => {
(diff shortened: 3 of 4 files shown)


morgan-coded:
I have signed the CLA!

PR #3241, opened 2026-05-29
Docs: clarify that anonymous codegen operations are silently ignored

### LINKED ISSUE #1997: [api-codegen-preset] Types for anonymous operations are not generated / anonymous operations are ignored
## Overview/summary

In retrospective this seems like a reasonable behavior, but might be overlooked by other people as well and cost them some time and nerves. I think it's _**at least worth mentioning in the docs**_ that anonymous operations are counted as documents but omitted / ignored in generation phase.
Having only anonymous queries in the project will **not** trigger any errors but will result in empty output `*.generated.d.ts` 

Consider such an example straight from Shopify docs (https://shopify.dev/docs/api/admin-graphql/2024-10/queries/customers?language=Node.js):

```
const client = new shopify.clients.Graphql({session});
const data = await client.query({
  data: `query {
    customers(first: 5, query: "country:canada") {
      edges {
        node {
          id
        }
      }
    }
  }`,
});
```
apart from annotating with magic `#graphql` tag it also needs a name. So to get `graphql-codegen` going one will have to change above code to something like:

```
const client = new shopify.clients.Graphql({session});
const data = await client.query({
  data: `
  #graphql
  query customers {
    customers(first: 5, query: "country:canada") {
      edges {
        node {
          id
        }
      }
    }
  }`,
});
```

There's literally only a single mention of this fact in the `graphql-codegen` docs and it's in `React / Vue` chapter (I only found it because I knew what I was looking for): https://the-guild.dev/graphql/codegen/docs/guides/react-vue#writing-graphql-queries

Image

Respective "Vanilla TypeScript" chapter does not contain such a warning.

morgan-coded:

Fixes #1997

`@shopify/api-codegen-preset` only generates types for **named** queries and
mutations. Anonymous operations are counted as documents — so they don't trigger
the "no documents to parse" error — but they're skipped during generation. A
project containing only anonymous operations therefore produces an empty
`*.generated.d.ts` with no error or warning, which is easy to misdiagnose. The
README mentioned operations "must have a name" but didn't explain this
silent-failure behavior.

- Expands the note in the README's "Generating types" section to explain that
  anonymous operations are silently skipped (no error, empty output), and adds a
  short named-vs-anonymous example.

No code changes — this is the upstream graphql-codegen behavior; this PR just
documents it.

#### How was this verified?
- README-only change; no new links added (markdown link check unaffected).
- Confirmed against the package's own fixtures, which use only named operations.

## Type of change

- [x] Patch: Bug (non-breaking change which fixes an issue) — documentation fix
- [ ] Minor: New feature
- [ ] Major: Breaking change

## Checklist

- [x] I have used `pnpm changeset` to create a draft changelog entry
- [ ] I have added/updated tests for this change — n/a (docs only)
- [x] I have documented new APIs/updated the documentation for modified APIs

### DIFF
--- a/.changeset/docs-codegen-anonymous-operations.md
+++ b/.changeset/docs-codegen-anonymous-operations.md
@@ -1,0 +1,8 @@
+---
+'@shopify/api-codegen-preset': patch
+---
+
+Document that anonymous GraphQL operations are silently ignored during type
+generation. Only named queries and mutations produce types; a project
+containing only anonymous operations generates an empty `*.generated.d.ts`
+without an error. (README-only change.)
--- a/packages/api-clients/api-codegen-preset/README.md
+++ b/packages/api-clients/api-codegen-preset/README.md
@@ -332,7 +332,28 @@ pnpm graphql-codegen --project=UIExtensions
 > [!NOTE]
 > Codegen will fail if it can't find any documents to parse.
 > To fix that, either create a query or set the [`ignoreNoDocuments` option](https://the-guild.dev/graphql/codegen/docs/config-reference/codegen-config#configuration-options) to `true`.
-> Queries and mutations must have a name for the parsing to work.
+
+> [!IMPORTANT]
+> Your queries and mutations **must be named** for types to be generated.
+> Anonymous operations (e.g. `query { ... }` with no operation name) are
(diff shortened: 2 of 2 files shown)


morgan-coded:
I have signed the CLA!

PR #3265, opened 2026-06-12
Remove non-embedded React Router app configuration

byrichardpowell:
## Context

This PR is stacked on #3216. That PR removes the last package-driven redirect to `/auth/login` for embedded Admin document requests with missing `shop` / `host` params by rendering the App Bridge recovery page instead.

With that behavior in place, the React Router package should no longer expose public API surfaces that imply apps can opt into a non-embedded mode. The previously available knobs `<AppProvider embedded={false}>`, `shopifyApp({isEmbeddedApp: false})`, and template `embedded = false` guidance are misleading and can lead to broken auth loops.

## What changed

- Removed the `embedded` prop from React Router `AppProvider`.
- Made `AppProvider` require `apiKey` and always render App Bridge alongside Polaris web components.
- Added a runtime guard that rejects `shopifyApp({isEmbeddedApp: ...})` with a clear error.
- Renamed the internal `NonEmbeddedAdminContext` type to `MerchantCustomAdminContext` to avoid implying general non-embedded app support.
- Updated/generated React Router API docs.
- Added a major changeset with migration steps, including the corresponding React Router template updates.

## Migration notes

For app routes:

```diff
- <AppProvider embedded apiKey={apiKey}>
+ <AppProvider apiKey={apiKey}>
    <Outlet />
  </AppProvider>
```

For login/Polaris-only pages, remove `AppProvider` instead of using `embedded={false}`:

```diff
- <AppProvider embedded={false}>
-   <s-page>...</s-page>
- </AppProvider>
+ <s-page>...</s-page>
```

Remove `isEmbeddedApp` from React Router `shopifyApp` config:

```diff
const shopify = shopifyApp({
  // ...
- isEmbeddedApp: false,
});
```

For apps based on `shopify-app-template-react-router`, the dedicated `/auth/login` UI route can be removed. Move the `shopify.login(request)` call into the root route action and have the root login form post to itself instead of `/auth/login`:

```diff
- <Form method="post" action="/auth/login">
+ <Form method="post">
```

## Validation

- `pnpm --filter @shopify/shopify-app-react-router lint`
- `pnpm --filter @shopify/shopify-app-react-router exec jest --runInBand`
- `pnpm --filter @shopify/shopify-app-react-router build`
- `git diff --check`

Note: local commands emit the existing Node engine warning for `shopify-app-session-storage-mongodb` because this machine is on Node `v20.10.0` and that package requests `>=20.19.0`; the commands still pass.


Suleimanlatrsh:
Verified locally: the full react-router suite passes (343 tests) and eslint is clean on the changed files. The AppProvider tests were rewritten to assert the new always-App-Bridge behavior rather than just dropping the old assertions, which is the right approach. The major version bump is correctly present for the breaking change.

One thing to confirm before merge: the only red check is "Dispatch docs sync to shopify-dev," which is an infra/permissions job unrelated to this change. Can we confirm it isn't set as a required check, or get it re-run with the right token?

Suleimanlatrsh:
Follow-up on the CI question and merge readiness:

Confirmed the "Dispatch docs sync to shopify-dev" check is **not** a required status check on main (required set is cla, Lint, and the Test_* jobs). So that red check does not block merge. All required checks are green.

The remaining blocker is that main has moved and the branch now has conflicts. I tried updating the branch and it failed on these files:
- `packages/apps/shopify-app-react-router/docs/generated/generated_docs_data.json` (deleted on main, modified here)
- `packages/apps/shopify-app-react-router/docs/generated/generated_docs_data_v2.json`
- `packages/apps/shopify-app-react-router/src/react/components/AppProvider/AppProvider.tsx`
- `packages/apps/shopify-app-react-router/src/react/components/AppProvider/__tests__/AppProvider.test.tsx`
- `packages/apps/shopify-app-react-router/src/server/authenticate/admin/__tests__/doc-request-path.test.ts`

The two generated docs files can be regenerated after resolving. The `AppProvider.tsx` conflict is a real source conflict against this PR's refactor, so it needs your call rather than mine. Once you rebase and push, I'll re-verify locally and approve (code and tests were already green on my end).

Suleimanlatrsh:
I resolved the merge conflicts and pushed the update (merged main in). Summary of how each conflict was resolved:

- `AppProvider.tsx` and its test: kept this PR's direction (always render App Bridge, no `embedded` prop) and adopted main's `POLARIS_URL` constant instead of the hardcoded Polaris URL.
- `doc-request-path.test.ts`: took main's version, which is a superset (it adds the attacker-controlled-shop response-header regression test on top of the identical App Bridge render test).
- Generated docs: dropped the old `generated_docs_data.json` (removed on main's doc-generation cleanup) and regenerated `generated_docs_data_v2.json`. The regenerated output correctly no longer references `embedded` / `NonEmbeddedProps`.

Verified locally on the merged tree: `tsc -p tsconfig.build.json` clean, eslint clean, and the full react-router suite passes 347/347. Please sanity-check the AppProvider resolution since it's your breaking change. Waiting on CI, will approve once green.

Suleimanlatrsh (approved):
Approving. Resolved the merge conflicts and pushed; all required checks are green (the only red check is the non-required "Dispatch docs sync" infra job). Verified locally on the merged tree: tsc build clean, eslint clean, react-router suite 347/347. The conflict resolutions preserve this PR's non-embedded removal and fold in main's POLARIS_URL refactor plus the new attacker-controlled-shop regression test.

Suleimanlatrsh:
Update: verified end to end with real Shopify traffic. Ran the fixed build inside the React Router template and loaded it in a live Admin session. App renders via App Bridge (app-bridge.js loads, embedded UI shows), and a request with missing shop/host returns an App Bridge page instead of redirecting to login. Conflicts with main are resolved and docs regenerated; all required checks green.

Suleimanlatrsh:
Final status: approved, green, verified in a live Admin (App Bridge renders, no login redirect). Conflicts with main resolved and docs regenerated. Ready to merge.

PR #3299, opened 2026-06-24
Separate cookie signing key from app API secret

byrichardpowell:
## Summary

Sign OAuth state and session cookies with a key derived from the app API secret rather than the API secret itself. This prevents signed cookie HMACs from being reused as valid Shopify webhook HMACs.

Also increases OAuth state nonce entropy from 15 decimal digits to 32 random bytes (64 hex chars).

Token-exchange-only flows that do not use signed OAuth/session cookies are unaffected.

## Why

Previously, OAuth state and session cookies were signed as:

```
sig = HMAC_SHA256(apiSecretKey, value)
```

Shopify webhook HMACs are also:

```
X-Shopify-Hmac-Sha256 = HMAC_SHA256(apiSecretKey, rawWebhookBody)
```

Same key, same algorithm, same encoding. An attacker who triggered OAuth begin could obtain a `(state, state.sig)` pair and replay `state.sig` as a valid webhook HMAC with `rawBody == state`, allowing forged webhooks (e.g. `APP_UNINSTALLED`) to be accepted.

## What changed

- `runtime/crypto/utils.ts` — new `deriveSHA256HMACKey(secret, info)` using HKDF-SHA256; `createSHA256HMAC` now accepts derived key material.
- `runtime/http/cookies.ts` — signs cookies with a key derived using `info = "shopify-app-js/cookie-signing/v1"`; signature comparison uses a constant-time compare.
- `lib/auth/oauth/nonce.ts` — nonce is now 32 random bytes encoded as 64 hex chars.
- Tests:
  - New regression in `lib/webhooks/__tests__/validate.test.ts` — replaying a cookie signature as a webhook HMAC is rejected.
  - New Express integration regression in `__tests__/integration/oauth.test.ts` — forged `APP_UNINSTALLED` using captured OAuth state + signature is rejected and app installation is preserved.
  - Tests that hand-constructed legacy cookie signatures updated to use the real cookie signer.

## Compatibility

Existing signed OAuth state / session cookies minted before this change will no longer verify after deploy. Users in the middle of OAuth at deploy time may need to retry; users with stale session cookies will be re-authed by the normal flow.

## Validation

- `pnpm --filter @shopify/shopify-api lint` ✅
- `pnpm --filter @shopify/shopify-api tsc --noEmit` ✅
- `pnpm --filter @shopify/shopify-api test` — 612/612 ✅
- `@shopify/shopify-api` runtime tests (http/cookies/crypto) — 24/24 ✅
- `pnpm --filter @shopify/shopify-api build` ✅
- `pnpm --filter @shopify/shopify-app-express lint` ✅
- `pnpm --filter @shopify/shopify-app-express tsc --noEmit` ✅
- `pnpm --filter @shopify/shopify-app-express test` — 85/85 ✅


packages/apps/shopify-api/lib/auth/oauth/nonce.ts
```diff
@@ -1,15 +1,9 @@
...
-  const bytes = crypto.getRandomValues(new Uint8Array(length));
-
-  const nonce = bytes
-    .map((byte: number) => {
-      return byte % 10;
-    })
-    .join('');
-
-  return nonce;
+  return Array.from(bytes, (byte) => byte.toString(16).padStart(2, '0')).join(
```
craigmichaelmartin: Could use `asHex` function which already exists from `runtime/crypto/utils.ts`

packages/apps/shopify-api/runtime/crypto/utils.ts
```diff
@@ -34,6 +43,31 @@ export async function createSHA256HMAC(
...
+    'HKDF',
+    false,
+    ['deriveBits'],
+  );
+
+  return cryptoLib.subtle.deriveBits(
+    {
+      name: 'HKDF',
+      hash: 'SHA-256',
+      salt: new Uint8Array(0),
```
craigmichaelmartin, line 63: This is fine, but if we wanted to be :chef-kiss: we could use `apiKey` (stable, non-secret identifier) as the salt

craigmichaelmartin (approved):
Looks good!

packages/apps/shopify-api/lib/auth/oauth/nonce.ts
byrichardpowell: Good catch, applied in 44dc59e.

packages/apps/shopify-api/runtime/crypto/utils.ts
byrichardpowell: Thanks for the suggestion. Going to leave the salt empty here:

- HKDF spec treats the salt as a non-secret value used to strengthen the extract step when the IKM has weak/uneven entropy. Our IKM (`apiSecretKey`) is already high-entropy random material, which is the case where an empty salt is explicitly fine.
- Domain separation between contexts is what `info` is for, and we already pass `shopify-app-js/cookie-signing/v1` there.
- `apiKey` adds no entropy (it's a stable public identifier), and using it would mean plumbing `apiKey` through `Cookies` and every call site for marginal benefit in our threat model.

Happy to revisit if you feel strongly!

PR #3328, opened 2026-07-13
Fix duplicate App Proxy security parameter handling

byrichardpowell:

App Proxy authentication currently parses the same query string in two different ways: request handling reads the first value for parameters such as `shop`, while converting `URLSearchParams` with `Object.fromEntries` keeps the last value for HMAC validation. A request with repeated security parameters can therefore make session lookup and signature verification operate on different values.

Application-defined App Proxy parameters may legitimately repeat, so rejecting every repeated parameter would be unnecessarily breaking.

- moves duplicate-aware normalization into the shared `@shopify/shopify-api` HMAC validator
- passes `URLSearchParams` directly from the Remix and React Router App Proxy authenticators so duplicate provenance is retained
- rejects repeated App Proxy protocol/security parameters (`hmac`, `shop`, `signature`, and `timestamp`)
- preserves repeated application-defined parameters using the existing comma-joined App Proxy canonicalization
- uses a null-prototype accumulator so application parameters such as `constructor`, `toString`, and `__proto__` are handled safely
- adds shared validator tests and end-to-end authenticator regressions for both frameworks
- includes patch changesets for `@shopify/shopify-api`, `@shopify/shopify-app-react-router`, and `@shopify/shopify-app-remix`

### Validation

- `pnpm lint`
- `pnpm build --filter='./packages/apps/shopify-api...' --filter='./packages/apps/shopify-app-express...' --filter='./packages/apps/shopify-app-remix...' --filter='./packages/apps/shopify-app-react-router...'`
- `pnpm test:ci --filter='./packages/apps/shopify-api'`
- `pnpm test:ci --filter='./packages/apps/shopify-app-express'`
- `pnpm test:ci --filter='./packages/apps/shopify-app-react-router'`
- `pnpm test:ci --filter='./packages/apps/shopify-app-remix'`

## Type of change

## Checklist

- [x] I have documented new APIs/updated the documentation for modified APIs (not applicable; no new public API or documentation surface)

### DIFF
--- a/packages/apps/shopify-app-remix/src/server/authenticate/public/appProxy/authenticate.ts
+++ b/packages/apps/shopify-app-remix/src/server/authenticate/public/appProxy/authenticate.ts
@@ -95,10 +95,9 @@ async function validateAppProxyHmac(
       searchParams.delete('index');
     }
 
-    let isValid = await api.utils.validateHmac(
-      Object.fromEntries(searchParams.entries()),
-      {signator: 'appProxy'},
-    );
+    let isValid = await api.utils.validateHmac(searchParams, {
+      signator: 'appProxy',
+    });
 
     if (!isValid) {
       const cleanPath = url.pathname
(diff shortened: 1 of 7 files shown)


packages/apps/shopify-app-remix/src/server/authenticate/public/appProxy/authenticate.ts
gonzaloriestra, line 1: Is this file identical to packages/apps/shopify-app-react-router/src/server/authenticate/public/appProxy/authenticate.ts? In that case, maybe we can extract it, no? No need to do it now.

gonzaloriestra (approved):
Untested, but makes sense 👍

Suleimanlatrsh:
Approving. Verified locally: hmac-validator, remix appProxy, and react-router appProxy suites all pass, and reverting the fix makes 5 of the new tests fail, so the coverage genuinely guards the behavior.

On Gonzalo's question: the remix and react-router `authenticate.ts` files are byte-for-byte identical here. Not a blocker (his comment says no need now), but worth a follow-up issue to extract the shared logic so the two copies don't drift.

Suleimanlatrsh (approved):

Suleimanlatrsh:
Update: verified end to end with real Shopify traffic. Ran the app in the React Router template on a live dev store with an App Proxy configured, and had Shopify sign and forward a real duplicate-param request. The fix's comma-join reproduces Shopify's actual signature; the shipped build rejects the same real request (400) while the fixed logic validates it. Confirmed against a genuine Shopify signature, not just unit tests.

Suleimanlatrsh:
Final status: approved, green, and verified against real Shopify traffic (real signed App Proxy request with duplicate params; shipped rejects 400, fix accepts 200). Ready to merge.

PR #3343, opened 2026-07-27
feat(express): Add support for expiring offline access tokens

Suleimanlatrsh:

Expiring offline access tokens are becoming mandatory. Support was added to the core API, React Router, Remix, PHP, and Python packages, but not to the Express package. This PR closes that gap so Express apps can mint expiring offline tokens and refresh them automatically.

Reference implementation: #2941 (React Router), #2946 (Remix), #2938 (core).

- Adds a new `expiringOfflineAccessTokens` future flag to the Express `shopifyApp` config (off by default, backward compatible).
- Requests an expiring offline token at the OAuth callback (`api.auth.callback({..., expiring: config.future?.expiringOfflineAccessTokens})`). Express has no token-exchange strategy, so the code flow callback is the only place offline tokens are minted; without this the platform never returns a refresh token and the refresh path is dead code.
- Ports the offline-session helpers from the React Router package, adapted to Express (no `ShopifyAdmin` distribution concept, plain error handling instead of throwing a web `Response`):
  - `refresh-token`
  - `ensure-offline-token-is-not-expired` (refreshes when the flag is on, the token expires within 5 minutes, and a refresh token is present)
  - `load-offline-session`
  - `ensure-valid-offline-session`
- Wires refresh-on-use into `validateAuthenticatedSession` for the offline-token case. The refresh runs before the `isActive` check so an already-expired token is refreshed rather than treated as inactive. A failed refresh is caught, logged, and falls through to re-auth.
- Exposes a public `shopify.ensureValidOfflineSession(shop)` helper so background work (webhooks, cron, queues) can load an offline session with a valid token, since Express has no `unauthenticated.admin` factory.
- Exports the `FutureFlags` type from the package entry point.
- Tests: callback requests an expiring token only when the flag is on; helper unit tests; `validateAuthenticatedSession` refresh-before-use and refresh-failure-falls-through.
- Docs for the new flag and helper in `docs/reference/shopifyApp.md`.

Notes / decisions:
- Dropped RR's `distribution !== ShopifyAdmin` guard. It is not needed here because the `session.refreshToken` presence check already blocks tokens that cannot be refreshed, and Express has no `ShopifyAdmin` distribution.
- Session storage must persist the `refreshToken` and `refreshTokenExpires` fields for this to work.
- Known limitation (shared with RR/Remix): within the refresh window, concurrent requests can each trigger a refresh. If the platform rotates the refresh token on use, concurrent refreshes may conflict.

## Type of change

## Checklist


.changeset/express-expiring-offline-tokens.md
```diff
@@ -0,0 +1,20 @@
...
+To enable expiring offline access tokens, your session storage must persist the `refreshToken` and `refreshTokenExpires` fields. Enable the flag in your app configuration:
+
+```ts
+const shopify = shopifyApp({
+  // ... other config
+  future: {
+    expiringOfflineAccessTokens: true,
+  },
+});
+```
```
byrichardpowell: This is good.  Can we flesh it out more? Ideally this changeset should be a comprehensive guide on every change a developer needs to make (Flags, DB Migrations, code changes).  I would checkout the node template (uses express), commit locally, link the template to the local build of this package (there are docs in the repo), perform the migration yourself, then use a git diff to see every change you made.  Confirm the migration is complete by running the template E2E inside the shopify iframe.

.changeset/express-expiring-offline-tokens.md
```diff
@@ -0,0 +1,20 @@
...
+We are introducing support for expiring offline access tokens. This feature improves security by limiting the lifespan of offline access tokens and automatically refreshing them using refresh tokens.
+
+- **New future flag**: Added `expiringOfflineAccessTokens` (boolean) to the `future` configuration in `shopifyApp`. When enabled, the library checks whether the offline access token is expired or nearing expiration (within 5 minutes) and, if so, refreshes it using the stored refresh token. Defaults to `false` for backward compatibility.
+
+- **Automatic token refresh**: `validateAuthenticatedSession` refreshes the offline token before use when the flag is enabled. A new `shopify.ensureValidOfflineSession(shop)` helper is also exposed so background work (webhooks, cron jobs, queues) can load an offline session with a valid token.
+
+To enable expiring offline access tokens, your session storage must persist the `refreshToken` and `refreshTokenExpires` fields. Enable the flag in your app configuration:
+
+```ts
+const shopify = shopifyApp({
```
byrichardpowell: Where does shopifyApp come from? Generally with docs & changesets it's best to provide complete examples.

packages/apps/shopify-app-express/src/helpers/__tests__/ensure-valid-offline-session.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    shopify.config.future = {expiringOfflineAccessTokens: false};
+    await storeOfflineSession();
+
+    const result = await ensureValidOfflineSession(
+      {api: shopify.api, config: shopify.config},
+      TEST_SHOP,
+    );
+
+    expect(result?.accessToken).toBe('old-access-token');
+  });
```
byrichardpowell: Does the developer get any feedback if they call `ensureValidOfflineSession` when the future flag is turned off? 

if not that seems is that a bit of a trap? E.g: The developer thinks they are using expiring offline tokens correctly, but really it's just a no op?  Some options I can think of:

1. Only expose the method if the future flag is enabled (might be difficult)
2. Throw an error if the method is called but the future flag is not enabled.

packages/apps/shopify-app-express/src/middlewares/validate-authenticated-session.ts
```diff
@@ -52,6 +53,20 @@ export function validateAuthenticatedSession({
       let shop =
         api.utils.sanitizeShop(req.query.shop as string) || session?.shop;
 
+      if (session && !config.useOnlineTokens) {
```
byrichardpowell: Should we also be checking the future flag for expiring offline tokens is enabled?

byrichardpowell (commented):
Left some comments.  Please can you also confirmed you have tested the following flows E2E in the node template:

1. The future flag turned of (we don't store refresh tokens, we neevr try to refresh)
2. The future flag turned on with online tokens (we don't store refresh tokens, we neever try to refresh)
3. The future flag turned on, with offline tokens (we store refresh tokens, middleware refreshes them automatically)

You should be able to build this version of the package, add it to a local checkout of the note template then install an app using that template to check the workflows.

It's important we migrate an app running locally, confirm the behaviour, then document the changes you have to make in the changeset.  We sometimes use diff codeblocks showing before and after to make it super clear how to migrate

.changeset/express-expiring-offline-tokens.md
Suleimanlatrsh: Done. Rewrote the changeset into a step-by-step migration guide: update the dependencies, make sure session storage persists the refreshToken/refreshTokenExpires fields, enable the flag, re-install so an expiring token is minted, and use the new helper in background work. Added before/after diff blocks for the flag change and the background-job usage.

.changeset/express-expiring-offline-tokens.md
Suleimanlatrsh: Done. Added the import and a complete, copy-pasteable example so it stands on its own.

packages/apps/shopify-app-express/src/helpers/__tests__/ensure-valid-offline-session.test.ts
Suleimanlatrsh: Good catch. ensureValidOfflineSession now throws a ShopifyError when the flag is off, so a misconfigured app fails loudly instead of silently skipping the refresh. Updated the test to expect the throw.

packages/apps/shopify-app-express/src/middlewares/validate-authenticated-session.ts
Suleimanlatrsh: Done. The middleware now checks config.future?.expiringOfflineAccessTokens in the condition, so it skips the refresh path entirely when the flag is off. Also gated the OAuth callback's expiring request on the same flag, since without it the platform never returns a refresh token and the whole path is dead code.

Suleimanlatrsh:
Ran all three flows end to end in the Node template (Express) with this branch's build linked in, and confirmed the stored sessions on a real dev store:

- Flag off: offline token has no refresh token and no expiry (permanent) — existing behavior preserved.
- Flag on + offline tokens: offline token comes back expiring (~1h) with a stored refresh token (refreshTokenExpires ~3 months). The OAuth callback requesting expiring is what triggers this.
- Flag on + online tokens: the offline session still captures a refresh token, but the online session used for requests has none, and the middleware skips the refresh path (gated on !useOnlineTokens).

These match the automated tests. Also merged latest main and resolved a conflict in the middleware test (kept both the new 503 test and the expiring-token tests). All review comments above are addressed.

.changeset/express-expiring-offline-tokens.md
```diff
@@ -0,0 +1,60 @@
...
+
+## How to migrate
+
+### 1. Update your dependencies
+
+Update `@shopify/shopify-app-express`, `@shopify/shopify-api`, and your session storage package to the latest version.
+
+### 2. Make sure your session storage persists refresh tokens
+
+The refresh token is stored on the session, so your session storage must persist the `refreshToken` and `refreshTokenExpires` fields. The official SQL-based adapters (SQLite, MySQL, PostgreSQL, etc.) add these columns automatically through their migrations when you update to the latest version. If you use a custom session storage, add the two fields to your schema.
```
byrichardpowell: I think the versions of the session storage adapters that add support for expiring tokens include docs on migrating. In which case, I think it's worth linking to those guides for each Session storage adapter.

.changeset/express-expiring-offline-tokens.md
```diff
@@ -0,0 +1,60 @@
...
+      path: "/api/webhooks",
+    },
++   future: {
++     expiringOfflineAccessTokens: true,
++   },
+    sessionStorage: new SQLiteSessionStorage(DB_PATH),
+  });
+```
+
+### 4. Re-install the app so it mints an expiring token
```
byrichardpowell: Is this a distinct step? I also wonder if it's somehwat mis-leading.  I don't think you need to reinstall the app, you just need a merchant to visit your app and for your app to call ensureValidOfflineSession.  Is that correct?

If so, maybe we remove this section and section 5 becomes section 4?

byrichardpowell (approved):
Just 2 small comments on the changesets.  Good work 👍

.changeset/express-expiring-offline-tokens.md
Suleimanlatrsh: Good call. Added links to each adapter's README (SQLite, MySQL, PostgreSQL, Prisma) and noted that the SQL adapters migrate automatically while Prisma needs a schema change + migration.

.changeset/express-expiring-offline-tokens.md
Suleimanlatrsh: You're right that new installs need nothing extra, and I've reworded this. One caveat specific to this package: since it uses the OAuth code flow (not token exchange like React Router/Remix), simply visiting the app plus calling ensureValidOfflineSession will not upgrade an existing non-expiring token. That helper is a no-op on a token with no refresh token and no expiry, and the still-valid token means ensureInstalledOnShop won't re-trigger OAuth. So existing installs only convert on their next re-auth. I dropped the misleading 're-install' step and replaced it with a short 'How existing installs pick up expiring tokens' note covering new vs existing. Open to exposing migrateToExpiringToken in a follow-up if we want a no-re-auth upgrade path.

PR #3350, opened 2026-07-29
fix: log actual shop in 'Authenticating admin request' (fixes #2916)

### LINKED ISSUE #2916: Please remove log that is always null
## Overview

This log message:

```
      logger.info('Authenticating admin request', {
        shop: getShopFromRequest(request),
      });
```

[Link to the location of the log](https://github.com/Shopify/shopify-app-js/blob/3e6f4ad26565f58176848140aaeed16237c05b5e/packages/apps/shopify-app-remix/src/server/authenticate/admin/authenticate.ts#L160-L162)

Fills my logs with `null`. I actually don't mind it, if shop were populated, but it's kinda silly to have it claim to have a value and have it be null. Especially since a couple of lines later the debug log actually does print shop and sessionId - an eminently more useful log, at a lower visibility level. If we don't want to put shop sessionId in production for customer privacy reasons, at least delete the context parameter that is always `null` because `shop` is not in the query parameters anymore? Thanks!

Suleimanlatrsh:
## Why
The `info` log at the start of `authenticateAdmin` called `getShopFromRequest(request)`, which reads the `shop` query param. For token-exchange / embedded requests the shop lives in the JWT, not the URL, so the log always emitted `shop: null`. Fixes #2916.

## What
- Move `logger.info('Authenticating admin request', {shop})` to **after** `getSessionTokenContext` resolves, so it logs the shop extracted from the session token.
- Ports the fix to **both** `@shopify/shopify-app-remix` **and** `@shopify/shopify-app-react-router` (the latter had the identical bug at the same line and is the package new apps use).
- Adds a regression test in each package asserting the logged shop value (verified it fails with `{shop: null}` on the current code and passes with the fix).

## Supersedes / relationship
Supersedes #3159 (which only touched remix and never ran the test matrix because it was opened by `github-actions[bot]`). This branch is pushed from a human account so CI runs.

## Behaviour tradeoff (please confirm)
Moving the log means a request that throws inside `getSessionTokenContext` (e.g. a token-exchange bounce) no longer emits this specific info line. The earlier `debug` 'Attempting to authenticate session token' and the failure-path logs still fire, so observability loss is limited to the info-level line. Alternative would be to keep the log first and just drop the misleading `shop` key. Flagging so a reviewer picks deliberately.

### DIFF
--- a/.changeset/fix-null-shop-in-admin-auth-log.md
+++ b/.changeset/fix-null-shop-in-admin-auth-log.md
@@ -1,0 +1,6 @@
+---
+'@shopify/shopify-app-remix': patch
+'@shopify/shopify-app-react-router': patch
+---
+
+Fix 'Authenticating admin request' info log always showing null for shop. The shop is now logged after it is extracted from the session token context, so it reflects the actual shop value instead of null.
--- a/packages/apps/shopify-app-react-router/src/server/authenticate/admin/__tests__/session-token-header-path.test.ts
+++ b/packages/apps/shopify-app-react-router/src/server/authenticate/admin/__tests__/session-token-header-path.test.ts
@@ -1,3 +1,5 @@
+import {LogSeverity} from '@shopify/shopify-api';
+
 import {shopifyApp} from '../../..';
 import {
   APP_URL,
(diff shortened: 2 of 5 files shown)


byrichardpowell (approved):

Suleimanlatrsh:
Folding these changes into the existing PR #3159 instead. Closing this duplicate.

PR #3097, opened 2026-03-13
[Feature] Add token exchange auth strategy and hooks support to shopify-app-express

mrmarufpro:
### WHY are these changes introduced?

Fixes #630, Fixes #3208 — adds token exchange authentication support to `shopify-app-express`, bringing it to parity with the Remix package.

Shopify's [token exchange API](https://shopify.dev/changelog/new-oauth2-token-exchange-api-shopify-managed-install-authorization-flows-available) reduces flickering and OAuth redirects for embedded apps. While this was already supported in the Remix package, Express apps had no way to use it. This PR implements token exchange for Express behind the `unstable_newEmbeddedAuthStrategy` future flag so existing apps are unaffected.

### WHAT is this pull request doing?

- **`future` config option** — adds a feature flag system (`unstable_newEmbeddedAuthStrategy`, `expiringOfflineAccessTokens`) with logging for disabled flags
- **Token exchange middleware** (`perform-token-exchange.ts`) — when `unstable_newEmbeddedAuthStrategy` is enabled and a Bearer token is present in the `Authorization` header, `validateAuthenticatedSession` performs a token exchange instead of redirecting to OAuth
- **`ensureInstalled` shortcut** — when the token exchange strategy is on, `ensureInstalledOnShop` skips the session check and goes straight to embedding/loading the app
- **`hooks.afterAuth`** — async callback invoked after both OAuth and token exchange flows, deduplicated via `IdempotentPromiseHandler` to prevent double-firing on concurrent requests
- **`registerWebhooks({session})`** — convenience method on `ShopifyApp` wrapping `api.webhooks.register`
- **`expiringOfflineAccessTokens` flag** — threads the flag into `auth.callback` and `auth.tokenExchange` calls
- **`ensureOfflineTokenIsNotExpired` helper** — proactively refreshes offline tokens nearing expiry in the standard OAuth path
- Docs updated for `shopifyApp`, `validateAuthenticatedSession`, `auth`, and a new token-exchange guide added
- Tests added for all new code paths

## Type of change

- [x] Minor: New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have used `pnpm changeset` to create a draft changelog entry (do NOT update the `CHANGELOG.md` files manually)
- [x] I have added/updated tests for this change
- [x] I have documented new APIs/updated the documentation for modified APIs (for public APIs)

PR #3351, opened 2026-07-30
feat(express): Add token exchange auth strategy

Suleimanlatrsh:

Adds token exchange authentication to `shopify-app-express`, bringing it to parity with Remix/React Router. Token exchange lets embedded apps get access tokens without the OAuth redirect flicker, and it's the path needed as expiring offline tokens become mandatory (Jan 1 2027).

Builds on the original work by @mrmarufpro in #3097 (thank you), reworked per review and layered on top of the expiring-offline-token support merged in #3343.

Behind the `unstable_tokenExchange` future flag (off by default):

- **Strategy selection, one branch up front.** `validateAuthenticatedSession` decides once: flag on + `isEmbeddedApp` → token exchange; otherwise → the legacy OAuth code flow. The two paths are separate functions.
- **Token exchange middleware**: swaps the session token for an access token (offline, plus online when `useOnlineTokens`), stores it, runs `afterAuth` once (idempotent under concurrency), and handles invalid/revoked tokens.
- **Request-type parity with React Router**: document requests (page loads) with a missing/stale token render the App Bridge shim to fetch a fresh token and reload; fetch requests get a 401 with the retry header only when a retry would help.
- **`hooks.afterAuth`** config option, fired on both the token-exchange and OAuth callback paths.
- **`shopify.registerWebhooks({session})`** helper, since token exchange skips the OAuth callback that registered webhooks.
- **`ensureInstalledOnShop` is token-exchange aware**: under managed installation, install state is not a session-storage question, so when the flag is on it skips the storage check, embeds the app, and lets the first authenticated request mint the session. This avoids a dead-end on a fresh shop (no session yet) where the page load would otherwise redirect into the guarded OAuth routes.
- **OAuth routes guarded**: `auth.begin()` / `auth.callback()` return a clear error when token exchange is active.
- **Startup log** advertising the `unstable_tokenExchange` and `expiringOfflineAccessTokens` flags.
- Reuses the merged expiring-offline-token module's `WITHIN_MILLISECONDS_OF_EXPIRY` constant. Note: the token-exchange path re-exchanges rather than calling the refresh-token helper, so it does not duplicate the expiring-token refresh logic.

### Tests

- Token-exchange middleware tests: valid token → exchange + store + `afterAuth`; invalid token on a fetch request → 401 + retry header; invalid token on a document request → App Bridge bounce; no token exchange when not embedded or flag off.
- All existing tests pass (105 total).

### End-to-end testing

Verified live on a real dev store (`suli-store`) with this branch's build linked into the Node template, across the full embedded/non-embedded x token-exchange/OAuth x online/offline x expiring matrix:

| # | Config | Result |
| --- | --- | --- |
| 1 | embedded, token exchange, offline | token exchange runs, no OAuth redirect, `{"count":18}` from Admin GraphQL |
| 2 | embedded, OAuth, offline | OAuth code flow, renders in admin |
| 3 | embedded, token exchange, online | two exchanges (offline + online), online session |
| 4 | embedded, OAuth, online | two OAuth passes, online session |
| 5 | embedded, token exchange, offline + expiring | refresh token saved (confirmed via storage read-back) |
| 6 | embedded, OAuth, offline + expiring | refresh token saved |
| 7 | non-embedded, OAuth, offline | renders outside admin, Admin GraphQL works |
| 8 | non-embedded, OAuth, online | online session, works |
| 9 | non-embedded, OAuth, offline + expiring | refresh token saved (confirmed via storage read-back) |
| 10 | non-embedded, token exchange on | throws at startup with explicit message (`unstable_tokenExchange requires an embedded app`) |
| 11 | any, hit `/api/auth` while token exchange on | returns a clear error (OAuth routes not used under token exchange) |

Confirmed:
- Token exchange only runs for embedded apps; non-embedded falls back to OAuth, and non-embedded + token exchange throws with explicit feedback.
- Expiring offline tokens store a refresh token on both the token-exchange and OAuth paths (embedded and non-embedded), verified by reading the session back from storage (`hasRefreshToken: true`, `refreshTokenExpires` set).
- Token exchange loads with no OAuth redirect/flicker; the exchanged token makes real Admin GraphQL calls.
- Plus 107 automated tests and green CI.

### Type of change

- [x] Minor: New feature (non-breaking, off by default)

### Checklist

- [x] Changeset added (credits @mrmarufpro)
- [x] Tests for the token-exchange path
- [x] Docs (reference + token-exchange guide)

### DIFF
--- a/.changeset/express-token-exchange.md
+++ b/.changeset/express-token-exchange.md
@@ -2,8 +2,38 @@
 '@shopify/shopify-app-express': minor
 ---
 
-Add token exchange authentication support for embedded apps, behind the `unstable_tokenExchange` future flag. When enabled, embedded apps fetch access tokens via token exchange (no OAuth redirect flicker) instead of the auth code flow. Non-embedded apps, and apps with the flag off, continue to use the OAuth code flow.
+Add token exchange authentication support for embedded apps, behind the `tokenExchange` future flag.
 
-Also adds a `hooks.afterAuth` config option and a `shopify.registerWebhooks({session})` helper for registering webhooks outside the OAuth callback.
+When enabled, embedded apps get access tokens via [token exchange](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-app-express/docs/reference/guides/token-exchange.md) instead of the OAuth redirect flow: on load, App Bridge provides a short-lived session token that the app exchanges with Shopify for an API access token. This removes the redirect flicker the Auth Code flow causes on load. Non-embedded apps, and apps with the flag off, continue to use the Auth Code flow. (Auth Code flow and token exchange are both OAuth flows.)
 
-This builds on the original work by @mrmarufpro in #3097. Thank you for contributing this.
+**Requirements:** the app must be embedded (`isEmbeddedApp: true`), use [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation), and load [App Bridge](https://shopify.dev/docs/api/app-bridge-library) on the frontend. Enabling the flag on a non-embedded app throws at startup.
+
+**Enabling it:**
+
+```diff
+  const shopify = shopifyApp({
+    api: {/* ... */},
+    auth: {path: '/api/auth', callbackPath: '/api/auth/callback'},
+    webhooks: {path: '/api/webhooks'},
++   future: {
++     tokenExchange: true,
++   },
++   hooks: {
++     afterAuth: async ({session}) => {
++       await shopify.registerWebhooks({session});
++     },
++   },
+  });
+```
+
+**What changes when enabled:**
+
+- `validateAuthenticatedSession` uses token exchange for embedded apps (decided once, up front: token exchange when the flag is on and the app is embedded, otherwise the Auth Code flow).
+- Fetch requests with a missing or stale session token get a `401` with the `X-Shopify-Retry-Invalid-Session-Request` header, so App Bridge fetches a fresh token and retries.
+- Document requests with a missing or stale session token render App Bridge, which fetches a fresh token and reloads. No top-level OAuth redirect.
+- The OAuth routes (`auth.begin` / `auth.callback`) are not used and return an error if called.
+- Revoked (but unexpired) tokens are not re-authenticated automatically. Handle a `401` from the Admin API by retrying the request (which triggers a fresh exchange) or re-authenticating. Expired tokens are handled automatically.
+
+This also adds a `hooks.afterAuth` config option and a `shopify.registerWebhooks({session})` helper for registering webhooks outside the OAuth callback.
+
+Builds on the original work by [@mrmarufpro](https://github.com/mrmarufpro) in [#3097](https://github.com/Shopify/shopify-app-js/pull/3097). Thank you for contributing this.
--- a/.changeset/express-token-exchange.md
+++ b/.changeset/express-token-exchange.md
@@ -1,0 +1,9 @@
+---
+'@shopify/shopify-app-express': minor
+---
+
+Add token exchange authentication support for embedded apps, behind the `unstable_tokenExchange` future flag. When enabled, embedded apps fetch access tokens via token exchange (no OAuth redirect flicker) instead of the auth code flow. Non-embedded apps, and apps with the flag off, continue to use the OAuth code flow.
+
+Also adds a `hooks.afterAuth` config option and a `shopify.registerWebhooks({session})` helper for registering webhooks outside the OAuth callback.
+
+This builds on the original work by @mrmarufpro in #3097. Thank you for contributing this.
--- a/packages/apps/shopify-app-express/docs/reference/guides/token-exchange.md
+++ b/packages/apps/shopify-app-express/docs/reference/guides/token-exchange.md
@@ -1,8 +1,8 @@
 # Token exchange
 
-Token exchange lets embedded apps get access tokens without the OAuth redirect flow. When an embedded app loads, App Bridge provides a short-lived session token; the app exchanges that token with Shopify for an API access token. This removes the redirect flicker that the OAuth code flow causes on load.
+Token exchange lets embedded apps get access tokens without the OAuth redirect flow. When an embedded app loads, App Bridge provides a short-lived session token; the app exchanges that token with Shopify for an API access token. This removes the redirect flicker that the Auth Code flow causes on load.
 
-Support is behind the `unstable_tokenExchange` future flag and is off by default.
+Support is behind the `tokenExchange` future flag and is off by default.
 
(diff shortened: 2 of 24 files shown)


.changeset/express-token-exchange.md
```diff
@@ -0,0 +1,9 @@
+---
+'@shopify/shopify-app-express': minor
+---
+
+Add token exchange authentication support for embedded apps, behind the `unstable_tokenExchange` future flag. When enabled, embedded apps fetch access tokens via token exchange (no OAuth redirect flicker) instead of the auth code flow. Non-embedded apps, and apps with the flag off, continue to use the OAuth code flow.
+
+Also adds a `hooks.afterAuth` config option and a `shopify.registerWebhooks({session})` helper for registering webhooks outside the OAuth callback.
```
byrichardpowell: Please can we provide code examples for how to enable it (git diffs are good)

.changeset/express-token-exchange.md
```diff
@@ -0,0 +1,9 @@
+---
+'@shopify/shopify-app-express': minor
+---
+
+Add token exchange authentication support for embedded apps, behind the `unstable_tokenExchange` future flag. When enabled, embedded apps fetch access tokens via token exchange (no OAuth redirect flicker) instead of the auth code flow. Non-embedded apps, and apps with the flag off, continue to use the OAuth code flow.
```
byrichardpowell: Remove `unstable` from this future flag.  It's not unstable anymore, it's just the green path instead.

packages/apps/shopify-app-express/docs/reference/guides/token-exchange.md
```diff
@@ -0,0 +1,46 @@
...
+
+Support is behind the `unstable_tokenExchange` future flag and is off by default.
+
+## Requirements
+
+- The app must be **embedded** (`isEmbeddedApp: true`, the default).
+- The app must use [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation).
+- The frontend must load [App Bridge](https://shopify.dev/docs/api/app-bridge-library), which provides the session token.
+
+Non-embedded apps, and apps with the flag off, continue to use the OAuth code flow.
```
byrichardpowell: IIRC, "continue to use Auth Code flow" is more accurate, since Auth code flow and token exchange are both OAuth Spec's

.changeset/express-token-exchange.md
byrichardpowell: Seems like you already have great docs in `token-exchange.md`.  Let's duplicate the important stuff here.

.changeset/express-token-exchange.md
```diff
@@ -0,0 +1,9 @@
...
+
+Also adds a `hooks.afterAuth` config option and a `shopify.registerWebhooks({session})` helper for registering webhooks outside the OAuth callback.
+
+This builds on the original work by @mrmarufpro in #3097. Thank you for contributing this.
```
byrichardpowell: Make the user name and PR number Links?

mrmarufpro:
Thanks for picking this up @Suleimanlatrsh, and for the credit.

One thing before this ships: `ensureInstalledOnShop` isn't touched here, and I think it dead-ends any store with no offline session in storage.

Token exchange only writes the session inside `performTokenExchange`, which is reachable only through `validateAuthenticatedSession`. The template mounts that on `/api/*`, and `ensureInstalledOnShop` on everything else:

```js
app.use("/api/*", shopify.validateAuthenticatedSession());
app.use("/*", shopify.ensureInstalledOnShop(), /* serve SPA */);
```

So document requests never reach token exchange, and `ensureInstalledOnShop` still infers installation from storage:

```ts
const session = await config.sessionStorage.loadSession(api.session.getOfflineId(shop));

if (!session && !req.originalUrl.match(exitIframeRE)) {
  return redirectToAuth({req, res, api, config});
}
```

On a store that hasn't exchanged yet that means: page load → no session → `redirectToAuth` → `/api/auth` → the 400 added in this PR. A reload repeats it, even though the request carries an `id_token` that could be exchanged on the spot. Same dead end after uninstall/reinstall (`APP_UNINSTALLED` clears sessions) or a revoked access token.

#3097 handled it by not consulting storage when the flag is on:

```ts
if (config.future.unstable_tokenExchange) {
  if (api.config.isEmbeddedApp && req.query.embedded !== '1') {
    await embedAppIntoShopify(api, config, req, res, shop);
    return undefined;
  }
  addCSPHeader(api, req, res);
  return next();
}
```

Under managed install, install state isn't a storage question — so the middleware just forces the embed, sets CSP, and continues, and the first `/api/*` call mints the session.

Traced from source rather than reproduced on a store, so I may be missing something — loading the app on a store with no session row should settle it.

mrmarufpro:
If it's useful, I can send the `ensureInstalledOnShop` change as a small follow-up PR — either against `suleiman/express-token-exchange` so it ships with this, or against `main` right after this merges. Just the flag branch in that middleware plus a test for the empty-storage case, nothing else.

Happy either way if you'd rather fold it in yourself — didn't want it to fall between the two PRs.

Suleimanlatrsh:
Thanks @mrmarufpro, great catch ! you're right. Ported the token-exchange branch into `ensureInstalledOnShop`: when `unstable_tokenExchange` is on it skips the storage-based install check, embeds the app, and lets the first authenticated request mint the session via token exchange (rather than redirecting a fresh shop into the now-guarded OAuth routes). Added a test for the empty-storage case. Folded it into this PR so it ships together.

byrichardpowell (approved):

.changeset/express-token-exchange.md
Suleimanlatrsh: Done, renamed the flag to `tokenExchange` (dropped `unstable_`) across code, tests, docs, and this changeset.

.changeset/express-token-exchange.md
Suleimanlatrsh: Added a git-diff example to the changeset showing how to enable it (future flag + afterAuth registering webhooks).

Suleimanlatrsh: Pulled the important bits into the changeset: requirements, what changes when enabled (401 + retry header, document bounce, guarded OAuth routes), and the revoked-token note.

.changeset/express-token-exchange.md
Suleimanlatrsh: Done, @mrmarufpro and #3097 are now links.

packages/apps/shopify-app-express/docs/reference/guides/token-exchange.md
Suleimanlatrsh: Good call. reworded to 'Auth Code flow' throughout (docs + code), since both it and token exchange are OAuth flows.

PR #3352, opened 2026-07-31
Tighten @shopify/graphql-client public types (major)

Suleimanlatrsh:
## Why
Reworks the stale #1927 (open ~1.5 years, conflicting, contributor went quiet). The idea (safer, more descriptive public types for `@shopify/graphql-client`) is still wanted, and byrichardpowell had agreed to accept it. This is a clean redo from main.

## What
- `ResponseErrors.graphQLErrors` is now `GraphQLError[]` (with `message`, `locations`, `path`, and error `extensions`) instead of `any[]`, and a `GraphQLError` type is exported (and re-exported by the admin/storefront clients).
- `GQLExtensions` documents the Admin `cost`/`throttleStatus` and Storefront `context` shapes, while keeping a **permissive index signature (`any`)** so any other extension key still works and existing property access keeps compiling.
- Error `extensions` are typed separately from `GQLExtensions` (errors carry `{code, ...}`, not cost/context).
- Nested `cost`/`throttleStatus` fields are optional (the response body is unvalidated `any` at runtime, so we don't promise fields that may be absent).
- `RequestOptions` fields are now `readonly`.
- README type tables updated across graphql-client, admin-api-client, and storefront-api-client.

## Deliberately different from #1927
- **Does not narrow `CustomRequestInit.method`** to `'GET' | 'POST'`. That type is shared with the admin REST client, which uses `PUT`/`DELETE`; narrowing it breaks the REST client build. The GraphQL client already hardcodes `POST` internally.
- **Drops the `OperationVariables<TVariables>` generic**, which was a no-op in the original (nothing threaded `TVariables` through `request`).

## Breaking
Marked `major` via changeset for graphql-client and the three packages that re-export these types (admin-api-client, storefront-api-client, shopify-api). In practice it's very unlikely to affect an app: the extension types keep a permissive index signature, so existing property access keeps working. The main things a strict compiler could flag are reassigning `readonly` `RequestOptions` fields, or reading non-standard properties off a `graphQLErrors` entry.

## Testing
- `pnpm build` for graphql-client and all dependents (admin/storefront clients, shopify-api, express/remix/react-router) builds clean.
- `pnpm test:ci --filter='./packages/api-clients/graphql-client'`: 704 tests pass.

## Review history
Addresses two review rounds from byrichardpowell and river: dropped the type-test file / typecheck script, expanded + softened the changeset, and reworked the extension/error types so the change is backward-compatible in practice.


.changeset/gql-client-type-improvements.md
```diff
@@ -0,0 +1,11 @@
+---
+'@shopify/graphql-client': major
```
byrichardpowell, line 2: Since this package is consumed and exposed publicly by other packages, this is in-fact a breaking change  for those packages too.  We'll need to see which packages publicly expose this package and add those to the list here.

.changeset/gql-client-type-improvements.md
```diff
@@ -0,0 +1,11 @@
...
+---
+
+Tighten the public GraphQL client types for better type safety and editor hints.
+
+- `ResponseErrors.graphQLErrors` is now typed as `GraphQLError[]` (with `message`, `locations`, `path`, and `extensions`) instead of `any[]`. A new `GraphQLError` type is exported.
+- `GQLExtensions` now describes the `cost`/`throttleStatus` shape while keeping a top-level index signature, so non-cost extension keys still work.
+- `RequestOptions` fields are now `readonly`.
+
+This is a breaking change: code that reads `graphQLErrors` or `extensions` as loosely typed values, or reassigns `RequestOptions` fields, may need updating.
```
byrichardpowell: I think we can soften this. Since breaking changes can be scary it's worth noting that whilst this is technically a breaking change, it's very unlikely an app is affected by this in a significant way.

packages/api-clients/graphql-client/src/graphql-client/tests/types.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  options.variables = {id: '2'};
+
+  void options;
+}
+
+describe('graphql-client public types', () => {
+  it('type-checks via tsc (no runtime assertions)', () => {
+    expect(true).toBe(true);
+  });
+});
```
byrichardpowell: Not sure this test file is needed at all.  We don't usually test types.

packages/api-clients/graphql-client/tsconfig.typecheck.json
```diff
@@ -0,0 +1,11 @@
...
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true
+  },
+  "include": [
+    "./src/graphql-client/types.ts",
+    "./src/graphql-client/tests/types.test.ts"
+  ],
+  "exclude": ["./node_modules"]
+}
```
byrichardpowell: I think we can simplify here.  The tests seems overkill

.changeset/gql-client-type-improvements.md
```diff
@@ -0,0 +1,16 @@
+---
+'@shopify/graphql-client': major
+'@shopify/admin-api-client': major
+'@shopify/storefront-api-client': major
+'@shopify/shopify-api': major
```
byrichardpowell, line 5: Check if this is exposed in Remix/React Router & Express please. I think so? If so, we'd update this list.

Suleimanlatrsh: Checked, and yes to all three, so I've added them to the changeset as major:

- **shopify-app-remix** and **shopify-app-react-router** define their own `GraphQLResponse` type over `FetchResponseBody` (which carries `GQLExtensions` and `graphQLErrors`), so the changed types are baked into their own public API.
- **shopify-app-express** doesn't define its own wrapper types, but its public `ShopifyApp` type exposes `api: Shopify`, so the shopify-api client response types are reachable through it. Included it as major on that basis, happy to drop it to a transitive/patch bump if you'd rather treat it that way.

All three build clean against the new types, so this is semver signaling only. Full list is now 7 packages: graphql-client, admin-api-client, storefront-api-client, shopify-api, shopify-app-remix, shopify-app-react-router, shopify-app-express.

byrichardpowell (approved):

PR #3361, opened 2026-08-10
Fix 500 responses from authenticate.webhook when refreshing an expired offline token fails

iRaxe:

With the `expiringOfflineAccessTokens` future flag enabled, `authenticate.webhook` calls `ensureValidOfflineSession`, which attempts to refresh the stored offline token when it is expired. After a merchant uninstalls the app the stored token is typically expired by the time `APP_UNINSTALLED` (and later the mandatory GDPR webhooks) arrive, and Shopify rejects the refresh request because the app is no longer installed. The `refreshToken` helper then throws `Response(500)`, so the webhook delivery fails — silently, since thrown `Response`s are not logged — and Shopify keeps retrying it indefinitely. The app never gets a chance to run its uninstall cleanup, and every retry repeats the doomed refresh request.

Observed in production: every `app/uninstalled` delivery for a shop whose offline token had expired responded 500 on retry, forever, including for app-review stores.

At the point where the session is loaded, the request's HMAC has already been validated, and webhook contexts explicitly support a missing session ("webhooks can arrive after the app was uninstalled"). So `authenticate.webhook` now catches the `Response` thrown by the session refresh, logs a debug message, and delivers the webhook with `session: undefined` instead of failing the delivery. Non-`Response` errors (e.g. session storage failures) still propagate so those deliveries are retried.

Includes a regression test covering an expired offline session whose refresh request is rejected by Shopify, and a changeset.

## Type of change

## Checklist

### DIFF
--- a/.changeset/webhook-refresh-failure.md
+++ b/.changeset/webhook-refresh-failure.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-app-remix': patch
+---
+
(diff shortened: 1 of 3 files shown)


iRaxe:
CLA signed for both accounts

PR #3365, opened 2026-08-18
Remove eventId and resourceId from events webhooks

pobed2:

Shopify is dropping the `shopify-event-id` and `shopify-resource-id` headers from **events** deliveries, so the libraries should stop reading them.

Stops reading both headers. `shopify-event-id` was required, so events deliveries without it would have failed validation once Shopify stopped sending it. `eventId` and `resourceId` are also gone from the events validation result and the webhook context — use `webhookId` as the idempotency key. Webhooks are untouched and keep `X-Shopify-Event-Id`.

Going out as a major, per the discussion below: removing both fields from the public types breaks TypeScript apps that read them, and a fallback from `eventId` to `webhookId` isn't semantically sound, since one event can fan out to several deliveries that each carry their own `webhookId`. The changeset carries the migration note, and the generated docs data is regenerated so the published reference stops advertising the removed fields.

## Type of change

## Checklist


lizkenyon (commented):
So this is technically a breaking, so we have a couple of options

Option 1: Mark as major

 ### What you do

 1. Keep the current PR code. The headers are no longer read.
 2. Change the changeset from minor to major.
 3. Regenerate the docs files.
 4. Write a migration note. Tell apps to use webhookId.

Option 2: Deprecate fields, keep similar behavior

 ### What you do

 1. Stop requiring the shopify-event-id header. This is the validation fix.
 2. Keep eventId and resourceId in the public types.
 3. Keep ShopifyEventsHeader.EventId and ShopifyEventsHeader.ResourceId.
 4. Mark all four with @deprecated and a note.
 5. Fill eventId with a fallback value so it stays a string.
 6. Ship as a patch or minor.
 7. Remove the fields in the next major.

 ### The fallback choice

 eventId was a required string. If you keep the type but return nothing, callers get undefined. That is still a break for strict code.

 To keep the string contract, set a fallback:

 ```ts
   eventId: eventId ?? webhookId,
 ```
Question for you:

**Is this fallback acceptable based on what eventId and webhookId are intended to represent?** 

Context:
We try to minimize breaking changes and group them into a breaking change release. 
But if semantically the fallback does not make sense, we would have to go ahead and do the breaking change.

pobed2:
@lizkenyon 

> 
> To keep the string contract, set a fallback:
> 
>   `eventId: eventId ?? webhookId,`


Unfortunately, `webhookId` and `eventId` are fundamentally different things (and that's why we're removing `eventId`), so this fallback doesn't really work.

I think this means we need to make this a major version change?

s-harshdeep (approved):
LGTM

lizkenyon (approved):
TY!

PR #3366, opened 2026-08-19
Clarify Manual Pricing vs Shopify App Pricing in React Router and Remix billing reference docs

maxwell-blamauer:
## What this does

Two one-sentence JSDoc changes in `@shopify/shopify-app-react-router` and `@shopify/shopify-app-remix`, plus regenerated docs data and one consolidated patch changeset:

1. **`BillingContext` description:** "Provides utilities that apps can use to request [Manual Pricing](https://shopify.dev/docs/apps/launch/billing/manual-pricing) billing for the app using the Admin API." (production sentence + a two-word label and link)
2. **`createUsageRecord` description:** adds "[Shopify App Pricing](https://shopify.dev/docs/apps/launch/billing/shopify-app-pricing) supports usage-based plans through the [App Events API](https://shopify.dev/docs/api/app-events) instead of usage records."

All other method descriptions are untouched. No runtime changes.

## Why

AI agents and developers reading the generated billing references treat the Billing API as *the* way to bill a new public app, when Shopify App Pricing is the intended default. This is the package-side half of a Shopify.dev docs update that retitles the billing reference pages and adds a "Choose a pricing method" notice.

## Evidence from controlled local testing

Built a repeatable local A/B harness that gives the rendered production page and the updated page to the same model with identical prompts and settings, then scores which pricing system each answer recommends by default (methodology documented internally). Every answer is preserved and audited.

| Measure (5 scenarios x 3 runs per side, audited) | production | this change + page updates |
|---|---:|---:|
| New public app -> recommends Shopify App Pricing by default | 0/9 | 9/9 |
| Existing Billing API / one-time purchase -> keeps Manual Pricing | 6/6 | 6/6 |

The `createUsageRecord` sentence is specifically load-bearing: without it, agents concluded usage-based pricing was a model Shopify App Pricing does not support and routed new apps to Manual Pricing (1/3 correct); with it, 3/3. An explicit "supports ... through the App Events API" assertion is what prevents that misreading.

## Notes

- Docs-only JSDoc change shipped via the published `.d.ts` (IDE hovers) and the generated reference data.
- One consolidated patch changeset covers both packages.

### DIFF
--- a/.changeset/app-pricing-billing-docs.md
+++ b/.changeset/app-pricing-billing-docs.md
@@ -1,0 +1,6 @@
+---
+'@shopify/shopify-app-react-router': patch
+'@shopify/shopify-app-remix': patch
+---
+
+Clarify billing reference docs: label the `billing` object as Manual Pricing, and note that Shopify App Pricing supports usage-based plans through the App Events API instead of usage records.
--- a/.changeset/app-pricing-manual-pricing-label-remix.md
+++ b/.changeset/app-pricing-manual-pricing-label-remix.md
@@ -1,5 +1,0 @@
----
-'@shopify/shopify-app-remix': patch
----
-
-Label the `billing` object as Manual Pricing in its reference description and link to the Manual Pricing documentation.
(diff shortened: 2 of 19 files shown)


maxwell-blamauer:
Difference
Screenshot 2026-08-19 at 5 50 11 PM

lizkenyon (approved):
TY!

